PR #16, opened 2021-08-05
fix: policy filter in get all user api

rsbh:
In `Get Users` api when we apply the `role.tags` filter on the `policies` field, it is filtering the User list, not the subfield.

This happened due to the 'Where` clause is getting applied on the Users list.
So instead of doing that we can filter the casbin_rule policies at the time of aggregation.

This fix also adds support to pass multiple role tags in the query.

### DIFF
--- a/src/app/user/resource.ts
+++ b/src/app/user/resource.ts
@@ -25,12 +25,12 @@ export const create = async (payload: any) => {
 export const getListWithFilters = async (query: JSObj) => {
   // ? 1) Get all users with all policies
 
-  const roleTagFilter = extractRoleTagFilter(query);
+  const roleTagFilter = extractRoleTagFilter(query) || [];
   const policyFilters = R.omit(['fields'], query);
 
   const allUsersWithAllPolicies = await getSubjecListWithPolicies(
     'user',
-    roleTagFilter
+    [].concat(roleTagFilter)
   );
 
   // 3) fetch all groups with the matching attributes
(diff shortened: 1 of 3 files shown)


src/app/user/resource.ts
```diff
@@ -25,12 +25,12 @@ export const create = async (payload: any) => {
 export const getListWithFilters = async (query: JSObj) => {
   // ? 1) Get all users with all policies
 
-  const roleTagFilter = extractRoleTagFilter(query);
+  const roleTagFilter = extractRoleTagFilter(query) || [];
```
NNcrawler, line 28: In case of multi tags filter, how will the queryparams look like? And will it be parsed to array?

rsbh: We are using `qs` as [query parser]( https://github.com/odpf/shield/blob/main/src/config/composer.ts#L23)
it supports 4 options to pass array

```
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'indices' })
// 'a[0]=b&a[1]=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'brackets' })
// 'a[]=b&a[]=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'repeat' })
// 'a=b&a=c'
qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'comma' })
// 'a=b,c'

```

PR #20, opened 2021-09-02
feat: support json path to get proper respose from neseted values

pyadav:
Co-authored-by: Rahmat HD <rahmat.hd@gojek.com>


src/plugin/proxy/utils.ts
```diff
@@ -120,7 +121,8 @@ export const generateRoutes = (contents: Array<YAMLRoute> = []) => {
           iam: {
             permissions: route?.permissions || [],
             hooks: route?.hooks || []
-          }
+          },
+          jsonpath: route?.jsonpath
```
rsbh, line 125: can we use some other variable name as it gives the impression that it supports the `jsonpath` syntax and parsing?
Either change the variable name or implement [jsonpath](https://jsonpath.com)

PR #21, opened 2021-09-15
feat: supporting grpc proxy via go

kushsharma:
- major rewrite in `go`
- supports GRPC proxy


structs/middleware.go
```diff
@@ -0,0 +1,13 @@
+package structs
+
+import "net/http"
+
+type Middleware interface {
```
krtkvrm, line 5: why are we creating the interface here and not where it is used?

kushsharma: It's a good practice to define an interface where it's being used if that's the only place it's getting used. But in cases, where you want to announce in the codebase what should be a middleware interface, looks like, it's better to keep it out otherwise you have to define this in multiple places.

PR #23, opened 2021-09-30
Bootstrap

krtkvrm:


api/handler/v1/v1.go
```diff
@@ -0,0 +1,21 @@
+package v1
+
+import (
+	"context"
+	"github.com/odpf/salt/server"
+
+	shieldv1 "github.com/odpf/shield/api/protos/github.com/odpf/proton/shield/v1"
```
ravisuhag, line 7: @krtkvrm We can use proton gen code from buf build remotely and avoid keeping protos and gen code locally. cc @rohilsurana

rohilsurana: Looks like [the proto](https://github.com/odpf/proton/pull/45) has not been merged to master at `proton` yet. Unless that is done we won't have this proto in the buf registry.

Have received feedback that we can't use it unless there is support for branches. Will check with buf folks on that.

cmd/serve.go
```diff
@@ -13,5 +13,6 @@ func serveCommand(logger log.Logger, appConfig *config.Shield) *cli.Command {
 	}
 	c.AddCommand(proxyCommand(logger, appConfig))
 	c.AddCommand(adminCommand())
```
kushsharma, line 15: @krtkvrm can you remove this admin command now? I added this for admin APIs only. Also, we need to have a single command that starts proxy as well as admin APIs.

api/protos/shield.proto
```diff
@@ -0,0 +1,496 @@
+syntax = "proto3";
+package odpf.shield.v1;
```
rohilsurana, line 2: Is this needed here or are we using protobuf definitions from from proton?

PR #27, opened 2021-11-08
feat: Users api

rsbh:


api/handler/v1/user.go
```diff
@@ -2,33 +2,172 @@ package v1
...
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/odpf/shield/internal/user"
 
 	shieldv1 "go.buf.build/odpf/gw/odpf/proton/odpf/shield/v1"
 )
 
+type UserService interface {
+	GetUser(ctx context.Context, id string) (user.User, error)
```
krtkvrm, line 20: How about `GetUser` -> `Get`? In the code, we will be using as `UserService.Get()` ?

store/postgres/migrations/20211107162613_create_users_table.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS users (
+  id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+  name varchar UNIQUE NOT NULL,
+  email varchar UNIQUE NOT NULL,
+  metadata jsonb,
+  created_at timestamptz NOT NULL DEFAULT NOW(),
+  updated_at timestamptz NOT NULL DEFAULT NOW(),
+  deleted_at timestamptz
+);
```
krtkvrm, line 9: new line in the end 😅 , as that's how a file is defined, ending with a empty line.
Ref: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

internal/user/user.go
```diff
@@ -0,0 +1,58 @@
+package user
+
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+type User struct {
```
krtkvrm, line 9: We are moving these common structs to `model` pkg. 
cc: @AkarshSatija @ravisuhag

AkarshSatija (approved):

PR #29, opened 2021-11-18
fix: support http and http2 along with h2c and gRPC for proxy

krtkvrm:
This PR fixes the issue where we cannot use proxy for http & http2. This just works for h2c & grpc. 
Solving this by using different RoundTrippers for grpc and others
Tested this for
- [x] http using [echo server](https://ealenn.github.io/Echo-Server/)
- [x] http2 using [postb.in](https://postb.in/)
- [x] grpc using [integration/grpc_test.go#L200](https://github.com/odpf/shield/blob/main/integration/grpc_test.go#L200)
- [x] h2c using  [integration/rest_test.go#L260](https://github.com/odpf/shield/blob/main/integration/rest_test.go#L260)


krtkvrm:
continuing this on #32

PR #33, opened 2021-11-23
feat: add hook extract response data

krtkvrm:
- [x] Add hook interface that is similar to middleware, but for response from proxy
- [x] Add authz hook to extract attributes of type:
  - [x] json
  - [x] grpc


krtkvrm:
closing this & continuing working on #41

PR #35, opened 2021-12-06
chore: cleanup shield startup code and logging

aok1j1:
- [x] use zap logger instead of logrus
- [x] combine api and proxy into one cli command


cmd/serve.go
```diff
@@ -1,17 +1,221 @@
...
+	shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer shutdownCancel()
+
+	s.Shutdown(shutdownCtx)
+}
+
+func startServer(logger log.Logger, appConfig *config.Shield, err error, ctx context.Context, db *sql.SQL) *server.MuxServer {
+	s, err := server.NewMux(server.Config{
+		Port: appConfig.App.Port,
+	}, server.WithMuxGRPCServerOptions(getGRPCMiddleware(appConfig, logger)))
```
krtkvrm, line 103: we need to also configure zap to inject logger into ctx as well, which can be used by lower layers like grpc & store

cmd/serve.go
```diff
@@ -1,17 +1,221 @@
...
+		if err := ruleRepo.InitCache(ctx, ruleCacheRefreshDelay); err != nil {
+			return nil, nil, err
+		}
+		cleanUpFunc = append(cleanUpFunc, ruleRepo.Close)
+		pipeline := buildPipeline(logger, h2cProxy, ruleRepo)
+		go func(thisService config.Service, handler http.Handler) {
+			proxyURL := fmt.Sprintf("%s:%d", thisService.Host, thisService.Port)
+			logger.Info("starting h2c proxy", "url", proxyURL)
+
+			//s, err := server.NewMux(server.Config{
```
krtkvrm, line 147: is this comment for todo? if so, can you add `@TODO`? helpful to aggregate todos

cmd/serve.go
```diff
@@ -1,17 +1,221 @@
...
+	for _, service := range appConfig.Proxy.Services {
+		if service.RulesPath == "" {
+			return nil, nil, errors.New("ruleset field cannot be left empty")
+		}
+		blobFS, err := (&blobFactory{}).New(ctx, service.RulesPath, service.RulesPathSecret)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// TODO: option to use default http round tripper for http1.1 backends
```
krtkvrm, line 134: this is done in #32

cmd/serve.go
aok1j1: Injecting custom logger(wrapper over zap) into grpc would require some additional changes. will add a todo for now to push this for testing and keep track of. will try to make changes for it in background meanwhile.

rsbh (approved):

PR #39, opened 2021-12-09
fix: grpc payload deserialization should only happen for requested fields

kushsharma:


krtkvrm:
How about closing this PR..? as we are doing the same thing in #43 but also with nested message support

krtkvrm:
closing this as nested proto extraction PR has been merged #43

PR #42, opened 2021-12-10
chore: update docker build

rsbh:


Dockerfile
```diff
@@ -1,6 +1,16 @@
...
 
-COPY shield /usr/bin/shield
+RUN apk add make
 
-EXPOSE 8080
-ENTRYPOINT ["shield", "serve", "proxy"]
\ No newline at end of file
+WORKDIR /go/src/app
+
+COPY go.mod go.sum ./
```
krtkvrm, line 7: should we also copy `resource_configs` dir?

rsbh: we copy that in line [10](https://github.com/odpf/shield/pull/42/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R10)

this is just for caching mod install

krtkvrm (approved):

PR #43, opened 2021-12-13
feat: extract data from nested and repeated proto messages

krtkvrm:
- [x] Query Parser. Example: https://github.com/odpf/shield/pull/43/files#diff-8345ba859bc3ff6801d983f3df89e9b78a277d9c237337c0e272604be6636b10R18

Parsing Grammar
Token Size is 1
"/" Character not allowed as its reserved for END

Definitions
- INT: any single digit integer
- END: end of string, denoted by "/"

Table of Allowed Characters
| Current Character | allowed next characters |
|-------------------|-------------------------|
| INT           	| INT, ".", "[", END	  |
| "*"               | "]"                  	  |
| "."               | INT			          |
| "["				| "*"					  |
| "]"				| ".", END			      |

- [x] Building Dynamic Proto from Query 
`1.2.3[*]`
would be converted to
```
message a{
b field1 = 1;
}

message b{
c field2 = 2;
}

message c{
repeated string field3 = 3;
}
```
- [x] Query Runner to extract out data


pkg/body_extractor/grpc_payload.go
```diff
@@ -117,30 +138,152 @@ func fieldFromProtoMessage(msg []byte, tagIndex int) (string, error) {
...
+	val, err := RunQuery(dynamicMsgKey, parsedQuery)
 	if err != nil {
 		return "", err
 	}
 	return val.(string), nil
 }
 
-// should only be built once
-var genericProtoCache *desc.MessageDescriptor
+func buildPayloadProto(queries []Query) (*desc.MessageDescriptor, error) {
```
kushsharma, line 149: building this at every call doesn't feel like the best idea, we should cache it.

pkg/body_extractor/grpc_payload.go
```diff
@@ -117,30 +138,152 @@ func fieldFromProtoMessage(msg []byte, tagIndex int) (string, error) {
...
+		dm = dynamic.NewMessage(md)
+		if err := dm.ConvertFrom(pm); err != nil {
+			return nil, err
+		}
+	}
+
+	return RunQuery(dm, rest)
+}
+
+func ParseQuery(query string) ([]Query, error) {
```
kushsharma, line 230: Just a suggestion, we can refactor this to use a finite state machine with a bunch of `if/switch` statements and a few variables to manage the state. It will look like a LL(1) parser.

pkg/body_extractor/grpc_payload.go
```diff
@@ -117,30 +138,152 @@ func fieldFromProtoMessage(msg []byte, tagIndex int) (string, error) {
...
 	}
-	genericProtoCache = desc
-	return genericProtoCache, nil
+
+	if len(rest) == 0 {
+		return val, nil
+	}
+
+	dm, ok := val.(*dynamic.Message)
+	if !ok {
```
kushsharma, line 212: do we know in what cases this will not be ok?

pkg/body_extractor/grpc_payload.go
```diff
@@ -24,12 +28,24 @@ type GRPCPayloadCompressionFormat uint8
 const (
 	compressionNone GRPCPayloadCompressionFormat = 0
 	compressionMade GRPCPayloadCompressionFormat = 1
-	maxInt                                       = int(^uint(0) >> 1)
+
+	maxInt = int(^uint(0) >> 1)
+
+	NestedArray = "MessageArray"
```
kushsharma, line 34: Why not call this `MessageArray`

pkg/body_extractor/grpc_payload.go
```diff
@@ -24,12 +28,24 @@ type GRPCPayloadCompressionFormat uint8
 const (
 	compressionNone GRPCPayloadCompressionFormat = 0
 	compressionMade GRPCPayloadCompressionFormat = 1
-	maxInt                                       = int(^uint(0) >> 1)
+
+	maxInt = int(^uint(0) >> 1)
```
kushsharma, line 32: how are we using this?

krtkvrm: oh this is from earlier change, will remove it

krtkvrm: in grpc request parser
https://github.com/odpf/shield/blob/main/pkg/body_extractor/grpc_payload.go#L84

pkg/body_extractor/grpc_payload.go
krtkvrm: done

pkg/body_extractor/grpc_payload.go
krtkvrm: not sure about the cases, but considering `msg.TryGetFieldByNumber` and `TryGetRepeatedFieldByNumber` both return back `interface{}`, this is just a safety net.

pkg/body_extractor/grpc_payload.go
krtkvrm: makes sense, doing it

krtkvrm: done

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/odpf/shield
 
 go 1.16
 
+replace github.com/jhump/protoreflect => /Users/kartik/go/src/github.com/jhump/protoreflect
```
ravisuhag, line 5: @krtkvrm wrong package

krtkvrm: oh... will remove this before merging
this is being used for debugging

rsbh (approved):

PR #46, opened 2021-12-15
feat: add custom header matcher to read proxy header in grpc

aok1j1:
To read/pass headers to grpc context, grpc uses [DefaultHeaderMatcher](https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/runtime#DefaultHeaderMatcher) by default. 

To resolve this in shield where the identity proxy header will have to be processed, used the [WithIncomingHeaderMatcher](https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/runtime#WithIncomingHeaderMatcher) to match the proxy header and pass it to the grpc ctx.


cmd/serve.go
```diff
@@ -127,6 +131,17 @@ func startServer(logger log.Logger, appConfig *config.Shield, err error, ctx con
 	return s
 }
 
+func customHeaderMatcherFunc(proxyHeader string) func(key string) (string, bool) {
```
krtkvrm, line 134: How about making header matcher for `[]string` and not just for 1 key. Have a look at how [DefaultHeaderMatcher](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L107) is implemented

rsbh (approved):

PR #66, opened 2022-01-21
qa/fix: hooks to create resource for owner role

krtkvrm:
- [x] Owner able to create resource when groups is not mentioned


internal/bootstrap/bootstrap.go
```diff
@@ -86,11 +86,11 @@ func getResourceDefaultPolicies(ns model.Namespace, action model.Action, owner m
 			Namespace: ns,
 			Role:      definition.OrganizationAdminRole,
 		},
-		{
-			Action:    action,
-			Namespace: ns,
-			Role:      owner,
-		},
+		//{
```
krtkvrm, line 89: we should discuss this

PR #85, opened 2022-02-22
feat: org admin api

ishanarya0:
Scope:

- [x] Add admin API
- [x] Remove admin API
- [x] List admin API


ishanarya0:
Queries:
1. Since we need to return multiple users that have been added, we can't use the `GetContext` API of `sqlx`. Currently, I am using the `SelectContext` API, which does the task perfectly but doesn't reflect the actual intention of the query. 

2. I am building a single INSERT query to pass multiple values -> `INSERT INTO ... values (x,y,z), (p,q,r), (a,b,c)...`
Any better way to do this?

3. Where shall we use `txnErr` and `dbErr` respectively?

@rsbh @krtkvrm

krtkvrm:
https://stackoverflow.com/questions/12486436/how-do-i-batch-sql-statements-with-package-database-sql

store/postgres/org.go
```diff
@@ -26,6 +27,9 @@ const (
 	createOrganizationQuery = `INSERT INTO organizations(name, slug, metadata) values($1, $2, $3) RETURNING id, name, slug, metadata, created_at, updated_at;`
 	listOrganizationsQuery  = `SELECT id, name, slug, metadata, created_at, updated_at from organizations;`
 	updateOrganizationQuery = `UPDATE organizations set name = $2, slug = $3, metadata = $4, updated_at = now() where id = $1 RETURNING id, name, slug, metadata, created_at, updated_at;`
+	addOrganizationAdmins   = `INSERT INTO relations(subject_namespace_id, subject_id, object_namespace_id, object_id, role_id) values %s RETURNING subject_id;`
+	listOrganizationAdmins  = `SELECT subject_id from relations where object_id = $1 and role_id = 'organization_admin';`
+	removeOrganizationAdmin = `DELETE from relations where object_id = $1 and subject_id = $2 and role_id = 'organization_admin';`
 )
```
ishanarya0, line 33: @krtkvrm @rsbh Are `subject_id`,  `object_id` and `role_id` enough to form a composite candidate key? Or can there be a case for duplication? 

P.S. Using in `listOrganizationAdmins` and `removeOrganizationAdmin` queries.

ishanarya0:
> @ishanarya0 Let's add corresponding CLI commands as well for this. 

Sure, I am planning to add the CLI commands to this after it's successfully merged. Meanwhile, I am working on the Project Admin APIs.

krtkvrm:
relation from spicedb also needs to be deleted: `DeleteRelation(ctx context.Context, relation model.Relation)`

PR #86, opened 2022-02-23
fix: fix tags for shield config

bsushmith:
Add `mapstructure` tags to parse shield config through viper in expected ways

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,13 +11,13 @@ import (
 
 type Shield struct {
 	// configuration version
-	Version  int           `yaml:"version"`
-	Proxy    ProxyConfig   `yaml:"proxy"`
-	Log      LogConfig     `yaml:"log"`
-	NewRelic NewRelic      `yaml:"new_relic"`
-	App      Service       `yaml:"app"`
-	DB       DBConfig      `yaml:"db"`
-	SpiceDB  SpiceDBConfig `yaml:"spice_db"`
+	Version  int           `yaml:"version" mapstructure:"version"`
+	Proxy    ProxyConfig   `yaml:"proxy" mapstructure:"proxy"`
+	Log      LogConfig     `yaml:"log" mapstructure:"log"`
+	NewRelic NewRelic      `yaml:"new_relic" mapstructure:"new_relic"`
+	App      Service       `yaml:"app" mapstructure:"app"`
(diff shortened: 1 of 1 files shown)


krtkvrm:
@bsushmith can you explain the changes ?

bsushmith:
Currently, the way the config is read and parsed is thru viper, and viper uses `mapstructure` tags and if those are not present, it uses the var names(lowercase) in the struct itself as the config key string to parse from config file. 

https://github.com/odpf/salt/blob/main/config/config.go#L174 uses mapstructure tags to get the flattened keys. At the moment, we have mapstructure tags are present for all internal fields in the shield config, but not at the top level.

example:
``` go
type DBConfig struct {
	Driver          string        `yaml:"driver" mapstructure:"driver"`
	URL             string        `yaml:"url" mapstructure:"url"`
	MaxIdleConns    int           `yaml:"max_idle_conns" mapstructure:"max_idle_conns" default:"10"`
	MaxOpenConns    int           `yaml:"max_open_conns" mapstructure:"max_open_conns" default:"10"`
	ConnMaxLifeTime time.Duration `yaml:"conn_max_life_time" mapstructure:"conn_max_life_time" default:"10ms"`
	MaxQueryTimeout time.Duration `yaml:"max_query_timeout" mapstructure:"max_query_timeout" default:"100ms"`
}
```

without these changes, the config for example `spicedb` would be expected as `spicedb.pre_shared_key` rather than `spice_db.pre_shared_key` which leads to confusion as to half of it is getting parsed using one way and other part in other way.

rsbh:
These changes will break the current env keys, we need to update the env keys to handle this change.
viper create different keys with `yaml/json` and `mapstructure`  tags
ex. `yaml spicedb` will remove the underscore and create env `SPICEDB`
but with map structure, env keys preserve the underscore. `SPICE_DB`

ravisuhag:
> These changes will break the current env keys, we need to update the env keys to handle this change. viper create different keys with `yaml/json` and `mapstructure` tags ex. `yaml spicedb` will remove the underscore and create env `SPICEDB` but with map structure, env keys preserve the underscore. `SPICE_DB`

I think this is configurable.

ravisuhag:
This is already handled in a separate PR.

PR #88, opened 2022-02-25
fix: Add empty email id check

singhvikash11:


krtkvrm:
Can you please add PR description

singhvikash11:
## What?

> Add empty email-id check and throw an exception if email-id is empty.
Add email-id while updating user details for self.

## Why?

> From UI there is no check whether an email id is empty or blank. API can create a user with an empty email id and in case of a missing IAP header, it loads that user with an empty email id.

> User is not able to update its own details like name, metadata map against an email id since email id is missing.

singhvikash11:
@rsbh I modified code to refer to email from IAP header.

PR #92, opened 2022-03-02
feat(logger): inject custom zap logger into grpc

bsushmith:
making shield relevant changes using https://github.com/odpf/salt/pull/27

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -14,8 +14,6 @@ import (
 	"github.com/odpf/shield/grpc_interceptors"
 	"github.com/odpf/shield/pkg/sql"
 
-	"go.uber.org/zap"
-
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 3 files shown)


cmd/setup.go
```diff
@@ -44,7 +42,7 @@ func getGRPCMiddleware(cfg *config.Shield, logger log.Logger) grpc.ServerOption
 	return grpc.UnaryInterceptor(
 		grpcMiddleware.ChainUnaryServer(
 			grpc_interceptors.EnrichCtxWithIdentity(cfg.App.IdentityProxyHeader),
-			grpczap.UnaryServerInterceptor(zap.NewExample()),
+			grpczap.UnaryServerInterceptor(logger.(*log.Zap).GetInternalZapLogger().Desugar()),
```
mabdh, line 45: @bsushmith we can make it more defensive by always checking the casting and have a fallback scenario for this

bsushmith: Then, in that case, would the default fallback be `zap.NewExample()` itself? 🤔

krtkvrm:
@bsushmith any updates on this PR ?

bsushmith:
This PR is no longer needed. The corresponding changes were added and merged through the #164 - https://github.com/odpf/shield/pull/164/files#diff-d0706a4ae4ecf91c5224ca388f31a087a36af06c40dc81321da7958b25c70b23

PR #109, opened 2022-06-14
fix: Handle shield unauthorized error

singhvikash11:
As a non-admin User, tried to remove/add another user from a group, as expected Shield throws here an unauthorized error. But the API handler layer does not capture the error and throws grcpInternalError and its result to 500 HTTP error code in response.


singhvikash11:
As a non-admin User, tried to remove/add another user from a group, as expected Shield throws here an unauthorized error. But the API handler layer does not capture the error and throws grcpInternalError and its result to 500 HTTP error code in response.

bsushmith (approved):

PR #111, opened 2022-07-05
chore: add NewRelic segments for PG & SpiceDB

krtkvrm:
Adds NR Segments for

- [x] Postgres
- [x] SpiceDB

### DIFF
--- a/store/postgres/action.go
+++ b/store/postgres/action.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	newrelic "github.com/newrelic/go-agent"
 	"time"
 
 	"github.com/odpf/shield/pkg/utils"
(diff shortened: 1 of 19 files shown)


krtkvrm:
raising from another branch which has conflicts resolved

store/postgres/action.go
```diff
@@ -42,6 +43,14 @@ func (s Store) selectAction(ctx context.Context, id string, txn *sqlx.Tx) (model
 	var fetchedAction Action
 
 	err := s.DB.WithTimeout(ctx, func(ctx context.Context) error {
+		nr := newrelic.DatastoreSegment{
+			Product:    newrelic.DatastorePostgres,
+			Collection: fmt.Sprintf("actions"),
```
mabdh, line 48: Do we need `Sprintf` here? and should it be `Action` with uppercase `A` (similary like `Policy` above)?

store/postgres/org.go
```diff
@@ -87,6 +105,14 @@ func (s Store) CreateOrg(ctx context.Context, orgToCreate model.Organization) (m
 func (s Store) ListOrg(ctx context.Context) ([]model.Organization, error) {
 	var fetchedOrgs []Organization
 	err := s.DB.WithTimeout(ctx, func(ctx context.Context) error {
+		nr := newrelic.DatastoreSegment{
```
mabdh, line 108: We can create a function that only accept `ctx`, `Collection` and `Operation` args and return `nr`.
So everywhere in PG, calling it would be
```
nr := createNRSegment(ctx, "organizations", "List Organizations")
defer nr.End()
```

PR #118, opened 2022-07-15
feat: allow storing bool and num in metadata

### LINKED ISSUE #113: feat: allow storing boolean and number in group's metadata
**Is your feature request related to a problem? Please describe.**
I was trying to store extra group information with below details
```
[PUT] /v1beta1/groups/{id}
{
  "name": "My Group",
  "slug": "my-group",
  "orgId": "{ORG_ID}",
  "metadata": {
    "description": "my-group-description"
    "is_active": true
  }
}
``` 
But shield responded with
```
Status: 400 Bad Request
{
    "code": 3,
    "message": "invalid syntax in body",
    "details": []
}
```
When I tried changing `metadata.is_active` field to `"true"` (string) as opposed to `true` (boolean), it works.
But it is not viable in our use case because we need it to be stored as a boolean.

It seems like group's `metadata` field is only accepting map of string because it behaves the same when using integer/number.

**Describe the solution you'd like**
Allow storing `boolean` and `number` to group's `metadata` field.

niharbansal02:
Closes #113 

Allows storing boolean and numeric values in metadata.
- groups
- project
- org
- role
- user


bsushmith:
This PR seems to be doing a lot more changes than required and which do not seem tied to the linked issue. Is this expected? @niharbansal02

niharbansal02:
@bsushmith There are some PRs in review. This branch is branched from those branches, so there are some stacked commits.

niharbansal02:
@ravisuhag yes. I didn't know this earlier, I've taken care of this in #122

mabdh (approved):

PR #123, opened 2022-07-20
refactor: move domains to package `core/{domain}`

mabdh:
This is the first PR of some multiple tasks from this issue https://github.com/odpf/shield/issues/121

**Ideas**
The structure would be layered like this
- [server(handler)/hook/middleware] > service > repository
- All components `server(handler)/hook/middleware` should talk to service layer to process data

**Changes**
- move all entities in `model` to its own domain
- move [schema store interface](https://github.com/odpf/shield/blob/main/internal/schema/schema.go#L22) to its own domain 
- moving the [definition vars](https://github.com/odpf/shield/tree/main/internal/bootstrap/definition) to its domain package 
- move [authz spicedb](https://github.com/odpf/shield/tree/main/internal/authz) to store layer
- move config to its domain SpiceDBConfig to spicedb.Config
- refactor the cyclic import
  - avoid import cycle user<>group move ListUserGroups in user domain/service to group domain/service, let user be an entity that does not depend on anything
- permission package has responsibility to all domains, need to separate to its own domain
  - relation package [create function](https://github.com/odpf/shield/blob/main/internal/relation/relation.go#L34), pass a relation struct directly to store layer without picking the component  to make the create function reusable
  - move relation functions to to the domain that uses it
- Refactor CheckPermission to `CheckPermission(ctx context.Context, usr user.User, resourceNS namespace.Namespace, resourceIdxa string, action action.Action) (bool, error)` to avoid having dependency to `resource` package.
- Move CheckAuthz function in permission to ResourceService
- remove `structs` package
- refactor some errors with prefix `Err`
- Move integration test to `test/integration` folder
- Rename `v1.go` to `v1beta1.go`


mabdh:
This PR will also solve https://github.com/odpf/shield/issues/125

krtkvrm (approved):

Issue #129, opened 2022-07-25
Panic serving HTTP if `resource` config in authz hook rule is not string (json map)


mabdh:
**Describe the bug**
Shield proxy throw socket hang up and panic when the `resource` config in authz hook rule is not string.

**To Reproduce**
1. Prepare a mock backend that return this response with this endpoint `POST http://localhost:3000/test`
```response
{
    "data": {
        "id": "12345"
    }
}
```
3. Resource config
```resource.yaml
backends:
  - name: dummy
    resource_types:
      - name: dummy
        actions:
          read:
            - owner
          update:
            - team.team_member
```
4. Rules config
```rules.yaml
rules:
  - backends:
      - name: dummy
        target: "http://localhost:3000/"
        prefix: "/"
        frontends:
          - name: test_get_api
            path: "/test"
            method: "GET"
          - name: test_post_api
            path: "/test"
            method: "POST"
            hooks:
              - name: authz
                config:
                  action: authz_action
                  attributes:
                    resource:
                      key: data
                      type: json_payload
                    organization:
                      key: X-Shield-Org
                      type: header
                      source: request
                    project:
                      key: X-Shield-Project
                      type: header
                      source: request
                    resource_type:
                      value: "dummy"
                      type: constant
```
5. Run shield and mock server and hit path `/test` with POST to shield proxy
6. This will throw panic `http: panic serving [::1]:59990: interface conversion: interface {} is map[string]interface {}, not string`

**Expected behavior**
Should not throw panic but just error out instead

krtkvrm:
The attributes we extract out of the request or response is always expected to be a string.

PR #130, opened 2022-07-25
refactor: move server, store, hook, middleware to internal package

mabdh:
**Summary**

- Move hook and middleware package to proxy package and move proxy package to internal package
- Move server and store to internal package
- Move `api/handler` to internal as it is only being used by internal packages
- Do component creation only on cmd level
- Remove unnecessary log fatal from the main code, log fatal and panic should only be allowed to be executed in cmd, return error instead
- Decouple api.Dependencies from v1beta1
- Move server and proxy serving to  server and proxy package
- Fix graceful shutdown of all storages, proxies, and server
- Move resources and rules config to `config` folder
- Move `logger` and `grpc_interceptor` to `pkg`
- Rename `Id` to `ID`


mabdh:
This will also close bug in #129

krtkvrm (approved):

Issue #126, opened 2022-07-22
Creating namespace via API with no id will create a row with empty id


mabdh:
**Describe the bug**
Creating namespace via API with body contains only `name` (the id was not passed). This will create a row in table namespaces and return response with empty id.

**To Reproduce**
1. 
```
curl --location --request POST 'http://localhost:8000/admin/v1beta1/namespaces' \
--header 'X-Shield-Email: user@odpf.io' \
--header 'Content-Type: application/json' \
--data-raw '{
    "name": "req ns"
}'
```
2. Response got
```
{
    "namespace": {
        "id": "",
        "name": "req ns",
        "createdAt": "2022-07-22T06:02:34.396844Z",
        "updatedAt": "2022-07-22T06:02:34.396844Z"
    }
}
```
3. Row in `namespaces` table
<img width="996" alt="Screen Shot 2022-07-22 at 14 00 12" src="https://user-images.githubusercontent.com/4079709/180382627-4bcb06db-4091-434e-914a-adeae33e59e3.png">

**Expected behavior**
ID in table row should not be empty

mabdh:
This is also happened with `CreateRole`

mabdh:
Some discrepancies between each model

Using ID uuid and autogenerated
 - groups
 - organizations
 - policies
 - projects
 - relations
 - resources
 - users

Using ID varchar with no default value and NULLable
 - actions
 - namespaces
 - roles
  
The latter might raise an issue if we create/update the data via API. We might not need API creation for them since the data should be coming from resource.yml.

ravisuhag:
@mabdh Is this closed in 131?

mabdh:
yes correct @ravisuhag

PR #131, opened 2022-07-25
refactor: split stores into multiple repositories

mabdh:
Abstract out into 3 types of Repository
- Repository > main storage / Postgres
- AuthzRepository > for authz relation / SpiceDB
- ConfigRepository > for yaml config / Blob

Changes
- Split postgres store into multiple repositories
- Use pgx instead of lib/pq driver to catch PG specific error
- Cleanup the postgres code
- Separate the concern of Create/Update & Get/List for some entities
- Making UpdateNamespace not updating ID as primary key (this could leads to updating foreign keys)
- Move slug vs id to service layer


mabdh:
This will also resolve #126 
I have added the safety check to error out if empty

AkarshSatija (approved):

PR #132, opened 2022-07-26
feat: remove orgId dependency in authz

niharbansal02:


AkarshSatija:
Can we mention all the APIs being affected by this?

internal/proxy/hook/authz/authz.go
```diff
@@ -30,12 +32,18 @@ type Authz struct {
 	// To skip all the next hooks and just respond back
 	escape hook.Service
 
+	Deps api.Deps
```
mabdh, line 35: Don't add deps dependency, use project service interface that you have defined instead

internal/proxy/hook/authz/authz.go
```diff
@@ -194,16 +202,26 @@ func (a Authz) ServeHook(res *http.Response, err error) (*http.Response, error)
 	return a.next.ServeHook(res, nil)
 }
 
-func createResources(permissionAttributes map[string]interface{}) ([]resource.Resource, error) {
+func createResources(ctx context.Context, permissionAttributes map[string]interface{}, p ProjectService) ([]resource.Resource, error) {
 	var resources []resource.Resource
 	projects, err := getAttributesValues(permissionAttributes["project"])
 	if err != nil {
 		return nil, err
 	}
```
mabdh, line 211: Is it really needed to pass the project service as argument? If not, better to do this
```suggestion
func (a Authz) createResources(ctx context.Context, permissionAttributes map[string]interface{}) ([]resource.Resource, error) {
	var resources []resource.Resource
	projects, err := getAttributesValues(permissionAttributes["project"])
	if err != nil {
		return nil, err
	}
```

internal/proxy/hook/authz/authz.go
```diff
@@ -194,16 +202,26 @@ func (a Authz) ServeHook(res *http.Response, err error) (*http.Response, error)
...
 
-	orgs, err := getAttributesValues(permissionAttributes["organization"])
-	if err != nil {
-		return nil, err
+	var organizationIds []string
+	var projectIds []string
+	for _, proj := range projects {
+		project, err := p.Get(ctx, proj)
```
mabdh, line 215: With the previous changes, here you could do
```suggestion
		project, err := a.projectService.Get(ctx, proj)
```

internal/proxy/hook/authz/authz.go
```diff
@@ -177,7 +185,7 @@ func (a Authz) ServeHook(res *http.Response, err error) (*http.Response, error)
 		attributes[key] = value
 	}
 
-	resources, err := createResources(attributes)
+	resources, err := createResources(res.Request.Context(), attributes, a.projectService)
```
mabdh, line 188: With the changes from line 205, you could call `createResources`
```suggestion
	resources, err := a.createResources(res.Request.Context(), attributes)
```

mabdh (changes requested):

AkarshSatija (approved):

internal/proxy/hook/authz/authz.go
niharbansal02: Used Project Service to Get Organization.

PR #133, opened 2022-07-28
feat: slug support for list groups

niharbansal02:


internal/store/postgres/groups.go
```diff
@@ -212,10 +212,14 @@ func (s Store) ListGroups(ctx context.Context, org organization.Organization) ([
 	}
 
 	if org.ID != "" {
-		query = query + fmt.Sprintf(" WHERE org_id='%s'", org.ID)
+		organization, err := s.GetOrg(ctx, org.ID)
```
mabdh, line 215: @niharbansal02 there are a lot of changes regarding postgres store [here](https://github.com/odpf/shield/pull/131)
and please don't mix the logic of getting data via id and via slug in the db query itself unless it is really needed. and better not mix the definition of id and slug although they both unique but slug != id.
for this use case, we better separate the concern and make sure each function does 1 responsibility.

mabdh: also, please do add unit test everywhere we make changes and make sure the test is passed.

mabdh:
Is there any issue for this PR? @niharbansal02

krtkvrm:
the changes exist on main, was part of pr earlier

PR #134, opened 2022-07-28
feat: get resource from store using backend and resource type instead of urn

krtkvrm:
- [x] Add backend and resource type column
- [x] Change behaviour of bootstrap to update namespace table
- [x] Change behaviour in authz middleware to get resource from backend and resource type


mabdh:
@krtkvrm there are some changes in postgres store, I think you need to adjust the code to that.

krtkvrm:
Reopening as https://github.com/odpf/shield/pull/143

PR #142, opened 2022-08-02
fix: dockerfile dev file

mabdh:
We use Dockerfile.dev to build a tag dev. Our current Dockerfile.dev copied shield to /usr/local/bin/shield but the Dockerfile is copying shield to usr/bin/shield.

We expect the same path between Dockerfile and Dockerfile.dev. We need to update Dockerfile.dev so shield is copied to /usr/bin/shield.

**Changes**
- Copy shield to `usr/bin`
- Add `glibc` for compatibility purpose

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -11,6 +11,13 @@ COPY . .
 
 RUN make build
 
-FROM alpine:3.15
-COPY --from=builder /go/src/app/shield /usr/local/bin/
-ENTRYPOINT [ "shield" ]
+FROM alpine:3.16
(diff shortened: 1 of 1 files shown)


krtkvrm:
can you please add description ?

mabdh:
@krtkvrm the details is in the issue https://github.com/odpf/shield/issues/141
but anyway let me add it again here

mabdh:
@krtkvrm @AkarshSatija can we merge this?

ravisuhag (approved):

PR #143, opened 2022-08-02
feat: get resource from store using backend and resource type instead of urn

krtkvrm:
- [x] Add backend and resource type column
- [x] Change behaviour of bootstrap to update namespace table
- [x] Change behaviour in authz middleware to get resource from backend and resource type

### DIFF
--- a/core/namespace/namespace.go
+++ b/core/namespace/namespace.go
@@ -14,10 +14,12 @@ type Repository interface {
 }
 
 type Namespace struct {
-	ID        string
-	Name      string
-	CreatedAt time.Time
-	UpdatedAt time.Time
+	ID           string
+	Name         string
+	Backend      string
+	ResourceType string
+	CreatedAt    time.Time
+	UpdatedAt    time.Time
 }
 
 func strListHas(list []string, a string) bool {
--- a/core/namespace/namespace.go
+++ b/core/namespace/namespace.go
@@ -17,7 +17,7 @@ type Namespace struct {
 	ID           string
 	Name         string
 	Backend      string
-	ResourceType string `json:"resource_type"`
+	ResourceType string
 	CreatedAt    time.Time
 	UpdatedAt    time.Time
 }
--- a/core/namespace/namespace.go
+++ b/core/namespace/namespace.go
@@ -17,7 +17,7 @@ type Namespace struct {
 	ID           string
 	Name         string
 	Backend      string
-	ResourceType string
+	ResourceType string `json:"resource_type"`
 	CreatedAt    time.Time
 	UpdatedAt    time.Time
 }
@@ -36,5 +36,9 @@ func IsSystemNamespaceID(nsID string) bool {
 }
 
 func CreateID(backend, resourceType string) string {
+	if resourceType == "" {
+		return backend
+	}
+
 	return fmt.Sprintf("%s_%s", backend, resourceType)
 }
--- a/core/resource/resource.go
+++ b/core/resource/resource.go
@@ -20,6 +20,7 @@ type Repository interface {
 	Create(ctx context.Context, resource Resource) (Resource, error)
 	List(ctx context.Context, flt Filter) ([]Resource, error)
 	Update(ctx context.Context, id string, resource Resource) (Resource, error)
+	GetByNamespace(ctx context.Context, name string, ns namespace.Namespace) (Resource, error)
 }
 
 type ConfigRepository interface {
@@ -64,6 +65,8 @@ type Filter struct {
 }
 
 type YAML struct {
-	Name    string              `json:"name" yaml:"name"`
-	Actions map[string][]string `json:"actions" yaml:"actions"`
+	Name         string              `json:"name" yaml:"name"`
+	Backend      string              `json:"backend" yaml:"backend"`
+	ResourceType string              `json:"resource_type" yaml:"resource_type"`
+	Actions      map[string][]string `json:"actions" yaml:"actions"`
 }
--- a/core/resource/service.go
+++ b/core/resource/service.go
@@ -214,7 +214,7 @@ func (s Service) GetAllConfigs(ctx context.Context) ([]YAML, error) {
 	return s.configRepository.GetAll(ctx)
 }
 
-// TODO(krtkvrm): Separate Authz for Resources & System Namespaces
+// TODO(krkvrm): Separate Authz for Resources & System Namespaces
 func (s Service) CheckAuthz(ctx context.Context, res Resource, act action.Action) (bool, error) {
 	currentUser, err := s.userService.FetchCurrentUser(ctx)
 	if err != nil {
--- a/core/resource/service.go
+++ b/core/resource/service.go
@@ -214,13 +214,12 @@ func (s Service) GetAllConfigs(ctx context.Context) ([]YAML, error) {
 	return s.configRepository.GetAll(ctx)
 }
 
+// TODO: krkvrm | Separate Authz for Resources & System Namespaces
 func (s Service) CheckAuthz(ctx context.Context, res Resource, act action.Action) (bool, error) {
-	user, err := s.userService.FetchCurrentUser(ctx)
+	currentUser, err := s.userService.FetchCurrentUser(ctx)
 	if err != nil {
 		return false, err
 	}
-
-	res.URN = res.CreateURN()
 
 	isSystemNS := namespace.IsSystemNamespaceID(res.NamespaceID)
 	fetchedResource := res
@@ -228,7 +227,7 @@ func (s Service) CheckAuthz(ctx context.Context, res Resource, act action.Action
 	if isSystemNS {
 		fetchedResource.Idxa = res.URN
 	} else {
-		fetchedResource, err = s.repository.GetByURN(ctx, res.URN)
+		fetchedResource, err = s.repository.GetByNamespace(ctx, res.Name, res.Namespace)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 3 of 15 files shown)


internal/proxy/middleware/authz/authz.go
```diff
@@ -265,10 +267,16 @@ func createResources(permissionAttributes map[string]interface{}) ([]resource.Re
 	}
 
 	for _, res := range resourceList {
+		nsID := namespace.CreateID(backendNamespace[0], resourceType[0])
```
mabdh, line 270: Are we sure the backendNamespace and resourceType always have at least 1 element? This will cause panic if not

internal/store/postgres/namespace_repository_test.go
```diff
@@ -189,6 +207,8 @@ func (s *NamespaceRepositoryTestSuite) TestList() {
 				}
 			}
 			if !cmp.Equal(got, tc.ExpectedNamespaces, cmpopts.IgnoreFields(namespace.Namespace{}, "CreatedAt", "UpdatedAt")) {
+				fmt.Println(got)
+				fmt.Println(tc.ExpectedNamespaces)
```
mabdh, line 211: Maybe you forgot to remove these?

internal/store/postgres/resource_repository.go
```diff
@@ -213,3 +214,52 @@ func (r ResourceRepository) GetByURN(ctx context.Context, urn string) (resource.
...
+	}
+
+	getNamespaceQuery := dialect.Select("id").From(TABLE_NAMESPACES).Where(namespaceQueryExpression)
+	getResourcesByURNQuery, _, err := dialect.Select(&ResourceCols{}).From(TABLE_RESOURCES).Where(goqu.Ex{
+		"name":         goqu.L("$1"),
+		"namespace_id": goqu.Op{"in": getNamespaceQuery},
+	}).ToSQL()
+
+	return getResourcesByURNQuery, err
+}
```
mabdh, line 234: We can pass and assign the actual variables instead so we don't need to use if-else in GetContext later

```suggestion
func buildGetResourcesByNamespaceQuery(dialect goqu.DialectWrapper,  name string, ns namespace.Namespace) (string, []interface{}, error) {
	getNamespaceQuery := dialect.Select("id").From(TABLE_NAMESPACES)
	getNamespaceQuery = getNamespaceQuery.Where(goqu.Ex{
		"backend": ns.Backend),
	})
	if ns.ResourceType != "" {
	    getNamespaceQuery = getNamespaceQuery.Where(goqu.Ex{
		  "resource_type": ns.ResourceType),
	    })
	}
	
	getResourcesByURNQuery, params, err := dialect.Select(&ResourceCols{}).From(TABLE_RESOURCES).Where(goqu.Ex{
		"name":         name,
		"namespace_id": goqu.Op{"in": getNamespaceQuery},
	}).ToSQL()

	return getResourcesByURNQuery, params, err
}
```

internal/store/postgres/resource_repository.go
```diff
@@ -213,3 +214,52 @@ func (r ResourceRepository) GetByURN(ctx context.Context, urn string) (resource.
...
+
+	if errors.Is(err, sql.ErrNoRows) {
+		return resource.Resource{}, resource.ErrNotExist
+	} else if err != nil {
+		return resource.Resource{}, fmt.Errorf("%w: %s", dbErr, err)
+	}
+
+	if err != nil {
+		return resource.Resource{}, err
+	}
```
mabdh, line 262: Here, you can just use the result like this

```suggestion
	query, params, err := buildGetResourcesByNamespaceQuery(dialect, ns.ResourceType != "")
	if err != nil {
		return resource.Resource{}, fmt.Errorf("%w: %s", queryErr, err)
	}

	if err = r.dbc.WithTimeout(ctx, func(ctx context.Context) error {
		return r.dbc.GetContext(ctx, &fetchedResource, query, params...)
	}); err != nil {
	        switch {
	            case errors.Is(err, sql.ErrNoRows) {
		        return resource.Resource{}, resource.ErrNotExist
	            default:
	                 return resource.Resource{}, fmt.Errorf("%w: %s", dbErr, err)
	         }
	}
```

internal/proxy/middleware/authz/authz.go
krtkvrm: yes, 
- for backend namespace, we have check here: https://github.com/odpf/shield/pull/143/files#diff-389da525f8e810a300976024407d46ceccaa806befa60ab2cfe8da17892916aeR255
- if proxy doesn't catch resource_type, it will be `""`, defined here: https://github.com/odpf/shield/pull/143/files#diff-389da525f8e810a300976024407d46ceccaa806befa60ab2cfe8da17892916aeR106

internal/proxy/middleware/authz/authz.go
```diff
@@ -103,6 +103,8 @@ func (c *Authz) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	permissionAttributes["user"] = req.Header.Get(c.identityProxyHeader)
 
+	permissionAttributes["resource_type"] = ""
```
mabdh, line 106: Why don't we just  combine them altogether and write it like this?
```go
	permissionAttributes := map[string]interface{}{
            "namespace": rule.Backend.Namespace,
            "user": req.Header.Get(c.identityProxyHeader),
            "resource_type": "",
        }
```

internal/store/postgres/resource_repository.go
krtkvrm: resolved

internal/store/postgres/resource_repository.go
krtkvrm: resolved

internal/store/postgres/resource_repository.go
```diff
@@ -213,3 +214,46 @@ func (r ResourceRepository) GetByURN(ctx context.Context, urn string) (resource.
...
```
mabdh, line 256: CMMIW, if `err != nil`, this always returned in line `251` right?
So what is the point of adding line `254-256`?

core/namespace/namespace.go
```diff
@@ -14,10 +14,12 @@ type Repository interface {
...
 
 type Namespace struct {
-	ID        string
-	Name      string
-	CreatedAt time.Time
-	UpdatedAt time.Time
+	ID           string
+	Name         string
+	Backend      string
+	ResourceType string `json:"resource_type"`
```
spy16, line 20: Why json tag only this field?

core/resource/resource.go
```diff
@@ -20,6 +20,7 @@ type Repository interface {
 	Create(ctx context.Context, resource Resource) (Resource, error)
 	List(ctx context.Context, flt Filter) ([]Resource, error)
 	Update(ctx context.Context, id string, resource Resource) (Resource, error)
+	GetByNamespace(ctx context.Context, name string, ns namespace.Namespace) (Resource, error)
```
spy16, line 23: isn't there a string identifier that can be used to reference a namespace here? 

passing namespace struct leaves lot of things ambiguous:

1. what part of the struct are used in the lookup? 
2. will the query use CreatedAt and UpdatedAt as well? (probably not, but it's still possible to be done by the repo implementation)

core/resource/resource.go
```diff
@@ -67,6 +68,8 @@ type Filter struct {
 }
 
 type YAML struct {
```
spy16, line 70: i know it's not part of this change. but this struct name doesn't seem right. the fields are tagged with json & yaml tags. so it's not dedicated for yaml generation. and the name doesn't provide enough context

internal/store/postgres/resource_repository.go
spy16: +1. this `if err != nil` block is unreachable.

spy16: also, it is preferrable to indent error flows. 

```golang
err = r.dbc.WithTimeout(....)
if err != nil {
    if errors.Is(err, sql.ErrNoRows) {
        return resource.Resource{}, resource.ErrNotExist
    }
    return resource.Resource{}, fmt.Errorf("%w: %s", dbErr, err)
}
```

The reason is it makes it easier to avoid the kind of mistakes that's been made here..

core/resource/service.go
```diff
@@ -212,25 +212,24 @@ func (s Service) GetAllConfigs(ctx context.Context) ([]YAML, error) {
 	return s.configRepository.GetAll(ctx)
 }
 
+// TODO: krkvrm | Separate Authz for Resources & System Namespaces
```
spy16, line 215: `TODO(user): .....` might be a better format. (easier to search etc.)

core/resource/resource.go
mabdh: I am not sure where do we use the `json` tags is. But AFAIK this struct is used to parse yaml file and populated from it.

cc @krtkvrm

spy16: hmm, i think it's okay to leave the json tags in the struct. but i feel the name could be better (maybe based on what data it represents or something like that)

mabdh: @spy16 any suggestion on that?
Since this is just a placeholder for yaml data, I was thinking something like `Config`.

spy16: isn't this the resource-type definition from the file? 🤔 `ResourceTypeDef` could be used in that case?

core/namespace/namespace.go
```diff
@@ -35,5 +37,9 @@ func IsSystemNamespaceID(nsID string) bool {
 
 //postgres://shield:@:5432/
```
rohilsurana, line 38: Do we need this comment anymore?

core/resource/resource.go
krtkvrm: we can get resource by combination between resource name and:
1. namespace_id
2. backend + type
3. backend

thus, using struct

krtkvrm: signed off during standup

rohilsurana (approved):

PR #144, opened 2022-08-03
feat: prevent proxy context cancelled

krtkvrm:
Expectation
- [x] request cancellation or interruption from client side should not stop processing on shield side

### DIFF
--- a/internal/proxy/roundtripper.go
+++ b/internal/proxy/roundtripper.go
@@ -32,7 +32,10 @@ func (t *h2cTransportWrapper) RoundTrip(req *http.Request) (*http.Response, erro
 	// we need to apply errors if it failed in Director
 	if err, ok := req.Context().Value(ctxRequestErrorKey).(error); ok {
 		return nil, err
+	} else if req.Context().Err() != nil {
+		return nil, req.Context().Err()
 	}
+
 	t.log.Debug("proxy request", "host", req.URL.Host, "path", req.URL.Path,
 		"scheme", req.URL.Scheme, "protocol", req.Proto)
 
--- a/internal/proxy/roundtripper.go
+++ b/internal/proxy/roundtripper.go
@@ -1,7 +1,6 @@
 package proxy
 
 import (
-	"context"
 	"crypto/tls"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


internal/proxy/roundtripper.go
```diff
@@ -28,6 +29,30 @@ type h2cTransportWrapper struct {
 	hook hook.Service
 }
 
+type Context struct {
+	ctx context.Context
+}
+
+func (c Context) Deadline() (time.Time, bool) {
+	return time.Time{}, false
```
krtkvrm, line 37: doing 
``` go
t, b := c.ctx.Deadline()
return t, b
```

was throwing stack overflow

internal/proxy/roundtripper.go
```diff
@@ -28,6 +29,30 @@ type h2cTransportWrapper struct {
...
+}
+
+func (c Context) Done() <-chan struct{} {
+	return nil
+}
+
+func (c Context) Err() error {
```
krtkvrm, line 44: We can also only prevent "Context Cancelled" errors, but let others to return

internal/proxy/roundtripper.go
```diff
@@ -28,6 +29,30 @@ type h2cTransportWrapper struct {
...
 
+type Context struct {
+	ctx context.Context
+}
+
+func (c Context) Deadline() (time.Time, bool) {
+	return time.Time{}, false
+}
+
+func (c Context) Done() <-chan struct{} {
```
krtkvrm, line 40: we can also log when done != nil

mabdh:
@krtkvrm for the PR prefix, is it a `feat:` or a `fix:` ?

internal/proxy/roundtripper.go
```diff
@@ -28,6 +29,30 @@ type h2cTransportWrapper struct {
...
+}
+
+func (c Context) Value(key interface{}) interface{} {
+	return c.ctx.Value(key)
+}
+
+func WithoutCancel(ctx context.Context) context.Context {
+	return Context{ctx: ctx}
+}
+
```
mabdh, line 55: I think this could go to its own file called `context.go` maybe?

spy16 (approved):

PR #146, opened 2022-08-03
refactor!: api http statuscode

mabdh:
- Add `mockery` for mocking interface
- For APIs that require Auth Email header, return 403 Forbidden if not present/user does not exist
- Fix error type naming
- Create a function to check empty string
- Add tests in handler

Status code changes:
- Bad request -> 400
- Conflict/Insert existing resource -> 409
- Unauthorized (in check permission) -> 403 (not 401 since that is used when not able to establish identity from credentials or credentials are missing)
- No identity header -> 401
- Server related error -> 500


mabdh:
Need to hold this until v0.3.x is stable enough

mabdh:
This can be reviewed

core/group/service.go
```diff
@@ -78,7 +79,7 @@ func (s Service) List(ctx context.Context, flt Filter) ([]Group, error) {
 }
 
 func (s Service) Update(ctx context.Context, grp Group) (Group, error) {
-	if grp.ID != "" {
+	if !str.IsStringEmpty(grp.ID) {
```
spy16, line 82: just a side note on this `str.IsStringEmpty` function. It is advertised and exposed as a public "generic" function. It has no document comment as well. But it does a `TrimSpace()` before doing an empty check. So it's not really generic. And using `!str.IsStringEmtpy()` in place of `grp.ID != ""` is actually changing the behaviour slightly (In this particular case, it is for the better). 

Usually it's better to avoid trying to extract everything into "reusable" functions. Just doing `strings.TrimSpace(val) != ""` is good enough as well.

core/group/service.go
```diff
@@ -99,13 +104,13 @@ func (s Service) AddUsers(ctx context.Context, groupIdOrSlug string, userIds []s
...
 
-	isAuthorized, err := s.relationService.CheckPermission(ctx, currentUser, namespace.DefinitionTeam, groupID, action.DefinitionManageTeam)
+	isAllowed, err := s.relationService.CheckPermission(ctx, currentUser, namespace.DefinitionTeam, groupID, action.DefinitionManageTeam)
 	if err != nil {
 		return []user.User{}, err
 	}
 
-	if !isAuthorized {
-		return []user.User{}, errors.Unauthorized
+	if !isAllowed {
```
spy16, line 112: merge this if into the above error handling block ? (all error handling together feels better, for me at-least).

```golang
if err != nil {
   return nil, err
} else if !isAllowed {
   return nil, errors.ErrForbidden
}
```

```

File: core/group/service.go [hunk 3]
```diff
@@ -99,13 +104,13 @@ func (s Service) AddUsers(ctx context.Context, groupIdOrSlug string, userIds []s
...
+		return []user.User{}, errors.ErrForbidden
```
spy16, line 113: why are we returning `[]user.User{}` ? We should be returning `nil` here.

core/group/service.go
```diff
@@ -187,7 +192,11 @@ func (s Service) ListAdmins(ctx context.Context, idOrSlug string) ([]user.User,
 func (s Service) AddAdmins(ctx context.Context, groupIdOrSlug string, userIds []string) ([]user.User, error) {
 	currentUser, err := s.userService.FetchCurrentUser(ctx)
 	if err != nil {
-		return []user.User{}, err
+		return []user.User{}, fmt.Errorf("%w: %s", user.ErrInvalidEmail, err.Error())
+	}
+
+	if len(userIds) < 1 {
```
spy16, line 198: this `if` block can be merged to the above error handling block (to keep all error handling together)

core/organization/service.go
```diff
@@ -81,7 +82,11 @@ func (s Service) Update(ctx context.Context, org Organization) (Organization, er
 func (s Service) AddAdmins(ctx context.Context, idOrSlug string, userIds []string) ([]user.User, error) {
 	currentUser, err := s.userService.FetchCurrentUser(ctx)
 	if err != nil {
-		return []user.User{}, err
+		return []user.User{}, fmt.Errorf("%w: %s", user.ErrInvalidEmail, err.Error())
```
spy16, line 85: in lot of places we are returning `[]user.User{}`. Wherever there's a need to use empty slice, we should be using `nil`. It prevents an allocation. And has no difference (It cannot cause any nil pointer issues).

⚠️ Only exception to this rule is during JSON marshal. Because `nil` slice value is written out as `null` and empty-slice is marshaled as `[]` (json array)

mabdh: @spy16 Agree on this, we could just return nil for the empty slices. I keep that part apart from the refactor and this PR just want to don't want to have more-huge changes on this. 

This could be addressed in separate PR and issue.

core/group/service.go
mabdh: answered here
https://github.com/odpf/shield/pull/146#discussion_r948686861

core/group/service.go
mabdh: I think you have a fair point on that. I think this one is not necessary to be extracted out. Keeping that inline would make this more readable. Thanks for the input @spy16

core/group/service.go
mabdh: For this, even I think the `s.relationService.CheckPermission` only needs to return an error. Returning the status `isAlowed` or `isAuthorized` is not really needed. Inside, it could check end return error and we can check the type of error in this CheckPermission or maybe in the handler.
But again, don't want to change the logic that much.

I would actually prefer not to use `else` because sometimes it is confusing but I think we can merge the error handling block together.

core/organization/service.go
spy16: makes sense.

internal/api/v1beta1/user.go
```diff
@@ -148,7 +148,7 @@ func (h Handler) GetCurrentUser(ctx context.Context, request *shieldv1beta1.GetC
 		return nil, grpcPermissionDenied
 	}
 
-	if str.IsStringEmpty(email) {
+	if strings.TrimSpace(email) == "" {
```
spy16, line 151: Do we do any other sanitisation steps? I don't know if this is user-provided value or not.

But usually, it's better to do it like this:
```golang
email = strings.TrimSpace(email)
if email == "" {
   return errors.New("invalid email")
}
```

Otherwise, I can send `"           foo@bar.com         "` and get stored.

mabdh: @spy16 This is user provided value. No, we mostly don't have any input checking in shield. Previously, for some domains you can even send an empty string as an id and stored in the DB.  I just added several points but maybe not covering all of them.
Agree on that, we better validate the input and return early or do sanitization. There should be another issue that focuses on this. For the email case in this PR, let me refactor as you suggested.

spy16: yea. may be create an issue about validation so that we can pick it up separately. definitely better not to add this into scope of this PR 👍

spy16 (approved):
Apart from few validation checks etc (which will be fixed in separate issue), this looks fine to me. But would recommend waiting for at-least 1 other reviewer since there are lot of changes here.

PR #155, opened 2022-08-15
feat!: config management and error handling

mabdh:
**Tasks**
- [x] Error handling when cli config not found, for api that does not need config, could still do the call, otherwise return warning
- [x] Loading config for client init, pick from the root, but the flag could override it
- [x] Loading protos from proton based on a commit in Makefile
- [x] Namespacing server commands
- [x] Init for server config

**Error handling in client config**
The existence of `$ODPF_CONFIG_DIR/shield.yaml`
- If user does not have this file in the path
  - calling shield client cli without --host would throw error `need a config or flag --host`
  - calling shield client cli that not requires auth
    - with `--host` would pass
  - calling shield cli that requires auth
    - with `--host` would return `--header` is required (we mark this as required)
    - with `--host` and `--header` would pass

**Some changes in cli**
```diff
+ shield config init // initializing client config
+ shield config list // list down client configurations
+ shield server init // initializing server config

- shield serve
+ shield server start

- shield migrate
+ shield server migrate

- shield migrate-rollback
+ shield server migrate-rollback
```


config/config.go
```diff
@@ -48,6 +48,11 @@ func Load() (*Shield, error) {
 		options = append(options, config.WithPath(filepath.Join(currentHomeDir, ".config")))
 	}
 
+	// override all config sources and prioritize one from file
+	if configfileFromFlag != "" {
```
ravisuhag, line 52: @mabdh This is for server configs, right?

cmd/client.go
```diff
@@ -5,20 +5,27 @@ import (
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithBlock(),
 	}
 
 	return grpc.DialContext(ctx, host, opts...)
 }
 
-func createClient(ctx context.Context, host string) (shieldv1beta1.ShieldServiceClient, func(), error) {
-	dialTimeoutCtx, dialCancel := context.WithTimeout(ctx, time.Second*2)
+func createClient(cmd *cobra.Command) (shieldv1beta1.ShieldServiceClient, func(), error) {
```
ravisuhag, line 22: @mabdh Any specific reason to pass the entire `cmd` object instead of just the host for this method?

cmd/client.go
```diff
@@ -33,3 +40,23 @@ func createClient(ctx context.Context, host string) (shieldv1beta1.ShieldService
 	client := shieldv1beta1.NewShieldServiceClient(conn)
 	return client, cancel, nil
 }
+
+func isClientCLI(cmd *cobra.Command) bool {
```
ravisuhag, line 44: @mabdh good for now, in future this method check of annotations can be moved to salt.

config/config.go
mabdh: Yes it is for server, maybe naming this explicitly with `server` will give more clarity, will update this thanks @ravisuhag

cmd/client.go
mabdh: This is because we use cmd flag as the source of truth of cli config. So we are copying data from cli.Config if exist and not the other way around. 

Passing cmd is just to avoid having one line `cmd.Flags().GetString("host")` for every client that needs it.

mabdh: Now I am thinking again, I think the ideal way should be making the `cli.Config` as the source of truth if there is any flag passed, we override the fields in the `cli.Config` itself and use the `cli.Config` in our logic.

ravisuhag: +1

spy16: +1. it is possible to bind cobra and viper together to do this automatically (assuming cli.Config is being loaded from that). But it requires some work. 

(alternate option might be to just have a helper function `readFlags(cmd *cobra.Command, into *cli.Config)` which checks if flags are set and overrides the struct fields may be)

mabdh: @spy16 @ravisuhag updated the code, can you guys please check, basically does  similar flow like the `readFlags` that @spy16 suggested above.

ravisuhag (approved):

ravisuhag:
@mabdh Let's also move the shield sample config file to the config package itself and call it `config.yaml` Ideally, we would want to encourage the behavior of using shield server init to generate server config and not manually copy-paste the file.

config/.shield.sample.yaml
```diff
@@ -0,0 +1,52 @@
+version: 1
```
ravisuhag, line 1: @mabdh let's rename this to just `config.yaml` (without sample word) since we use `shield.yaml` for client config.  wdyt?

mabdh: yes makes sense, updating it

.gitignore
```diff
@@ -9,7 +9,7 @@ proxies/*.yml
 temp
 shield
 !proto/*
-.shield.yaml
+config.yaml
```
ravisuhag, line 12:
```suggestion
/config.yaml
```
@mabdh Without this it will ignore any config.yaml file anywhere.

mabdh: yes you are correct, thanks @ravisuhag for the heads up. 
will recheck what else missing

PR #158, opened 2022-08-18
feat: streamline attribute extraction in middleware and hooks

### LINKED ISSUE #157: Streamline attribute extraction in middleware and hooks
**Is your feature request related to a problem? Please describe.**
In the middleware stack, the attributes such as projectId, groupId are extracted in each layer wherever needed.

**Describe the solution you'd like**
There can be a single layer that takes care of all the extraction that all the other layers could use.

**Describe alternatives you've considered**
Extract attributes in each layer and pass them on to the other layers for their use(if any).

niharbansal02:
Created a new layer for attribute extraction right after `rulematch`.
It pushes all the attributes into the context for the use of middleware and hook.

Current flow: Request -> [Middleware starts] Rule Matching -> Basic Auth -> Authz -> Prefix -> Backend -> [Hook starts] Authz -> Response

Modified flow with this PR: Request -> [Middleware starts] Rule Matching -> **_Attribute Extraction_** -> Basic Auth -> Authz -> Prefix -> Backend -> [Hook starts] Authz -> Response
Closes #157

### DIFF
--- a/internal/proxy/hook/authz/authz.go
+++ b/internal/proxy/hook/authz/authz.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/mitchellh/mapstructure"
 
@@ -12,10 +11,7 @@ import (
 	"github.com/odpf/shield/core/namespace"
 	"github.com/odpf/shield/core/project"
 	"github.com/odpf/shield/core/resource"
-	"github.com/odpf/shield/core/user"
 	"github.com/odpf/shield/internal/proxy/hook"
-	"github.com/odpf/shield/internal/proxy/middleware"
-	"github.com/odpf/shield/pkg/body_extractor"
 )
 
 type ResourceService interface {
@@ -33,9 +29,6 @@ type Authz struct {
 
 	projectService ProjectService
 
-	// TODO need to figure out what best to pass this
-	identityProxyHeaderKey string
-
 	resourceService ResourceService
 }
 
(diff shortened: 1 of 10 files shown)


internal/proxy/hook/authz/authz.go
```diff
@@ -71,8 +63,10 @@ func (a Authz) ServeHook(res *http.Response, err error) (*http.Response, error)
 		return a.escape.ServeHook(res, err)
 	}
 
-	ruleFromRequest, ok := hook.ExtractRule(res.Request)
-	if !ok {
+	attributes := res.Request.Context().Value("requestAttributes").(map[string]any)
```
spy16, line 66: this can cause panics if the type changes or the value is not set etc. 

Also, using primitive Go types as context-keys is highly discouraged. Ideally, the package which is responsible for injecting this particular value into context should also expose a function to extract the value. The details about what key is used, what is the type of the key, etc. should be hidden awy.

```golang
package requestctx

var reqAttribsKey = ctxKey("request_attributes")

type ctxKey string // should be private.

func InjectAttribs(ctx context.Context, value map[string]any) context.Context {
   return context.WithValue(ctx, reqAttribsKey, value)
}

func AttribsFrom(ctx context.Context) map[string]any {
   // even if the second value is ignored, this operation is safe. 
   // using "v := ctx.Value(reqAttribsKey).(map[string]any)" is NOT safe.
   v, _ := ctx.Value(reqAttribsKey).(map[string]any)
   return v
}
```

internal/proxy/middleware/attributes/attributes.go
```diff
@@ -0,0 +1,174 @@
...
+		a.log.Error("middleware: path param map doesn't exist")
+		a.notAllowed(rw)
+		return
+	}
+
+	for key, value := range paramMap {
+		requestAttributes[key] = value
+	}
+
+	attributesContext := context.WithValue(req.Context(), "requestAttributes", requestAttributes)
```
spy16, line 170: do not use primitive string value as key in contexts. Refer to my other comment. the snippet provided in my other comment can be added into this package and used as `attributesContext := InjectAttribs(req.Context(), requestAttributes)`.

internal/proxy/hook/authz/authz.go
```diff
@@ -71,8 +64,13 @@ func (a Authz) ServeHook(res *http.Response, err error) (*http.Response, error)
 		return a.escape.ServeHook(res, err)
 	}
 
-	ruleFromRequest, ok := hook.ExtractRule(res.Request)
+	attributes, ok := attributes2.GetAttributesFromContext(res.Request.Context())
```
spy16, line 67: can you do the change that we discussed for this naming? var can be renamed to `attribs` or `attribMap` or something like that.

AkarshSatija (approved):

PR #163, opened 2022-08-18
feat: enable upsert behaviour

ishanarya0:


krtkvrm (changes requested):
@ishanarya0 can you resolve conflicts

internal/api/v1beta1/user.go
```diff
@@ -121,6 +123,33 @@ func (h Handler) CreateUser(ctx context.Context, request *shieldv1beta1.CreateUs
 	}}, nil
 }
 
+func (h Handler) CreateMetadataKey(ctx context.Context, request *shieldv1beta1.CreateMetadataKeyRequest) (*shieldv1beta1.CreateMetadataKeyResponse, error) {
```
krtkvrm, line 126: can you add tests for these ?

internal/store/postgres/migrations/20220819074139_add_metadata_keys_table.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS metadata_keys;
```
krtkvrm, line 1: can you add new line at end of these sql files? [thats how posix define a file](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_205)

internal/store/postgres/user_repository.go
```diff
@@ -6,18 +6,31 @@ import (
...
 	"github.com/odpf/shield/core/user"
 	"github.com/odpf/shield/pkg/db"
+	"github.com/odpf/shield/pkg/uuid"
 )
 
 type UserRepository struct {
 	dbc *db.Client
 }
 
+type userXMetadata struct {
```
krtkvrm, line 24: can you describe what is userXMetadata ?

internal/store/postgres/migrations/20220819074139_add_metadata_keys_table.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS metadata_keys (
+    id           uuid        PRIMARY KEY     DEFAULT uuid_generate_v4(),
```
krtkvrm, line 2: why do we need `id` if we can just use `key`?

core/user/errors.go
```diff
@@ -3,10 +3,13 @@ package user
...
-	ErrNotExist     = errors.New("user doesn't exist")
-	ErrInvalidID    = errors.New("user id is invalid")
-	ErrInvalidEmail = errors.New("user email is invalid")
-	ErrMissingEmail = errors.New("user email is missing")
-	ErrInvalidUUID  = errors.New("invalid syntax of uuid")
-	ErrConflict     = errors.New("user already exist")
+	ErrNotExist         = errors.New("user doesn't exist")
+	ErrInvalidID        = errors.New("user id is invalid")
+	ErrInvalidEmail     = errors.New("user email is invalid")
+	ErrNotUUID          = errors.New("invalid syntax of uuid")
```
rohilsurana, line 9: Looks like this is not being used anywhere. Do we need this?

internal/api/v1beta1/errors.go
```diff
@@ -20,5 +20,6 @@ var (
 	grpcPermissionDenied    = status.Error(codes.PermissionDenied, errors.ErrForbidden.Error())
 	grpcUnauthenticated     = status.Error(codes.Unauthenticated, errors.ErrUnauthenticated.Error())
 
-	ErrEmptyEmailID = errors.New("email id is empty")
+	ErrEmptyEmailID    = errors.New("email id is empty")
+	ErrInvalideEmailID = errors.New("user email is invalid")
```
rohilsurana, line 24:
```suggestion
	ErrInvalidEmailID = errors.New("user email is invalid")
```

rohilsurana: Also seems like this is also not being used. Lets check if we need this.

internal/store/postgres/user_repository.go
ishanarya0: I defined this struct to store the rows from the join of the Users and Metadata table.
This could be handled by using both the User and Metadata struct together. But, that approach would be chaotic.

What do you think, which approach is better?

internal/store/postgres/migrations/20220819074139_add_metadata_keys_table.up.sql
ishanarya0: Yes, the key as the primary key would have worked fine as well. We can change it to key.

One thing I am not sure about is, let's say we later introduce get, update, and delete APIs as well. Would it be fine to use the "key" itself as the identifier or would it be good(safe) to use a UUID there?

core/user/errors.go
ishanarya0: Yes, `ErrNotUUID ` isn't being used.

internal/api/v1beta1/errors.go
ishanarya0: Yes, removed `ErrInvalidEmailID`.

internal/store/postgres/migrations/20220819074139_add_metadata_keys_table.up.sql
ishanarya0: Removed `id`

internal/store/postgres/user_repository.go
ishanarya0: changed struct to `joinUserMetadata`

krtkvrm:
@ishanarya0 can you fix lint

krtkvrm (approved):

PR #167, opened 2022-08-21
style: upgrade salt for command grouping

ravisuhag:


cmd/action_test.go
```diff
@@ -45,11 +19,6 @@ func TestClientAction(t *testing.T) {
 			want        string
 			err         error
 		}{
-			{
```
ravisuhag, line 48: @mabdh removing this test for two reasons. 
It will get trickier to write style test when any special characters come into place. As an example, bold and color characters will be hard to test this way. 
This test won't give much value since usage is auto-generated. I think we can start to test cmds by checking list of commands available in that cmds. which can be handled in another issue. wdyt?

mabdh: Make sense

mabdh (approved):

Issue #176, opened 2022-10-04
schema migration & revision in ShieldDB for resources


krtkvrm:
With respect to changes in principals for Resources, we are supposed to change PG schema which is very fixed with Team and Owner.  We will move to only Owner role for resource and it can be either of principals

Tasks on Resources: 
- Remove owner & team  
- Addition of new principals on owner

ravisuhag:
@krtkvrm is this closed with PR?

krtkvrm:
yes, this should have been closed automatically
closing it now

PR #177, opened 2022-10-04
feat: schema migration & revision in ShieldDB for resources

krtkvrm:
closes: #176


rohilsurana:
Looks good, lets just get rid of the comment.

Issue #181, opened 2022-10-24
Add role flow to resource APIs


ishanarya0:
Resource APIs to be able to create relations while resource creation. 

Requirements:

1 - Request signature
```
{
    "name": "g-odpf-id-f-1",
    "namespace": "odpf/firehose",
    "project": "g-odpf-id",
    "relations" : [
        {
           role: "owner",
           member: "user:email@domain"
        }

    ]
}
```

2 - `Creator` of the resource shall be stored only as metadata in `user_id` column. We need not add the creator as the owner. If owner relations are not sent, project/org owners will still inherit owner rights on the resource.

3 - We will not fetch relations for both List and Get Resource APIs.

Tasks:

1 - Update Proton
2 - Add role flow to resource APIs

PR #182, opened 2022-10-26
feat: add relation creation in resource API

ishanarya0:
In favour of #181


internal/api/v1beta1/resource.go
```diff
@@ -82,6 +97,32 @@ func (h Handler) CreateResource(ctx context.Context, request *shieldv1beta1.Crea
...
+			continue
+		}
+
+		_, err := h.relationService.Create(ctx, relation.RelationV2{
+			Object: relation.Object{
+				ID:          newResource.Idxa,
+				NamespaceID: newResource.NamespaceID,
+			},
+			Subject: relation.Subject{
+				RoleID:    r.RoleName,
```
rohilsurana, line 114: role is something that is part of object definition. Any reason why we have it bundled with subject?

krtkvrm: Because `subject` is being assigned a `role`

rohilsurana (approved):

PR #185, opened 2022-11-17
feat: upgraded authz functionality

krtkvrm:
We only have authz check for Resource Group Namespaces


core/namespace/definition.go
```diff
@@ -13,8 +13,8 @@ var DefinitionProject = Namespace{
 }
 
 var DefinitionTeam = Namespace{
```
rohilsurana, line 15: SHouldn't this be `DefinitionGroup` ?

internal/api/v1beta1/permission_check.go
```diff
@@ -17,16 +17,16 @@ import (
...
-	if err := req.ValidateAll(); err != nil {
-		formattedErr := getValidationErrorMessage(err)
-		logger.Error(formattedErr.Error())
-		return nil, status.Errorf(codes.NotFound, formattedErr.Error())
-	}
+	//if err := req.ValidateAll(); err != nil {
+	//	formattedErr := getValidationErrorMessage(err)
+	//	logger.Error(formattedErr.Error())
+	//	return nil, status.Errorf(codes.NotFound, formattedErr.Error())
+	//}
```
rohilsurana, line 24: Need this ?

core/namespace/definition.go
krtkvrm: we can do this in another PR

PR #186, opened 2022-11-21
docs: add intro, glossary, server-setup-wip

ishanarya0:


docs/docs/concepts/architecture.md
```diff
@@ -27,4 +24,83 @@ Another DB instance is for SpiceDB to store all the data needed for authorizatio
 
 ### SpiceDB
 
-Shield push all the policies and relationships data to SpiceDB. All this data is needed to make the authorization decision. Shield connects to SpiceDB instance via gRPC
+Shield push all the policies and relationships data to SpiceDB. All this data is needed to make the authorization decision. Shield connects to SpiceDB instance via gRPC.
+
+## Overall System Architecture
```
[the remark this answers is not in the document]
AkarshSatija: Let's split the shieldProxy and shield so that we can show external check calls to shield

docs/docs/guides/adding-metadata-key.md
```diff
@@ -0,0 +1,34 @@
+import Tabs from '@theme/Tabs';
```
AkarshSatija, line 1: These don't seem to be rendering but getting included in the view.

docs/docs/guides/check-permission.md
```diff
@@ -1,22 +1,27 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+import CodeBlock from '@theme/CodeBlock';
+
+# Checking Pemrissions
+
+There are two ways to check a user permission on a resource in shield,
+## API Interface
+
+<Tabs groupId="api">
```
AkarshSatija, line 10: Need to look into the rendering properly.

docs/docs/guides/adding-metadata-key.md
ishanarya0: I think there is confusion here CMIIW, this page will have only one tab - `HTTP` since there is no `CLI` counterpart to it.

docs/docs/guides/check-permission.md
ishanarya0: This also has one `HTTP` tab only. Should we remove the tabs, where there is only one tab?

PR #197, opened 2023-01-10
feat: contextual logging for proxy

krtkvrm:
Adding Contextual Logging in proxy along with RequestID for every requests
Example:
```sh
2023-01-10T17:27:27.839+0530    info    request_forwarded       {"host": "localhost:5556", "path": "/api/ping", "method": "GET", "remote_address": "[::1]:58332", "scheme": "HTTP/1.1", "request_id": "ceul59q87d5m0a4ec78g", "origin": "RoundTrip", "backend": "localhost:3000"}
2023-01-10T17:27:27.844+0530    info    request_completed       {"host": "localhost:5556", "path": "/api/ping", "method": "GET", "remote_address": "[::1]:58332", "scheme": "HTTP/1.1", "request_id": "ceul59q87d5m0a4ec78g", "origin": "RoundTrip", "backend": "localhost:3000", "status": "200 OK"}
2023-01-10T17:27:40.791+0530    info    request_forwarded       {"host": "localhost:5556", "path": "/api/ping", "method": "GET", "remote_address": "[::1]:58332", "scheme": "HTTP/1.1", "request_id": "ceul5d287d5m0a4ec790", "origin": "RoundTrip", "backend": "localhost:3000"}
2023-01-10T17:27:40.796+0530    info    request_completed       {"host": "localhost:5556", "path": "/api/ping", "method": "GET", "remote_address": "[::1]:58332", "scheme": "HTTP/1.1", "request_id": "ceul5d287d5m0a4ec790", "origin": "RoundTrip", "backend": "localhost:3000", "status": "200 OK"}
```


internal/proxy/middleware/obeservability/observability.go
```diff
@@ -0,0 +1,75 @@
...
+		zap.String("method", req.Method),
+		zap.String("remote_address", req.RemoteAddr),
+		zap.String("scheme", req.Proto),
+		zap.String("request_id", reqID),
+	)
+	req = req.WithContext(ctx)
+	m.next.ServeHTTP(rw, req)
+}
+
+func GetIncomingRequestID(req *http.Request) string {
```
spy16, line 55: Do these functions need to be public?

internal/proxy/middleware/obeservability/observability.go
```diff
@@ -0,0 +1,75 @@
...
+	if reqID == "" {
+		reqID = xid.New().String()
+	}
+	req.Header.Set(headerRequestID, reqID)
+
+	return reqID
+}
+
+func SetRequestID(ctx context.Context, reqID string) context.Context {
+	return context.WithValue(ctx, requestIDContextKey, reqID)
```
spy16, line 66: using plain string value as context key is not recommended (it will be fine as long as we have very specific & unique key, but linters will complain)

internal/proxy/roundtripper.go
```diff
@@ -29,16 +29,18 @@ type h2cTransportWrapper struct {
 }
 
 func (t *h2cTransportWrapper) RoundTrip(req *http.Request) (*http.Response, error) {
+	logger := log.ZapFromContext(req.Context()).GetInternalZapLogger().Desugar().With(
+		zap.String("origin", "RoundTrip"),
```
spy16, line 33: why is origin RoundTrip here?

krtkvrm: origin is from where log was exected

internal/proxy/middleware/obeservability/observability.go
krtkvrm: this can be used from other middleware as well

internal/proxy/middleware/obeservability/observability.go
```diff
@@ -0,0 +1,75 @@
+package obeservability
```
mabdh, line 1:
```suggestion
package observability
```

internal/proxy/middleware/obeservability/observability.go
ishanarya0: +1
Let's define a struct for this
`type requestIDContextKey struct{}`

internal/proxy/middleware/obeservability/observability.go
ishanarya0: then let's move `GetIncomingRequestID` under `pkg` as a utility function?

internal/store/spicedb/schema_generator/generator.go
```diff
@@ -3,8 +3,8 @@ package schema_generator
 import (
 	"github.com/odpf/shield/internal/schema"
 
-	sdbcore "github.com/authzed/authzed-go/proto/authzed/api/v0"
 	sdbnamespace "github.com/authzed/spicedb/pkg/namespace"
+	sdbcore "github.com/authzed/spicedb/pkg/proto/core/v1"
```
ishanarya0, line 7: why do we need to change this package?

internal/proxy/middleware/obeservability/observability.go
spy16: no, the question is, is this function required outside this package? if not, make it private. otherwise, this is the right place for it. (You can also rename it to `RequestID(_)` by the way. the `GetIncoming` doesn't seem to provide any useful information.

(It's better to avoid `utilities`, `utils`, `common`, etc. as much as possible because they become dumping ground for random things).

internal/proxy/roundtripper.go
spy16: oh okay. i would suggest to look at how to configure automatic inclusion of source information in Zap. it will automatically add the go file name and line_no, function name, etc. to each log entry.

issue with this approach is, if someone changes the function name, it's not guaranteed that they will change the log field as well. once they drift, this log is going to cause more confusion.

internal/proxy/middleware/obeservability/observability.go
krtkvrm: done

internal/proxy/roundtripper.go
krtkvrm: ok, makes sense

internal/store/spicedb/schema_generator/generator.go
krtkvrm: due to spicedb upgrade, we need to upgrade proto as well

PR #204, opened 2023-02-10
fix: serialization inconsistency while creating a user

kushsharma:
When a user is created/updated it can have additional metadata, this data although it is a generic blob was getting serialized differently when the User was Created vs Updated which was causing behaviors like extra quotes if the metadata type is a string.


krtkvrm:
please change to target branch to `rc/0.5.1`

ishanarya0:
Hey @kushsharma I tried updating an user with the following body
image

Then I used the Get By Id API and it returns
image

Can you help in fixing this as well? 
You might want to test this step where we are converting the `Metadata` object to `*structpb.Struct` (https://github.com/odpf/shield/blob/44409f5183435b2b8bad8fab2960a866675831a0/internal/api/v1beta1/user.go#L374)

ishanarya0:
@kushsharma  Also, please consider these use case as well: 
```
An array of JSON

    "metadata": {
        "k1": [
            {
                "em": "vm"
            },
            {
                "pe": "cm"
            },
            {
                "pm": "gm"
            }
        ],
        "k2": "v2"
    }
```

```
Nested JSON

    "metadata": {
        "k1": {
                "em": "vm"
            },
        "k2": "v2"
    }
```

kushsharma:
Sure, let me check. I actually hoped the `Update` method was working fine and simply reused the implementation. `Create` and `Update` behave similarly but their serialization strategy is inconsistent.

ishanarya0:
> Sure, let me check. I actually hoped the `Update` method was working fine and simply reused the implementation. `Create` and `Update` behave similarly but their serialization strategy is inconsistent.

Thanks! 
FYI we support different behaviors while updating a user, based on if an uuid is passed or an email. For uuid, we are replacing the entire metadata, while with email we will merge old metadata with new.

ishanarya0:
Note: we need to migrate the server first before deploying these changes.

We are changing the type of `value` column in `metadata` table to `jsonb`.

Queries:

```
UPDATE metadata SET value = CONCAT('"',value,'"') where value NOT LIKE '[%' AND value NOT LIKE '{%';

ALTER TABLE metadata ALTER COLUMN value TYPE jsonb USING (value::jsonb);
```

cc: @krtkvrm

ishanarya0:
Done with the changes, you may have a look @kushsharma

spy16 (approved):

internal/store/postgres/migrations/20230223172233_alter_value_metadata_column.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE metadata ALTER COLUMN value TYPE varchar USING (value::varchar);
+
+UPDATE metadata SET value = SUBSTRING(value, 2, LENGTH(value) - 2) WHERE value LIKE '"%"' AND (value NOT LIKE '[%' AND value NOT LIKE '{%');
```
krtkvrm, line 3: add new line at end

PR #214, opened 2023-02-17
fix: use salt/mux pkg

ishanarya0:
Use salt/mux pkg to host http server and grpc server on separate ports

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,7 +6,6 @@ linters:
   enable-all: false
   disable-all: true
   enable:
-    - vet
     - goimports
     - thelper
     - unconvert
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,6 +6,7 @@ linters:
   enable-all: false
   disable-all: true
   enable:
+    - vet
     - goimports
     - thelper
     - unconvert
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -148,12 +148,7 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	}()
 
 	// serving server
-	err = server.Serve(ctx, logger, cfg.App, nrApp, deps)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return server.Serve(ctx, logger, cfg.App, nrApp, deps)
 }
 
 func buildAPIDependencies(
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -34,7 +34,6 @@ import (
 	"github.com/odpf/shield/pkg/db"
 
 	"github.com/odpf/salt/log"
-	salt_server "github.com/odpf/salt/server"
 	"github.com/pkg/profile"
 	"google.golang.org/grpc/codes"
 )
@@ -49,7 +48,7 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	}
 
 	// @TODO: need to inject custom logger wrapper over zap into ctx to use it internally
-	ctx, cancelFunc := context.WithCancel(salt_server.HandleSignals(context.Background()))
+	ctx, cancelFunc := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGINT)
 	defer cancelFunc()
 
 	dbClient, err := setupDB(cfg.DB, logger)
@@ -148,23 +147,10 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 		}
 	}()
 
-	keystrokeTermChan := make(chan os.Signal, 1)
-	// we'll accept graceful shutdowns when quit via SIGINT (Ctrl+C)
-	signal.Notify(keystrokeTermChan, os.Interrupt, os.Kill, syscall.SIGTERM)
-
 	// serving server
 	err = server.Serve(ctx, logger, cfg.App, nrApp, deps)
 	if err != nil {
 		return err
-	}
-	// wait for termination
-	select {
-	case <-ctx.Done():
-		fmt.Printf("process: ctx done bye\n")
-		break
-	case <-keystrokeTermChan:
-		fmt.Printf("process: kill signal received. bye \n")
-		break
 	}
 
 	return nil
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -153,15 +153,10 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	signal.Notify(keystrokeTermChan, os.Interrupt, os.Kill, syscall.SIGTERM)
 
 	// serving server
-	muxServer, err := server.Serve(ctx, logger, cfg.App, nrApp, deps)
+	err = server.Serve(ctx, logger, cfg.App, nrApp, deps)
 	if err != nil {
 		return err
 	}
-	defer func() {
-		logger.Info("cleaning up server")
-		server.Cleanup(ctx, muxServer)
-	}()
-
 	// wait for termination
 	select {
 	case <-ctx.Done():
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -48,7 +48,7 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	}
 
 	// @TODO: need to inject custom logger wrapper over zap into ctx to use it internally
-	ctx, cancelFunc := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGINT)
+	ctx, cancelFunc := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
 	defer cancelFunc()
 
 	dbClient, err := setupDB(cfg.DB, logger)
--- a/internal/api/v1beta1/v1beta1.go
+++ b/internal/api/v1beta1/v1beta1.go
@@ -3,9 +3,10 @@ package v1beta1
 import (
 	"context"
 
-	"github.com/odpf/salt/server"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/odpf/shield/internal/api"
 	shieldv1beta1 "github.com/odpf/shield/proto/v1beta1"
+	"google.golang.org/grpc"
 )
 
 type Handler struct {
@@ -23,8 +24,8 @@ type Handler struct {
 	ruleService      RuleService
 }
 
-func Register(ctx context.Context, s *server.MuxServer, gw *server.GRPCGateway, deps api.Deps) {
-	gw.RegisterHandler(ctx, shieldv1beta1.RegisterShieldServiceHandlerFromEndpoint)
+func Register(ctx context.Context, address string, s *grpc.Server, gw *runtime.ServeMux, deps api.Deps) {
+	shieldv1beta1.RegisterShieldServiceHandlerFromEndpoint(ctx, gw, address, []grpc.DialOption{grpc.WithInsecure()})
 
 	s.RegisterService(
 		&shieldv1beta1.ShieldService_ServiceDesc,
--- a/internal/api/v1beta1/v1beta1.go
+++ b/internal/api/v1beta1/v1beta1.go
@@ -3,11 +3,9 @@ package v1beta1
 import (
 	"context"
 
-	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/odpf/shield/internal/api"
 	shieldv1beta1 "github.com/odpf/shield/proto/v1beta1"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
 )
 
 type Handler struct {
@@ -25,12 +23,7 @@ type Handler struct {
 	ruleService      RuleService
 }
 
-func Register(ctx context.Context, address string, s *grpc.Server, gw *runtime.ServeMux, deps api.Deps) error {
-	err := shieldv1beta1.RegisterShieldServiceHandlerFromEndpoint(ctx, gw, address, []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())})
-	if err != nil {
-		return err
-	}
-
+func Register(ctx context.Context, s *grpc.Server, deps api.Deps) error {
 	s.RegisterService(
 		&shieldv1beta1.ShieldService_ServiceDesc,
 		&Handler{
--- a/internal/api/v1beta1/v1beta1.go
+++ b/internal/api/v1beta1/v1beta1.go
@@ -24,8 +24,11 @@ type Handler struct {
 	ruleService      RuleService
 }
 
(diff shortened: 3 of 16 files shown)


internal/server/server.go
```diff
@@ -24,27 +25,58 @@ import (
...
+	}))
+
+	grpcGateway := runtime.NewServeMux(runtime.WithIncomingHeaderMatcher(customHeaderMatcherFunc(map[string]bool{cfg.IdentityProxyHeader: true})))
+	httpMux.Handle("/admin/", http.StripPrefix("/admin", grpcGateway))
+	grpcServer := grpc.NewServer(getGRPCMiddleware(cfg, logger, nrApp))
+	address := fmt.Sprintf("%s:%d", cfg.Host, cfg.Port)
+
+	v1beta1.Register(ctx, address, grpcServer, grpcGateway, deps)
+
+	go mux.Serve(
```
sudo-suhas, line 66: Do we need to run this as a goroutine?

cmd/serve.go
```diff
@@ -153,15 +153,10 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	signal.Notify(keystrokeTermChan, os.Interrupt, os.Kill, syscall.SIGTERM)
```
sudo-suhas, line 153: It would be better to just create a context that would be cancelled on receiving the signal. See https://github.com/odpf/compass/blob/0ab27e9/cli/serve.go#L55

cmd/serve.go
```diff
@@ -153,15 +153,10 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
...
 	if err != nil {
 		return err
 	}
-	defer func() {
-		logger.Info("cleaning up server")
-		server.Cleanup(ctx, muxServer)
-	}()
-
 	// wait for termination
 	select {
```
sudo-suhas, line 161: Won't be needed since `Serve` will only return once it has been shutdown.

cmd/serve.go
ishanarya0: I think this is handled here, CMIIW https://github.com/odpf/shield/blob/44409f5183435b2b8bad8fab2960a866675831a0/cmd/serve.go#L52

internal/server/server.go
ishanarya0: I'll move it out of the go routine, as we discussed.

cmd/serve.go
sudo-suhas: In that case, can we simplify it by using `signal.NotifyContext(...)`?

ishanarya0: This is done

pkg/body_extractor/grpc_payload.go
```diff
@@ -35,24 +35,6 @@ const (
 	Message      = "Message"
 )
 
-type queryCacheMutex struct {
```
ishanarya0, line 38: removed since it was unused, and linter was throwing error

ishanarya0:
Test scenarios covered:
1: Made request at HTTP port
2: Made request and gRPC port
3: SIGINT signal closes the server gracefully

.golangci.yml
```diff
@@ -6,7 +6,6 @@ linters:
   enable-all: false
   disable-all: true
   enable:
-    - vet
```
sudo-suhas, line 9: Why is this being disabled?

internal/api/v1beta1/v1beta1.go
```diff
@@ -23,8 +24,11 @@ type Handler struct {
 	ruleService      RuleService
 }
 
-func Register(ctx context.Context, s *server.MuxServer, gw *server.GRPCGateway, deps api.Deps) {
-	gw.RegisterHandler(ctx, shieldv1beta1.RegisterShieldServiceHandlerFromEndpoint)
+func Register(ctx context.Context, address string, s *grpc.Server, gw *runtime.ServeMux, deps api.Deps) error {
+	err := shieldv1beta1.RegisterShieldServiceHandlerFromEndpoint(ctx, gw, address, []grpc.DialOption{grpc.WithInsecure()})
```
sudo-suhas, line 28: Can we avoid using deprecated API (`grpc.WithInsecure()`)?

cmd/serve.go
```diff
@@ -49,7 +48,7 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
 	}
 
 	// @TODO: need to inject custom logger wrapper over zap into ctx to use it internally
-	ctx, cancelFunc := context.WithCancel(salt_server.HandleSignals(context.Background()))
+	ctx, cancelFunc := signal.NotifyContext(context.Background(), os.Interrupt, os.Kill, syscall.SIGTERM, syscall.SIGINT)
```
spy16, line 51: os.Interrupt is same as syscall.SIGINT   
os.Kill is same as syscall.SIGKILL
(Refer https://pkg.go.dev/os#Signal)

Do not trap SIGKILL signal. only trap SIGTERM and SIGINT

ishanarya0: My understanding of this: If we would trap SIGKILL, it will kill the process and hence the server will not shut gracefully. CMIIW.

internal/server/server.go
```diff
@@ -14,28 +15,29 @@ import (
...
 
-func registerHandler(ctx context.Context, s *server.MuxServer, gw *server.GRPCGateway, deps api.Deps) {
-	s.RegisterHandler("/admin*", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+func registerHandler(ctx context.Context, sh *http.ServeMux, gw *runtime.ServeMux, sg *grpc.Server, deps api.Deps, address string) error {
+	sh.Handle("/admin*", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprintf(w, "pong")
 	}))
 
-	s.RegisterHandler("/admin/ping", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	sh.Handle("/admin/ping", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
sudo-suhas, line 34: Why do we need to mount these routes manually? We could pass the `runtime.WithHealthEndpointAt` option in the `runtime.NewServeMux` constructor.

internal/server/server.go
```diff
@@ -44,35 +46,40 @@ func Serve(
...
+			Handler:        httpMux,
+			ReadTimeout:    120 * time.Second,
+			WriteTimeout:   120 * time.Second,
+			MaxHeaderBytes: 1 << 20,
+		}),
+		mux.WithGRPCTarget(fmt.Sprintf(":%d", cfg.GRPCPort), grpcServer),
+		mux.WithGracePeriod(5*time.Second),
+	); !errors.Is(err, context.Canceled) {
+		logger.Error("mux serve error", "err", err)
+		return err
```
sudo-suhas, line 77: Any reason for both logging and returning the error? If the only intention for returning the error is logging, it is already taken care of here.

internal/server/config.go
```diff
@@ -1,8 +1,12 @@
 package server
 
 type Config struct {
```
sudo-suhas, line 3: Can you make changes to [`config/config.yaml`](https://github.com/odpf/shield/blob/rc/0.5.1/config/config.yaml) as well?

.golangci.yml
ishanarya0: my bad, vet was throwing ` Error: loopclosure: loop variable tt captured by func literal (govet)` for some test files.
Should I add `//nolint:govet` to those files for now and look at this in a separate issue?

internal/server/server.go
ishanarya0: Changed this
`[]grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}`

internal/api/v1beta1/v1beta1.go
ishanarya0: Changed it
`[]grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}`

.golangci.yml
sudo-suhas: Can you try to see if it is a genuine issue? If not, we can add an exclusion in the golangci-lint configuration.

cmd/serve.go
spy16: Trapping sigkill is pointless for 2 reasons:

1. SIGKILL is usually issued by us or the system when the process is not responding for some reason. Under those circumstances, it's fair to assume that the graceful shutdown logic will also may not work as expected.
2. On issuing a sigkill, kernel will kill the process non-gracefully anyway (there is no guarantee that the graceful shutdown logic will be allowed to execute)

internal/server/server.go
ishanarya0: `runtime.WithHealthEndpointAt` accepts `grpc_health_v1.HealthClient` which needs `grpc.ClientConnInterface` to initialize. In our current implementation we are NOT making a connection ourself [like compass](https://github.com/odpf/compass/blob/7fc4a8236562d5d73f1253950a9d074c3c61fd82/internal/server/server.go#L111) do. Rather we are using [RegisterShieldServiceHandlerFromEndpoint](https://github.com/odpf/shield/blob/3081952afa0d76e0a4188bf90efee52624d72380/internal/api/v1beta1/v1beta1.go#L28)`.

Should we refactor to make a conn ourselves? Does it have some benefit?

.golangci.yml
spy16: that warning might be coming for a good reason. if your tests are running parallely, re-using loop variable can cause issue.

internal/server/server.go
sudo-suhas: Lets refactor, [`RegisterShieldServiceHandlerFromEndpoint`](https://github.com/odpf/shield/blob/3081952afa0d76e0a4188bf90efee52624d72380/internal/api/v1beta1/v1beta1.go#L28) is anyway creating a connection under the hood.

Not needing to mount the ping endpoint is a win and if even if there is some change needed in the response for health checkpoint, using the functionality provided by gRPC gateway's runtime will take care of it. Also, try to go through the implementation for the health endpoint provided by them. That might answer your question w.r.t the benefits.

internal/server/server.go
ishanarya0: Yup, my bad, returned with a `nil` error at this step.

.golangci.yml
ishanarya0: Fixed this, thanks!

internal/server/server.go
```diff
@@ -100,12 +121,3 @@ func getGRPCMiddleware(cfg Config, logger log.Logger, nrApp newrelic.Application
 			nrgrpc.UnaryServerInterceptor(nrApp),
 		))
 }
-
-func customHeaderMatcherFunc(headerKeys map[string]bool) func(key string) (string, bool) {
```
ishanarya0, line 104: this is unused. linter was throwing error for this.

pkg/body_extractor/grpc_payload_test.go
```diff
@@ -120,7 +119,7 @@ func TestExtract(t *testing.T) {
 					"ten",
 				},
 			},
-			query: "1.2.7[*]",
```
ishanarya0, line 123: This test had a bug in query

internal/proxy/hook/authz/authz_test.go
```diff
@@ -37,18 +37,16 @@ var expectedResources = []resource.Resource{
...
 	}, {
 		ProjectID:      "ab657ae7-8c9e-45eb-9862-dd9ceb6d5c71",
 		OrganizationID: "org1",
 		Name:           "resc2",
-		NamespaceID:    "ns1_kind",
+		NamespaceID:    "ns1/kind",
 	},
 }
 
 func TestCreateResources(t *testing.T) {
```
sudo-suhas, line 49: Please recheck why these changes are necessary, don't seem relevant for the PR.

internal/proxy/hook/authz/authz_test.go
```diff
@@ -109,8 +95,6 @@ func TestCreateResources(t *testing.T) {
 
 	for _, tt := range table {
 		t.Run(tt.title, func(t *testing.T) {
```
sudo-suhas, line 97: The right way to fix this test for it to be run in parallel would be to redeclare `tt` inside the for loop scope:

```go
for _, tt := range table {
	tt := tt
	t.Run(tt.title, func(t *testing.T) {
// ...
```

internal/server/health/health.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/status"
+)
+
+type HealthHandler struct{}
+
+func (h *HealthHandler) Check(ctx context.Context, in *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
```
sudo-suhas, line 13: We can omit the parameter field names since we don't use them.

```suggestion
func (h *HealthHandler) Check(context.Context, *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
```

internal/server/health/health.go
```diff
@@ -0,0 +1,24 @@
...
+	return &grpc_health_v1.HealthCheckResponse{Status: grpc_health_v1.HealthCheckResponse_SERVING}, nil
+}
+
+func (h *HealthHandler) Watch(in *grpc_health_v1.HealthCheckRequest, _ grpc_health_v1.Health_WatchServer) error {
```
sudo-suhas, line 17:
```suggestion
func (h *HealthHandler) Watch(*grpc_health_v1.HealthCheckRequest, grpc_health_v1.Health_WatchServer) error {
```

internal/server/server.go
```diff
@@ -14,65 +15,85 @@ import (
...
+			grpc.MaxCallSendMsgSize(cfg.GRPC.MaxSendMsgSize),
+		))
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	gw, err := server.NewGateway("", cfg.Port, server.WithGatewayMuxOptions(
-		runtime.WithIncomingHeaderMatcher(customHeaderMatcherFunc(map[string]bool{cfg.IdentityProxyHeader: true}))),
+	grpcGateway := runtime.NewServeMux(
```
sudo-suhas, line 57: CMIIW, but you need to add `runtime.WithIncomingHeaderMatcher(customHeaderMatcherFunc(map[string]bool{cfg.IdentityProxyHeader: true}))` here to retain the current behaviour.

internal/proxy/hook/authz/authz_test.go
ishanarya0: Yes, these are not for this problem. These tests had this wrong configuration, due to which they are failing. Not sure how they were working earlier in parallel mode.

internal/server/server.go
ishanarya0: Oh my mistake, I accidentally removed this. thanks!

sudo-suhas (approved):

PR #218, opened 2023-04-10
feat(authentication): add support to persist sessions and flows in database

Chief-Rishab:
- [x] Add support to save user sessions and flows in the postgres database
- [x] Documentation 
   - Update API and CLI reference
   - Update server configurations

### DIFF
--- a/core/authenticate/authenticate.go
+++ b/core/authenticate/authenticate.go
@@ -19,6 +19,7 @@ func (m AuthMethod) String() string {
 // Flow is a temporary state used to finish login/registration flows
 type Flow struct {
 	ID uuid.UUID
+
 	// authentication flow type
 	Method string
 
@@ -26,6 +27,7 @@ type Flow struct {
 	StartURL string
 	// FinishURL is where flow should end to after successful verification
 	FinishURL string
+
 	// Nonce is a once time use random string
 	Nonce string
 
--- a/core/authenticate/authenticate.go
+++ b/core/authenticate/authenticate.go
@@ -18,36 +18,36 @@ func (m AuthMethod) String() string {
 
 // Flow is a temporary state used to finish login/registration flows
 type Flow struct {
-	ID uuid.UUID
+	ID uuid.UUID `db:"id"`
 
 	// authentication flow type
-	Method string
+	Method string `db:"method"`
 
 	// StartURL is where flow should start from for verification
-	StartURL string
+	StartURL string `db:"start_url"`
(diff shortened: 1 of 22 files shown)


kushsharma:
@Chief-Rishab can you rebase this with main, I'll review it after

Chief-Rishab:
@kushsharma ready for review

core/authenticate/authenticate.go
```diff
@@ -18,36 +18,36 @@ func (m AuthMethod) String() string {
 
 // Flow is a temporary state used to finish login/registration flows
 type Flow struct {
-	ID uuid.UUID
+	ID uuid.UUID `db:"id"`
```
kushsharma, line 21: Although we can use these structs, normally in our postgres implementation we have created a database-specific struct schema for each entity and then they get translated to app-level domain models. This is done to keep them loosely coupled. Check repositories of group or project for reference.

kushsharma (changes requested):

docs/docs/tour/setup-server.md
```diff
@@ -75,17 +91,18 @@ entropy:
         - name: view
           roles:
             - owner
-            - organization/owner
```
kushsharma, line 78: we can change it to `organization:owner`

internal/store/postgres/migrations/20230410103624_create_flows_table.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS flows(
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    method varchar,
```
kushsharma, line 3: instead of `varchar` can we use `text`

Chief-Rishab: isn't `VARCHAR` and `TEXT` same if we don't specify the number to limit the characters?

https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-char-varchar-text/#:~:text=Different%20from%20other%20database%20systems,to%20check%20for%20the%20length.

kushsharma: Yes they are same, I thought in rest of the migrations we are using `text` so I thought we may want to keep it consistent, I just checked looks like we actually use `varchar`. Should be fine then.

kushsharma (approved):

PR #220, opened 2023-04-12
feat: add a cron job to remove expired sessions from database

Chief-Rishab:


cmd/serve.go
```diff
@@ -197,8 +206,8 @@ func buildAPIDependencies(
 		userService,
 		projectService)
 
-	sessionService := authenticate.NewSessionManager(postgres.NewSessionRepository(), grpc_interceptors.SessionValidity)
-	registrationService := authenticate.NewRegistrationService(postgres.NewFlowRepository(), userService, cfg.App.Authentication)
+	sessionService := authenticate.NewSessionManager(postgres.NewSessionRepository(dbc), grpc_interceptors.SessionValidity, logger)
```
kushsharma, line 209: nit: move the logger arg to first argument.

internal/store/postgres/session_repository.go
```diff
@@ -1,34 +1,175 @@
...
+
+		return authenticate.ErrDeletingSession
+	})
+}
+
+func (s *SessionRepository) DeleteExpiredSessions(ctx context.Context, logger log.Logger) error {
+	query, params, err := dialect.Delete(TABLE_SESSIONS).
+		Where(
+			goqu.Ex{
+				"expires_at": goqu.Op{"lt": goqu.L("NOW()")},
```
kushsharma, line 145: I think its better if we pass our expiry time manually if in case db timezone is not matching the application's timezone.

kushsharma:
We should also clean up `flows`, say older than 3 days?

kushsharma (changes requested):

cmd/serve.go
kushsharma: Also can you check if logger is part of context already? Using `logger := grpczap.Extract(ctx)`?

PR #223, opened 2023-04-15
feat: add a cron job to remove expired sessions from database 

Chief-Rishab:
- [x] Adding cron jobs that will remove expired sessions and flows from the db
- [x] Implement Refresh function to extend session expiry time

### DIFF
--- a/core/authenticate/session/service.go
+++ b/core/authenticate/session/service.go
@@ -24,6 +24,7 @@ type Repository interface {
 	Get(ctx context.Context, id uuid.UUID) (*Session, error)
 	Delete(ctx context.Context, id uuid.UUID) error
 	DeleteExpiredSessions(ctx context.Context, logger log.Logger, expiryTime time.Time) error
+	UpdateValidity(ctx context.Context, id uuid.UUID, validity time.Duration) error
 }
 
 type Service struct {
@@ -58,9 +59,8 @@ func (s Service) Create(ctx context.Context, userID string) (*Session, error) {
 }
 
 // Refresh extends validity of session
-func (s Service) Refresh(session *Session) error {
-	// TODO(kushsharma)
-	panic("not implemented")
+func (s Service) Refresh(ctx context.Context, sessionID uuid.UUID) error {
+	return s.repo.UpdateValidity(ctx, sessionID, s.validity)
 }
 
 func (s Service) Delete(ctx context.Context, sessionID uuid.UUID) error {
--- a/core/authenticate/session/service.go
+++ b/core/authenticate/session/service.go
@@ -23,7 +23,7 @@ type Repository interface {
 	Set(ctx context.Context, session *Session) error
 	Get(ctx context.Context, id uuid.UUID) (*Session, error)
 	Delete(ctx context.Context, id uuid.UUID) error
-	DeleteExpiredSessions(ctx context.Context, logger log.Logger) error
+	DeleteExpiredSessions(ctx context.Context) error
 	UpdateValidity(ctx context.Context, id uuid.UUID, validity time.Duration) error
 }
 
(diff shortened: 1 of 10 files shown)


kushsharma:
We should clean up expired flows as well, flow expiry could be treated as 7 days.

core/authenticate/session/service.go
```diff
@@ -8,30 +8,38 @@ import (
...
 	ErrNoSession       = errors.New("no session")
 	ErrDeletingSession = errors.New("error deleting session")
+	refreshTime        = "0 0 * * *" // Once a day at midnight (UTC)
 )
 
 type Repository interface {
 	Set(ctx context.Context, session *Session) error
 	Get(ctx context.Context, id uuid.UUID) (*Session, error)
 	Delete(ctx context.Context, id uuid.UUID) error
+	DeleteExpiredSessions(ctx context.Context, logger log.Logger, expiryTime time.Time) error
```
kushsharma, line 26: This method doesn't look appropriate to me. Session already have an expiry time and using this method we are passing time of `now` which tells it to expire all sessions past `now`. I don't think we would ever call this method without `now` time so making it an argument doesn't look good.

I feel it's better if the repository itself has a `Now` method to get current time and simply delete all sessions older than expiry time. 

Same in `flows` as well. Either we can add another column in flow with expiry_at just like session or we can initialize our postgres repository with validity duration. Then Cleanup method will simply do a Delete call where `created_at < now - validity`.

kushsharma (changes requested):

core/authenticate/session/service.go
Chief-Rishab: @kushsharma you pointed this out last time, that its better if we pass our expiry time manually if in case db timezone is not matching the application's timezone. That's why this expiryTime argument was added

internal/store/postgres/flow_repository.go
```diff
@@ -126,3 +128,40 @@ func (s *FlowRepository) Delete(ctx context.Context, id uuid.UUID) error {
 		return fmt.Errorf("no entry to delete")
 	})
 }
+
+func (s *FlowRepository) DeleteExpiredFlows(ctx context.Context, logger log.Logger) error {
+	expiryTime := time.Now().UTC().AddDate(0, 0, -7)
```
kushsharma, line 133: I think we should use a `Now` function to get the current time instead of `time.Now()` else we won't be able to mock the time to write unit tests. Check the session service for reference.

internal/store/postgres/flow_repository.go
```diff
@@ -126,3 +128,40 @@ func (s *FlowRepository) Delete(ctx context.Context, id uuid.UUID) error {
...
+		}
+
+		result, err := s.dbc.ExecContext(ctx, query, params...)
+		if err != nil {
+			err = checkPostgresError(err)
+			return fmt.Errorf("%w: %s", dbErr, err)
+		}
+
+		count, _ := result.RowsAffected()
+		logger.Info("deleted expired flows", "expired_flows_count", count)
```
kushsharma, line 163: Make this `debug`

internal/store/postgres/session_repository.go
```diff
@@ -140,3 +142,77 @@ func (s *SessionRepository) Delete(ctx context.Context, id uuid.UUID) error {
...
+		}
+
+		result, err := s.dbc.ExecContext(ctx, query, params...)
+		if err != nil {
+			err = checkPostgresError(err)
+			return fmt.Errorf("%w: %s", dbErr, err)
+		}
+
+		count, _ := result.RowsAffected()
+		logger.Info("deleted expired sessions", "expired_session_count", count)
```
kushsharma, line 176: `debug`

kushsharma (changes requested):

kushsharma (approved):
LGTM

PR #227, opened 2023-04-21
Admin UI for shield

pyadav:


internal/server/server.go
```diff
@@ -90,8 +92,16 @@ func Serve(
 	if err != nil {
 		return err
 	}
+
 	httpMux.Handle("/jwks.json", jwksHandler)
 
+	spaHandler, err := spa.Handler(ui.Assets, "dist/ui", "index.html", false)
+	if err != nil {
+		fmt.Println("Failed to load spa:", err)
```
kushsharma, line 100: Can we use an existing logger struct with `warn` tag?

ui/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: dist test
+
+dist:
```
kushsharma, line 3: Should we also add a small README.md in `ui` directory? Listing what steps to run for running/building ui?

Makefile
```diff
@@ -1,15 +1,19 @@
 GOVERSION := $(shell go version | cut -d ' ' -f 3 | cut -d '.' -f 2)
 
-.PHONY: build check fmt lint test test-race vet test-cover-html help install proto
+.PHONY: build check fmt lint test test-race vet test-cover-html help install proto ui
 .DEFAULT_GOAL := build
 PROTON_COMMIT := "056db0eef064bc5f82343e6a5a312567a39d9ec6"
 
+ui:
```
kushsharma, line 7: We should also ignore `ui` folder from various `test` commands

ui/vite.config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  build: {
+    outDir: "dist/ui",
+  },
+  server: {
+    proxy: {
+      "/admin": {
+        target: "http://localhost:8000",
```
kushsharma, line 13: Why is this hardcoded? Is it possible to put it in some .env?

internal/server/server.go
pyadav: added `logger.Warn`

ui/Makefile
pyadav: Added

Makefile
pyadav: just added into make test, rest of them are using specific folder to run test

ui/vite.config.ts
pyadav: Fixed

kushsharma (approved):

PR #229, opened 2023-04-23
feat: add member role in org with only membership permission

kushsharma:


internal/store/spicedb/relation_repository.go
```diff
@@ -102,15 +107,21 @@ func (r RelationRepository) AddV2(ctx context.Context, rel relation.RelationV2)
 }
 
 func (r RelationRepository) Check(ctx context.Context, rel relation.Relation, act action.Action) (bool, error) {
+	consistency := &authzedpb.Consistency{}
+	if r.fullyConsistent {
```
pyadav, line 111: Do we need if condition here as default bool value will be false.
Is there any performance hit event if we are passing false?

internal/store/spicedb/relation_repository.go
```diff
@@ -225,7 +236,13 @@ func (r RelationRepository) DeleteSubjectRelations(ctx context.Context, resource
 }
 
 func (r RelationRepository) LookupSubjects(ctx context.Context, rel relation.RelationV2) ([]string, error) {
+	consistency := &authzedpb.Consistency{}
+	if r.fullyConsistent {
+		consistency = &authzedpb.Consistency{Requirement: &authzedpb.Consistency_FullyConsistent{FullyConsistent: true}}
+	}
+
```
pyadav, line 243: Same as above

internal/store/spicedb/relation_repository.go
kushsharma: Good idea, refactored it as `r.getConsistency()` as I would still want to fallback to default behaviour

```go
func (r RelationRepository) getConsistency() *authzedpb.Consistency {
	if !r.fullyConsistent {
		return &authzedpb.Consistency{}
	}
	return &authzedpb.Consistency{Requirement: &authzedpb.Consistency_FullyConsistent{FullyConsistent: true}}
}
```

Issue #209, opened 2023-02-13
Questions related to design


kushsharma:
I am going through the shield to understand the current design, I have a couple of questions related to it:
1. APIs for creating a `user` accept [User model](https://github.com/odpf/proton/blob/33fc2ccc89e177d27058faa865b041b03e712d32/odpf/shield/v1beta1/shield.proto#L551) that has `slug` defined but the [User model](https://github.com/odpf/shield/blob/3f0a2e7e1b2df9bc9a98ddc7c2c21ee8b4faac60/core/user/user.go) for shield doesn't have it, nor does it persist in the database? Do you know if this is a bug?
2. The validation rules defined in [protobuf](https://github.com/odpf/proton/blob/33fc2ccc89e177d27058faa865b041b03e712d32/odpf/shield/v1beta1/shield.proto) are not really used to validate anywhere in the shield, is this by design?
3. If I go with validation rules defined for the `name` of the `User`, it doesn't accept spaces. My understanding of the name was to use something like `Kush Sharma`, is this incorrect?
4. For `users` I see we are using a dedicated table to manage `metadata` and have [dropped](https://github.com/odpf/shield/blob/main/internal/store/postgres/migrations/20220822135848_remove_metadata_users_column.up.sql) `metadata` column from `users` table but this is not the same for rest of the models like `org` or `group`. Is this because we wanted to query metadata via key in `users`?
5. In a recent commit on [protos](https://github.com/odpf/proton/pull/240) a lot of useful APIs were removed which were used for shield management, are we replacing them with something else? For example, how am I supposed to add `admins` to the `group` now?
6. I don't see a direct 1-to-1 mapping of a `user` to an `organization`, `user` gets added to a `group` and the `group` is attached to an `org`. Does that mean before creating `users`, it is mandatory to create a `group`?
7. I see a `relations` table being introduced in recent commits, is the way moving forward to manage 1-to-1 or 1-to-many relationships? Even for a `group` to org? It's managed via a column in the `group` table currently.
8. I went through the documentation and I got the gist of how the `proxy` is working but the `authz` part was lacking some critical explanations like how is `resources_config` shown in [Tour](https://odpf.github.io/shield/tour/setup-server) is really built. How am I supposed to know how to model a `system` type or a `resource_group`? How do they really fit? Can we explain it via a small use case in the docs? I tried using the config directly as it is and it throws a bunch of errors, is the config outdated?
[...]

PR #230, opened 2023-04-24
fix!: add user slug and add validator

Chief-Rishab:
- https://github.com/odpf/proton/pull/258

As per discussions in #209 . APIs for creating a user accept [User model](https://github.com/odpf/proton/blob/33fc2ccc89e177d27058faa865b041b03e712d32/odpf/shield/v1beta1/shield.proto#L551) that has slug defined but the [User model](https://github.com/odpf/shield/blob/3f0a2e7e1b2df9bc9a98ddc7c2c21ee8b4faac60/core/user/user.go) for shield doesn't have it, nor does it persist in the database.

In this PR, we are adding the missing slug in the Users model and db, also adding support to Get and Update User by slug.
 
Additionally, the http request body doesn't get validated against the proto messages validation rules. We are adding a grpc validator middleware for the same 

Note: **Breaking Changes** in existing APIs:
- Updating user email not possible
- Allow get user with email and slug


cmd/user.go
```diff
@@ -64,6 +65,10 @@ func createUserCommand(cliConfig *Config) *cli.Command {
 				return err
 			}
 
+			if reqBody.Slug == "" {
+				reqBody.Slug = str.GenerateUserSlug(reqBody.Name)
```
kushsharma, line 69: We should use user email

internal/api/v1beta1/user.go
```diff
@@ -90,6 +92,12 @@ func (h Handler) CreateUser(ctx context.Context, request *shieldv1beta1.CreateUs
 		email = currentUserEmail
 	}
 
+	name := request.GetBody().GetName()
+	slug := strings.TrimSpace(request.GetBody().GetSlug())
+	if slug == "" {
+		slug = str.GenerateUserSlug(name)
```
kushsharma, line 98: Use email instead

internal/store/postgres/user_repository.go
```diff
@@ -144,9 +145,110 @@ func (r UserRepository) GetByID(ctx context.Context, id string) (user.User, erro
...
+	}
+	transformedUser.Metadata = data
+
+	return transformedUser, nil
+}
+
 func (r UserRepository) Create(ctx context.Context, usr user.User) (user.User, error) {
-	if strings.TrimSpace(usr.Email) == "" {
-		return user.User{}, user.ErrInvalidEmail
+	if strings.TrimSpace(usr.Name) == "" || strings.TrimSpace(usr.Email) == "" || strings.TrimSpace(usr.Slug) == "" {
```
kushsharma, line 250: I think `Name` should be okay to be empty.

test/e2e/regression/api_test.go
```diff
@@ -184,7 +186,7 @@ func (s *APIRegressionTestSuite) TestProjectAPI() {
 		_, err := s.testBench.Client.UpdateProject(ctxOrgAdminAuth, &shieldv1beta1.UpdateProjectRequest{
 			Id: "random",
 			Body: &shieldv1beta1.ProjectRequestBody{
-				Name:  "new project",
+				Name:  "new-project",
```
kushsharma, line 189: Name can contain spaces.

kushsharma (changes requested):

internal/api/v1beta1/user.go
```diff
@@ -235,28 +245,9 @@ func (h Handler) UpdateUser(ctx context.Context, request *shieldv1beta1.UpdateUs
 	}
 
 	id := request.GetId()
-	if shielduuid.IsValid(id) {
```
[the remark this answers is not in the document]
Chief-Rishab: There are multiple methods to update the user in `user/service`, `UpdateByID` (for updating by id or slug) and an `UpdateByEmail` function. So this check is used to call the appropriate function based on id which can be email/slug/uuid

test/e2e/regression/api_test.go
Chief-Rishab: Should I update the proto for the same. project, group and organisation names too have the same regex. "^[A-Za-z0-9_-]+$" https://github.com/odpf/proton/blob/main/odpf/shield/v1beta1/shield.proto#L943

internal/api/v1beta1/user.go
Chief-Rishab: yes will update the slug generation strategy

internal/api/v1beta1/user.go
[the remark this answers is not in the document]
Chief-Rishab: cleaned it

test/e2e/regression/api_test.go
```diff
@@ -120,7 +122,7 @@ func (s *APIRegressionTestSuite) TestOrganizationAPI() {
 	s.Run("2. user attached to an org as member should have no basic permission other than membership", func() {
 		createOrgResp, err := s.testBench.Client.CreateOrganization(ctxOrgAdminAuth, &shieldv1beta1.CreateOrganizationRequest{
 			Body: &shieldv1beta1.OrganizationRequestBody{
-				Name: "org acme 2",
+				Name: "org-acme-2",
```
kushsharma, line 125: Any reason we added hyphen in name instead of space?

Chief-Rishab: Will restore back to original since we have the proto file changed. Earlier name regex didn't support spaces

Chief-Rishab: done

test/e2e/regression/api_test.go
```diff
@@ -118,7 +118,7 @@ func (s *APIRegressionTestSuite) TestOrganizationAPI() {
 		userResp, err := s.testBench.Client.CreateUser(ctxOrgAdminAuth, &shieldv1beta1.CreateUserRequest{Body: &shieldv1beta1.UserRequestBody{
 			Name:  "acme 2 member",
 			Email: "acme-member@odpf.io",
-			Slug:  "acme-2-member",
+			Slug:  "acme_2_member",
```
kushsharma, line 121: Apologies for noticing this late but we should also allow hyphens in slug. I think I'll merge this PR and you can raise another one with just this change and proton as well. Does that make sense?

Chief-Rishab: yes. Also one more change from your recent commit in main. Object namespace regex needs to have a `/`. Otherwise 
`shield/organisation` was failing according to the validator we have added in this PR.

Will add both of these changes in the next proto https://github.com/odpf/proton/pull/262/files

kushsharma: Let me update the user slug regex, you can then simply fix this PR only.

kushsharma: Cool, will merge this PR then

kushsharma (approved):

PR #232, opened 2023-04-28
feat!: add json schema validation for metadata and deprecate metadata keys

Chief-Rishab:
- require:
- https://github.com/odpf/proton/pull/257
- https://github.com/odpf/proton/pull/266

Metadata in user/organization/group/roles will be validated against the metadata json-schemas(`metaschema`). Default metaschemas will be populated in the db and can be modified utilising the metaschemas apis endpoint ( `meta/schemas`)  

For more details, documentation for managing metaschemas are added in Guides pages.

**BREAKING CHANGE!:** : Metadata keys currently used for validating user's metadata is removed in this PR (also create metadatakeys api will also be deprecated)


internal/api/v1beta1/metaschema.go
```diff
@@ -0,0 +1,151 @@
...
+	// todo  validate name in proto
+	name := request.GetName()
+
+	if request.GetBody() == nil {
+		return nil, grpcBadBodyError
+	}
+
+	if name != request.GetBody().GetName() {
+		return nil, status.Errorf(codes.InvalidArgument, ErrBadRequest.Error())
+	}
```
kushsharma, line 103: Why do we have two name fields?

internal/store/postgres/metaschemas/defaultGroupSchema.json
```diff
@@ -0,0 +1,18 @@
+{
```
kushsharma, line 1: we can simplify the file names to just `group` or `organization`.

internal/store/postgres/metaschemas/defaultGroupSchema.json
```diff
@@ -0,0 +1,18 @@
...
+    "title":"group metadata",
+    "description":"JSON-schema for validating group metadata",
+    "type":"object",
+    "properties":{
+       "label":{
+          "title":"Label",
+          "description":"Additional context about the group",
+          "type":"string"
```
kushsharma, line 9: labels should be a kv map(object).

internal/store/postgres/migrations/20230428170234_add_metaschema_table.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE IF NOT EXISTS metaschema (
```
kushsharma, line 1: Lets also add ID, created_at, updated_at columns as well

internal/store/postgres/organization_repository.go
```diff
@@ -189,6 +190,10 @@ func (r OrganizationRepository) Create(ctx context.Context, org organization.Org
 		return organization.Organization{}, fmt.Errorf("%w: %s", parseErr, err)
 	}
 
+	if err = validateMetadataSchema(marshaledMetadata, orgMetaSchemaName); err != nil {
```
kushsharma, line 193: Validation should happen in grpc handlers, that is `v1beta1` I think. By the time the request has reached postgres, it might have already gone through lot of services. We need to fail the call early if metadata is bad.

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,303 @@
...
+
+func validateMetadataSchema(marshaledMetadata []byte, schemaName string) error {
+	if metaSchemaCache[schemaName] == "" {
+		return fmt.Errorf("%s metaschema doesn't exists", schemaName)
+	}
+	metadataSchema := gojsonschema.NewStringLoader(metaSchemaCache[schemaName])
+	providedSchema := gojsonschema.NewBytesLoader(marshaledMetadata)
+	results, err := gojsonschema.Validate(metadataSchema, providedSchema)
+	if err != nil {
+		return errors.New("failed to validate metadata")
```
kushsharma, line 296: error should also contain why it failed?

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	if !results.Valid() {
+		return errors.New("metadata doesn't match the json-schema")
```
kushsharma, line 300: Does results has reason associated with it?

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,303 @@
...
+
+	metaSchemaCache[mschema.Name] = mschema.Schema
+	return schemaName, nil
+}
+
+func (m MetaSchemaRepository) InitMetaSchemas(ctx context.Context) error {
+	// load schemas in memory from the db
+	schemas, err := m.List(ctx)
+	if err != nil {
+		return errors.New("error in initialising metadata json-schemas")
```
kushsharma, line 249: append err to this new error as well

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	for _, s := range schemas {
+		metaSchemaCache[s.Name] = s.Schema
+	}
+
+	// if default metaschemas doesnt exists add to the db
+	if metaSchemaCache[userMetaSchemaName] == "" {
+		m.Create(ctx, metaschema.MetaSchema{
```
kushsharma, line 258: should we check if the call failed with err?

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,303 @@
...
+	if err != nil {
+		return errors.New("error in initialising metadata json-schemas")
+	}
+
+	for _, s := range schemas {
+		metaSchemaCache[s.Name] = s.Schema
+	}
+
+	// if default metaschemas doesnt exists add to the db
+	if metaSchemaCache[userMetaSchemaName] == "" {
```
kushsharma, line 257: instead of writing this 4 times, what if we create map with schema name and schema value and iterate over it?

cmd/serve.go
```diff
@@ -108,24 +109,14 @@ func StartServer(logger *log.Zap, cfg *config.Shield) error {
...
-	}
-
 	deps, err := buildAPIDependencies(logger, cfg, resourceBlobRepository, dbClient, spiceDBClient)
 	if err != nil {
 		return err
 	}
+	// load metadata schema in memory from db
+	if err := deps.MetaSchemaService.InitMetaSchemas(context.Background()); err != nil {
+		logger.Warn("metaschemas initialization failed", "err", err)
+	}
```
kushsharma, line 119: I think we should move this to `shield server migrate` command. We won't be needing this once done.

kushsharma (changes requested):

internal/api/v1beta1/metaschema.go
Chief-Rishab: instead of get by id, I was doing a get by name. So name here is the path variable from our get api

internal/store/postgres/migrations/20230428170234_add_metaschema_table.up.sql
Chief-Rishab: Earlier thought of adding it, but didn't find a need for the same. How will we use created_at and updated_at field. I thought if we are already updating the in-memory map while creating/ updating the schemas, these fields wouldn't be required.

cmd/serve.go
Chief-Rishab: Have split this function into two. InitMetaSchemas will be required whenever the server starts. This function populates the in-memory cache with the existing metaschemas in the database.

MigrateDefault function being used to populate the default schemas will run only during migration.

internal/store/postgres/metaschemas/defaultGroupSchema.json
Chief-Rishab: Can you give details of what nested properties should be? Or should I allow any key:value pair in labels field?

core/metaschema/metaschema.go
```diff
@@ -0,0 +1,25 @@
...
+	"time"
+)
+
+type Repository interface {
+	Create(ctx context.Context, metaschema MetaSchema) (MetaSchema, error)
+	Get(ctx context.Context, id string) (MetaSchema, error)
+	Update(ctx context.Context, id string, metaschema MetaSchema) (MetaSchema, error)
+	List(ctx context.Context) ([]MetaSchema, error)
+	Delete(ctx context.Context, id string) error
+	CreateDefaultInDB(ctx context.Context) error
```
kushsharma, line 14: We can rename it to `MigrateDefaults`

core/metaschema/metaschema.go
```diff
@@ -0,0 +1,25 @@
...
+	InitMetaSchemas(ctx context.Context) (map[string]string, error)
```
kushsharma, line 15: Why not use `list` method for this?

internal/api/v1beta1/metaschema.go
```diff
@@ -0,0 +1,194 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+var (
+	userMetaSchema  = "user"
+	groupMetaSchema = "group"
+	orgMetaSchema   = "organization"
+	roleMetaSchema  = "role"
+
+	MetaSchemaCache           = make(map[string]string)
```
kushsharma, line 25: This cache should be in `metaschema` service(core/metaschema/service.go).

internal/api/v1beta1/metaschema.go
```diff
@@ -0,0 +1,194 @@
...
+	providedSchema := gojsonschema.NewGoLoader(mdata)
+	results, err := gojsonschema.Validate(metadataSchema, providedSchema)
+	if err != nil {
+		return errors.Wrap(err, "failed to validate metadata")
+	}
+
+	if !results.Valid() {
+		return errors.New("metadata doesn't match the json-schema")
+	}
+	return nil
```
kushsharma, line 193: We can move validate code to metaschema service code as well, expose a Validate method in it.

kushsharma (changes requested):

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,269 @@
...
+
+// add default schemas to db once during database migration
+func (m MetaSchemaRepository) MigrateDefaults(ctx context.Context) error {
+	for name, schema := range defaultMetaSchemas {
+		if _, err := m.Create(ctx, metaschema.MetaSchema{
+			Name:   name,
+			Schema: schema,
+		}); err != nil {
+			err = checkPostgresError(err)
+			if errors.Is(metaschema.ErrConflict, err) || errors.Is(err, sql.ErrNoRows) {
```
kushsharma, line 262: Should we log that there was a conflict and no changes are made in schema?

internal/store/postgres/metaschema_repository.go
```diff
@@ -0,0 +1,269 @@
...
+		default:
+			return metaschema.MetaSchema{}, fmt.Errorf("%w: %s", dbErr, err)
+		}
+	}
+
+	return schemaModel.tranformtoMetadataSchema(), nil
+}
+
+// add default schemas to db once during database migration
+func (m MetaSchemaRepository) MigrateDefaults(ctx context.Context) error {
```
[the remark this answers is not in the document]
Chief-Rishab: sounds good to me but this case we've already discussed, what if user doesn't want to use the Metaschema validation at all? In that case having a default schema will be a problem.

Currently, a user can do both, either use the custom schema by using the APIs or if the schema is deleted/updated to an empty json the metaschema validation check will altogether be avoided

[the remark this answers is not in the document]
kushsharma: Our validation will still happen with `schema`, the `default_schema` will only be used in UI to revert back the schema to its original state.

kushsharma (approved):

PR #237, opened 2023-05-04
feat: support for cors handler in admin APIs

pyadav:
Added two flags to enable cors along with admin ui change due to new endpoint
```
app:
  port: 7400
  host: 127.0.0.1
  // CorsOrigin is origin value from where we want to allow cors
  CorsOrigin string `yaml:"cors_origin" mapstructure:"cors_origin"`
```


internal/server/interceptors/cors.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+	"strings"
+)
+
+func AllowCORS(h http.Handler, origin string) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", origin)
+		w.Header().Set("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE")
+		w.Header().Set("Access-Control-Allow-Headers", "Content-Type")
+		w.Header().Set("Access-Control-Allow-Credentials", "true")
```
kushsharma, line 13: We should only send these if the request method is `OPTIONS` right?

internal/server/interceptors/cors.go
```diff
@@ -0,0 +1,36 @@
+package interceptors
+
+import (
+	"net/http"
+	"strings"
+)
+
+func AllowCORS(h http.Handler, origin string) http.Handler {
```
kushsharma, line 8: Maybe we can use mux middleware for it? https://github.com/gorilla/handlers/blob/546854cf1d61a016260b224e2526a29ff5b5c5ae/cors.go

kushsharma (changes requested):

kushsharma:
We should add the new config fields to the sample `config.yaml` available in `config` directory as well.

internal/server/interceptors/cors.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"net/http"
+
+	"github.com/gorilla/handlers"
+)
+
+func WithCors(h http.Handler, origin string) http.Handler {
+	return handlers.CORS(
+		handlers.AllowedOrigins([]string{origin}),
+		handlers.AllowedMethods([]string{"GET", "HEAD", "POST"}),
```
kushsharma, line 12: Shouldn't we allow `delete` as well? Admin will use `delete`?

kushsharma (approved):

PR #252, opened 2023-06-03
fix: refactor server package out of internal directory

kushsharma:


ravisuhag:
@kushsharma any reason to move server to pkg directory? The idea so far has been to keep pkg as something which are generic packages and which at some point can even be moved out of repo.

kushsharma:
Yeah, I wanted to allow importing a few consts and methods from outside the shield as well. If it's put in an internal package, nobody can import them outside shield.

PR #260, opened 2023-06-13
feat: add list user invitations api and add orgID filter in list user groups api

Chief-Rishab:
- Require https://github.com/raystack/proton/pull/281


kushsharma:
Can we split these docs changes in different PR? What exactly is this?

Chief-Rishab:
@kushsharma comes from the recent name change from odpf to raystack in proton files

internal/store/postgres/invitation.go
```diff
@@ -26,7 +26,7 @@ func (from Invitation) transformToInvitation() (invitation.Invitation, error) {
 		}
 	}
 	var groupIDs []string
-	if val, ok := unmarshalledMetadata["group_ids"]; ok {
+	if val, ok := unmarshalledMetadata["group_ids"]; ok && (val != nil) {
```
kushsharma, line 29: Do you think this check was unnecessary? I don't remember the case, but I saw `val` being null in sometimes.

internal/store/postgres/invitation_repository.go
```diff
@@ -115,11 +115,43 @@ func (s *InvitationRepository) List(ctx context.Context, flt invitation.Filter)
 
 	query, params, err := stmt.ToSQL()
 	if err != nil {
-		return nil, fmt.Errorf("%w: %s", queryErr, err)
+		return []invitation.Invitation{}, fmt.Errorf("%w: %s", queryErr, err)
```
kushsharma, line 118: Why not `nil`?

internal/store/postgres/invitation.go
Chief-Rishab: I created a user invitatioon just for an org, (without any group_ids) and listed the pending invitations. There was an error for trying to unmarshall the null value in the server logs

Chief-Rishab: Exact error `interface conversion: interface {} is nil, not []interface {}`

internal/store/postgres/invitation_repository.go
Chief-Rishab: yes, can revert it to nil

internal/store/postgres/invitation.go
kushsharma: Cool, make sense

kushsharma (approved):
LGTM

PR #276, opened 2023-07-09
feat: support for audit logs

kushsharma:
- Shield can store general-purpose service audit logs via `/v1beta1/organization/{orgid}/auditlogs`.
- Shield will also emit audit logs produced inside system
- Audits can be enabled or disabled via config
```yaml
log:
    audit_events: db # none(default), stdout, db
```
- Audit logs can be queried by `org id` and `start/end` time.
```protobuf
message AuditLog {
  // A unique identifier of the audit log if not supplied will be autogenerated
  string id = 1;

  // Source service generating the event
  string source = 2 [
    (validate.rules).string = {
      min_len: 2
    },
    (google.api.field_behavior) = REQUIRED,
    (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
      description: "The source service generating the event.",
      example: "\"auth\""
    }
  ];

  // Action performed, e.g. project.create, user.update, serviceuser.delete
  string action = 3 [
    (validate.rules).string = {
      min_len: 2
    },
    (google.api.field_behavior) = REQUIRED
  ];

  // Actor performing the action
  AuditLogActor actor = 4;

  // Target of the action
  AuditLogTarget target = 5;

  // Context contains additional information about the event
  map<string, string> context = 6;

  google.protobuf.Timestamp created_at = 20 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
    description: "The time the log was created.",
    example: "\"2023-06-07T05:39:56.961Z\""
  }];
}
```
- https://github.com/raystack/proton/pull/289


internal/store/postgres/audit.go
```diff
@@ -0,0 +1,52 @@
...
+
+type Audit struct {
+	ID     string `db:"id"`
+	OrgID  string `db:"org_id"`
+	Source string `db:"source"`
+	Action string `db:"action"`
+
+	Actor    types.NullJSONText `db:"actor"`
+	Target   types.NullJSONText `db:"target"`
+	Metadata types.NullJSONText `db:"metadata"`
```
Chief-Rishab, line 18: Can we add a sequence column for each org as well? Suppose if someone wants to filter logs for a specific org, using this field they can get the logs in order

Chief-Rishab: Alternatively, we can filter with time (ascending/descending) in `auditlog_repository`, in current list logs function it will be returned in random order in my opinion

kushsharma: They can use `start_time` and `end_time` for the range. But yeah let me add sorting based on `created_at`.

Chief-Rishab:
LGTM

PR #289, opened 2023-08-02
feat: add org domain whitelisting

Chief-Rishab:
- Requires  https://github.com/raystack/proton/pull/295


core/domain/domain.go
```diff
@@ -0,0 +1,24 @@
...
+	List(ctx context.Context, flt Filter) ([]Domain, error)
+	Delete(ctx context.Context, id string) error
+}
+
+type Domain struct {
+	ID         string
+	Name       string
+	OrgID      string
+	Token      string
+	Verified   bool
```
kushsharma, line 21: Let's use a enum as `State` instead with possible values as `Pending` and `Verified`

core/domain/domain.go
```diff
@@ -0,0 +1,24 @@
...
+	VerifiedAt time.Time
```
kushsharma, line 22: Simple `UpdatedAt` should also suffice instead of verified at?

internal/store/postgres/migrations/20230731211912_create_domain_table.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS domains (
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name text NOT NULL,
+    org_id uuid NOT NULL REFERENCES organizations(id) ON DELETE CASCADE,
```
kushsharma, line 4: We can create a unique constraint on org_id and name. One index on `org_id`.

pkg/server/interceptors/authorization.go
```diff
@@ -245,6 +245,26 @@ var authorizationValidationMap = map[string]func(ctx context.Context, handler *v
...
+		return handler.IsAuthorized(ctx, schema.OrganizationNamespace, pbreq.GetOrgId(), schema.UpdatePermission)
+	},
+	"/raystack.frontier.v1beta1.FrontierService/ListOrganizationDomains": func(ctx context.Context, handler *v1beta1.Handler, req any) error {
+		pbreq := req.(*frontierv1beta1.ListOrganizationDomainsRequest)
+		return handler.IsAuthorized(ctx, schema.OrganizationNamespace, pbreq.GetOrgId(), schema.GetPermission)
+	},
+	"/raystack.frontier.v1beta1.FrontierService/GetOrganizationDomain": func(ctx context.Context, handler *v1beta1.Handler, req any) error {
+		pbreq := req.(*frontierv1beta1.GetOrganizationDomainRequest)
+		return handler.IsAuthorized(ctx, schema.OrganizationNamespace, pbreq.GetId(), schema.GetPermission)
+	},
```
kushsharma, line 263: I am wondering about the authz model, if we expect user already part of the org to find possible orgs to join, how would he join it?

internal/store/postgres/domain_repository.go
```diff
@@ -0,0 +1,206 @@
...
+func (s *DomainRepository) Update(ctx context.Context, toUpdate domain.Domain) (domain.Domain, error) {
+	if strings.TrimSpace(toUpdate.ID) == "" {
+		return domain.Domain{}, domain.ErrInvalidId
+	}
+
+	query, params, err := dialect.Update(TABLE_DOMAINS).Set(
+		goqu.Record{
+			"token":      toUpdate.Token,
+			"verified":   toUpdate.Verified,
+			"updated_at": goqu.L("now()"),
```
kushsharma, line 183: I don't see any updated at column?

core/domain/domain.go
```diff
@@ -0,0 +1,24 @@
...
+	CreatedAt  time.Time
```
kushsharma, line 23: How about having an expires at column as well?

core/domain/service.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type Service struct {
+	repository  Repository
+	userService UserService
+	orgService  OrgService
+}
+
+const (
+	DNSChallenge = "_frontier-challenge.%s"
```
kushsharma, line 33: What if we set it as `_frontier-domain-challenge=%s`

core/domain/service.go
```diff
@@ -0,0 +1,163 @@
...
+	txtLength    = 16
```
kushsharma, line 34: length should be at least 32 chars?

core/domain/service.go
Chief-Rishab: Have read about this format being used else where as well 
- https://letsencrypt.org/docs/challenge-types/#dns-01-challenge
- https://www.digitalocean.com/community/tutorials/how-to-acquire-a-let-s-encrypt-certificate-using-dns-validation-with-acme-dns-certbot-on-ubuntu-18-04#step-3-setting-up-acme-dns-certbot

pkg/server/interceptors/authorization.go
Chief-Rishab: there's another API for that (get orgs by domain)

core/domain/domain.go
Chief-Rishab: was thinking of keeping it abstract since all the verified domains will not require that field, the cron job can be just used to remove the pending state domains with created-at time being more than 7 days

core/domain/service.go
kushsharma: We are using TXT records as key value pairs, although I know letsencrypt use it as `_frontier-challenge.%s` but I think clear distinction between key and value should be better approach using `=` as a separator in `_frontier-domain-verification=%s`. Also we are making it specifically for domain verification by adding `domain` in the name.

internal/api/v1beta1/domain.go
```diff
@@ -0,0 +1,224 @@
...
+type DomainService interface {
+	Get(ctx context.Context, id string) (domain.Domain, error)
+	List(ctx context.Context, flt domain.Filter) ([]domain.Domain, error)
+	ListOrgByDomain(ctx context.Context, domain string) ([]string, error)
+	Delete(ctx context.Context, id string) error
+	Create(ctx context.Context, toCreate domain.Domain) (domain.Domain, error)
+	VerifyDomain(ctx context.Context, id string) (domain.Domain, error)
+	Join(ctx context.Context, orgID string, userID string) error
+	InitDomainVerification(ctx context.Context) error
+	Close()
```
kushsharma, line 32: Are we calling these two here?

internal/api/v1beta1/domain.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (h Handler) AddOrganizationDomain(ctx context.Context, request *frontierv1beta1.AddOrganizationDomainRequest) (*frontierv1beta1.AddOrganizationDomainResponse, error) {
+	logger := grpczap.Extract(ctx)
+
+	if request.GetOrgId() == "" || request.GetDomain() == "" {
+		return nil, grpcBadBodyError
+	}
```
kushsharma, line 40: Do we have these checks in proto as well?

internal/store/postgres/domain.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/raystack/frontier/core/domain"
+)
+
+type Domain struct {
+	ID        string       `db:"id"`
+	OrgID     string       `db:"org_id"`
+	Name      string       `db:"name"`
+	Token     string       `db:"token"`
+	State     string       `db:"state"`
+	UpdatedAt sql.NullTime `db:"updated_at"`
```
kushsharma, line 16: created and updated will be same when created for the first time so it should not be null right?

internal/store/postgres/domain_repository.go
```diff
@@ -0,0 +1,199 @@
...
+	} else if flt.Name != "" && flt.State != "" {
+		stmt = stmt.Where(goqu.Ex{
+			"name":  flt.Name,
+			"state": flt.State,
+		})
+	} else if flt.OrgID != "" {
+		stmt = stmt.Where(goqu.Ex{
+			"org_id": flt.OrgID,
+		})
+	}
```
kushsharma, line 76: What about org and name

internal/api/v1beta1/domain.go
Chief-Rishab: yes

cmd/serve.go
```diff
@@ -141,6 +142,13 @@ func StartServer(logger *log.Zap, cfg *config.Frontier) error {
 		deps.SessionService.Close()
 	}()
 
+	if err := deps.DomainService.InitDomainVerification(context.Background()); err != nil {
+		logger.Warn("domains database cleanup failed", "err", err)
+	}
+	defer func() {
+		deps.DomainService.Close()
```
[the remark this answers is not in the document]
Chief-Rishab: have tested the flow, the cron works as expected and this cron closes only when the server is shut

internal/api/v1beta1/domain.go
Chief-Rishab: yes, in the server file to initialise and close the cron to remove the pending records for domain verification

kushsharma: In server file we work with actual struct instead of interface, can you confirm? I think we can remove `Close`.

kushsharma (approved):
LGTM

PR #307, opened 2023-08-17
chore: remove user orgs from joinable orgs response

Chief-Rishab:


core/domain/service.go
```diff
@@ -182,9 +182,30 @@ func (s Service) ListOrgByDomain(ctx context.Context, email string) ([]string, e
 		return nil, err
 	}
 
+	// check if user is already a member of the organization. if yes, do not include the org in the response
```
kushsharma, line 185: We should rename the method name or accept a filter argument to only list orgs the user is not part of. Current implementation `ListOrgByDomain` doesn't really list all orgs by domain.

kushsharma:
@Chief-Rishab can you rebase and resolve the conflict, the PR looks good. I'll merge.

kushsharma (approved):

PR #313, opened 2023-08-23
feat: allow custom email template to invite user to join an org

Chief-Rishab:


config/sample.config.yaml
```diff
@@ -33,17 +33,24 @@ app:
...
-  invitation_with_roles: false
+  
+  # configs for user invitation to join an organization
+  invite:
+    # invitation_with_roles if set to true will allow people in org with the permission to send invitation to users
+    # with set of role ids. When the invitation is accepted, the user will be added to the org with the roles specified
+    # This can be a security risk if the user who is inviting is not careful about the roles he is adding
+    # and cause permission escalation
+    # Note: this is dangerous and should be used with caution
+    invitation_with_roles: false
```
kushsharma, line 44: we can just call it `with_roles` if we are putting it here? Also to make it backward compatible, lets add a small hook in `config.go` that checks if the parent `invitation_with_roles` is also set.

kushsharma (changes requested):

config/sample.config.yaml
```diff
@@ -33,17 +33,24 @@ app:
...
+  # configs for user invitation to join an organization
+  invite:
+    # with_roles if set to true will allow people in org with the permission to send invitation to users
+    # with set of role ids. When the invitation is accepted, the user will be added to the org with the roles specified
+    # This can be a security risk if the user who is inviting is not careful about the roles he is adding
+    # and cause permission escalation
+    # Note: this is dangerous and should be used with caution
+    with_roles: false
+    # invite email template (if not specified, default template will be used)
+    mail_invite:
```
kushsharma, line 46: we can rename this to `mail_template`, having `invite` in name is unnecessary as we are already in invite config

config/sample.config.yaml
```diff
@@ -33,17 +33,24 @@ app:
...
+      subject: "You have been invited to join an organization"
+      body: "<div>Hi Hi {{.UserID}},</div><br><p>You have been invited to join an organization: {{.Organization}}. Login to your account to accept the invitation.</p><br><div>Thanks,<br>Team Frontier</div>"
```
kushsharma, line 48: `Hi` twice?

core/invitation/config.go
```diff
@@ -0,0 +1,14 @@
...
+
+type Config struct {
+	// WithRoles if set to true will allow roles to be passed in invitation, when the user accepts the
+	// invite, the role will be assigned to the user
+	WithRoles bool `yaml:"with_roles" mapstructure:"with_roles" default:"false"`
+
+	MailInvite MailInviteConfig `yaml:"mail_invite" mapstructure:"mail_invite"`
+}
+
+type MailInviteConfig struct {
```
kushsharma, line 11: `MailTemplate`?

core/invitation/config.go
```diff
@@ -0,0 +1,14 @@
...
+	Subject string `yaml:"subject" mapstructure:"subject" default:"You have been invited to join an organization"`
+	Body    string `yaml:"body" mapstructure:"body" default:" <div>Hi Hi {{.UserID}},</div><br><p>You have been invited to join an organization: {{.Organization}}. Login to your account to accept the invitation.</p><br><div>Thanks,<br>Team Frontier</div>"`
```
kushsharma, line 13: Here as well, `Hi` is twice.

PR #317, opened 2023-09-01
feat: add disable org on creation flag and disallow removing of last admin from the org/group

Chief-Rishab:
-  Disable org on creation flag added. 
- Org APIs now accepts both org name and uuid. Have added checks to prevent actions on disabled orgs
- Additionally a check to prevent the last org admin and project owner to be removed. Another user must be given these permissions before removing that user

### DIFF
--- a/config/sample.config.yaml
+++ b/config/sample.config.yaml
@@ -29,6 +29,9 @@ app:
   # secret string "val://user:password"
   # optional
   resources_config_path_secret: env://TEST_RESOURCE_CONFIG_SECRET
+  # org_disable_on_creation if set to true will set the org status to disabled on creation. This can be used to
+  # prevent users from accessing the org until they contact the admin and get it enabled. Default is false
+  org_disable_on_creation: false
   # disable_orgs_listing if set to true will disallow non-admin APIs to list all organizations
   disable_orgs_listing: false
   # disable_orgs_listing if set to true will disallow non-admin APIs to list all users
--- a/config/sample.config.yaml
+++ b/config/sample.config.yaml
@@ -29,9 +29,9 @@ app:
   # secret string "val://user:password"
   # optional
   resources_config_path_secret: env://TEST_RESOURCE_CONFIG_SECRET
-  # org_disable_on_creation if set to true will set the org status to disabled on creation. This can be used to
+  # disable_orgs_on_create if set to true will set the org status to disabled on creation. This can be used to
   # prevent users from accessing the org until they contact the admin and get it enabled. Default is false
-  org_disable_on_creation: false
+  disable_orgs_on_create: false
   # disable_orgs_listing if set to true will disallow non-admin APIs to list all organizations
   disable_orgs_listing: false
   # disable_orgs_listing if set to true will disallow non-admin APIs to list all users
--- a/core/organization/service.go
+++ b/core/organization/service.go
@@ -29,19 +29,21 @@ type AuthnService interface {
 }
 
 type Service struct {
-	repository      Repository
-	relationService RelationService
-	userService     UserService
-	authnService    AuthnService
+	repository           Repository
+	relationService      RelationService
+	userService          UserService
+	authnService         AuthnService
+	orgDisableOnCreation bool
 }
 
 func NewService(repository Repository, relationService RelationService,
-	userService UserService, authnService AuthnService) *Service {
+	userService UserService, authnService AuthnService, orgDisableOnCreation bool) *Service {
 	return &Service{
-		repository:      repository,
-		relationService: relationService,
-		userService:     userService,
-		authnService:    authnService,
+		repository:           repository,
+		relationService:      relationService,
+		userService:          userService,
+		authnService:         authnService,
+		orgDisableOnCreation: orgDisableOnCreation,
 	}
 }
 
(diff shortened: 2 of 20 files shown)


config/sample.config.yaml
```diff
@@ -29,6 +29,9 @@ app:
   # secret string "val://user:password"
   # optional
   resources_config_path_secret: env://TEST_RESOURCE_CONFIG_SECRET
+  # org_disable_on_creation if set to true will set the org status to disabled on creation. This can be used to
+  # prevent users from accessing the org until they contact the admin and get it enabled. Default is false
+  org_disable_on_creation: false
```
kushsharma, line 34: let's rephrase this as `disable_orgs_on_create` to keep it consistent with the existing config?

core/organization/service.go
```diff
@@ -29,19 +29,21 @@ type AuthnService interface {
...
 type Service struct {
-	repository      Repository
-	relationService RelationService
-	userService     UserService
-	authnService    AuthnService
+	repository           Repository
+	relationService      RelationService
+	userService          UserService
+	authnService         AuthnService
+	orgDisableOnCreation bool
```
kushsharma, line 36: we can create a variable to store `defaultState` as enabled/disabled instead and keep the if statement in constructor.

internal/api/v1beta1/domain.go
```diff
@@ -32,20 +33,27 @@ type DomainService interface {
...
+		switch {
+		case errors.Is(err, organization.ErrNotExist):
+			return nil, grpcOrgNotFoundErr
+		default:
+			return nil, grpcInternalServerError
+		}
+	}
+	if orgResp.State == organization.Disabled {
+		return nil, grpcOrgDisabledErr
 	}
```
kushsharma, line 48: What if we move this check to org service only and create another method `GetRaw` in org service. Org service will have two methods Get() and GetRaw().
1. Get() method will return the org only if its in enabled state, this is what we should use across all frontier
2. GetRaw will return the org no matter what the status is, this should only be used when we are explicitly hitting org API to get the org resource.

The reason I am proposing this change is avoid making mistake in future code changes where we forget to check if org is enabled or not and can lead to a security bypass. What do you think?

kushsharma (changes requested):

internal/api/v1beta1/group.go
```diff
@@ -38,6 +39,7 @@ type GroupService interface {
 
 var (
 	grpcGroupNotFoundErr = status.Errorf(codes.NotFound, "group doesn't exist")
+	grpcMinOwnerCounrErr = status.Errorf(codes.InvalidArgument, "group must have at least one owner. Please add another owner before removing this one")
```
kushsharma, line 42: we can rephrase this as `group must have at least one owner, consider adding another before removing`

pkg/server/config.go
```diff
@@ -57,7 +57,8 @@ type Config struct {
 	TelemetryConfig telemetry.Config `yaml:"telemetry_config" mapstructure:"telemetry_config"`
 
 	Authentication authenticate.Config `yaml:"authentication" mapstructure:"authentication"`
-
+	// DisableOrgsOnCreation if set to true will turn the default state of new orgs as disabled. Default is false
+	DisableOrgsOnCreation bool `yaml:"disable_orgs_on_create" mapstructure:"disable_orgs_on_create" default:"false"`
```
kushsharma, line 61: Let's rename this variable to match the config? `DisableOrgsOnCreate`

internal/api/v1beta1/org.go
```diff
@@ -25,10 +25,15 @@ import (
 	frontierv1beta1 "github.com/raystack/frontier/proto/v1beta1"
 )
 
-var grpcOrgNotFoundErr = status.Errorf(codes.NotFound, "org doesn't exist")
+var (
+	grpcOrgNotFoundErr   = status.Errorf(codes.NotFound, "org doesn't exist")
+	grpcOrgDisabledErr   = status.Errorf(codes.NotFound, "org is disabled. Please contact your administrator to enable it")
+	grpcMinAdminCountErr = status.Errorf(codes.PermissionDenied, "org must have at least one admin. Please add another admin before removing this user from the org")
```
kushsharma, line 31: Same rephrasing as group

kushsharma (commented):
Left some minor comments rest LGTM

internal/api/v1beta1/group.go
```diff
@@ -252,6 +327,28 @@ func (h Handler) AddGroupUsers(ctx context.Context, request *frontierv1beta1.Add
...
+			return nil, grpcInternalServerError
+		}
+	}
+
+	owners, err := h.userService.ListByGroup(ctx, request.GetId(), schema.OwnerRelationName)
+	if err != nil {
+		logger.Error(err.Error())
+		return nil, grpcInternalServerError
+	}
+	if len(owners) == 1 {
```
kushsharma, line 348: Hey where do we check the user we are trying to remove is part of owners list? This would fail the call if I am just removing a random user as well right if we have 1 owner?

internal/api/v1beta1/group.go
```diff
@@ -252,6 +327,28 @@ func (h Handler) AddGroupUsers(ctx context.Context, request *frontierv1beta1.Add
...
+		case errors.Is(err, organization.ErrDisabled):
+			return nil, grpcOrgDisabledErr
+		case errors.Is(err, organization.ErrNotExist):
+			return nil, grpcOrgNotFoundErr
+		default:
+			return nil, grpcInternalServerError
+		}
+	}
+
+	owners, err := h.userService.ListByGroup(ctx, request.GetId(), schema.OwnerRelationName)
```
kushsharma, line 343: Have you tested this with `schema.OwnerRelationName`. I think we should check `schema.DeletePermission` instead to ensure he is the owner.

internal/api/v1beta1/org.go
```diff
@@ -343,11 +390,33 @@ func (h Handler) AddOrganizationUsers(ctx context.Context, request *frontierv1be
...
+			return nil, grpcInternalServerError
+		}
+	}
+
+	admins, err := h.userService.ListByOrg(ctx, orgResp.ID, organization.AdminPermission)
+	if err != nil {
+		logger.Error(err.Error())
+		return nil, grpcInternalServerError
+	}
+	if len(admins) == 1 {
```
kushsharma, line 411: Same as groups, where do we check if the user being removed is actually admin?

kushsharma (approved):

PR #326, opened 2023-09-06
feat: add user and orgs avatar and return org state in response body

Chief-Rishab:
- Requires https://github.com/raystack/proton/pull/304


kushsharma (approved):
LGTM

PR #386, opened 2023-10-13
feat: support passkey as authentication method

VimVek:
Passkey is a standard that enables passwordless authentication (aka login) in the browser. It essentially defines a common API for web applications to handle passwordless technologies via FIDO2 which includes things like public key cryptography, security keys (i.e. Yubikey), and biometrics (i.e. Touch ID/Face ID, Windows Hello).

In frontier config, under authentication we can add out Relaying Party details to use passkey.It takes 3 parameters, rpdisplayname, rpid, rporigins.

rpdisplayname: A string representing the name or display name of the Relying Party.
rpid: A string representing the ID of the Relying Party in the form of a web domain.
rporigins: An array of strings specifying the allowed origins (URLs) from which authentication requests are accepted by the Relying Party. This serves as a security measure to ensure that authentication requests originate only from trusted sources.

A concise explanation of the terms used:
Relying Party (RP): The RP is the web application or service that handles passkey issuance and authentication. It operates both a client-side component (website or app) for creating and authenticating passkeys and a server-side component for registering, storing, and verifying passkey credentials.

RP Displayname: This is the name or display name of the relying party, which is presented during the passkey authentication process.

RP ID: The RP ID is the identifier for the relying party, typically in the form of a web domain. It helps ensure that passkey authentication is directed to the correct RP.

RP Origins: These are the allowed origins (URLs) from which authentication requests are accepted by the RP. It's a security measure to ensure that authentication requests come from trusted sources.

User Registration:
In the StartFlow method, the application initially checks whether the user requires registration or login by examining the "passkey_credentials" field in the user's metadata. If present, it indicates that the user is already registered, and a login is required. If absent, user registration is necessary. To complete the registration process, the application generates a webauth credential and stores it in the User metadata under the key "passkey_credentials."

User Login:
In the StartFlow method, if "passkey_credentials" exists in the user's metadata, it signifies that the user is registered and can proceed with the login process. The application sets passkey user credentials and initiates the Begin Login process, which provides options and challenges for the user. Additionally, the application stores "sessionInBytes" (representing the session) and "passkey_type" (indicating login in this case) in the flow metadata. To complete the login process, the application utilizes webAuth.ValidateLogin to verify the current credentials against the stored credentials.


kushsharma:
LGTM

PR #387, opened 2023-10-15
feat: billing for organizations

kushsharma:
- billing is composed of: billing customers, subscriptions, features, plans, prices
- current design only supports `stripe` as billing engine at the moment
- each org needs to connect at least one billing customer account to enable billing
- https://github.com/raystack/proton/pull/315


ravisuhag:
@surendratiwari3 Actually we are considering in this in upcoming releases. Will post an update soon. Feel free to create an issue and add details from your context.

kushsharma:
@surendratiwari3 From what I understand, you will use `Frontier` to subscribe users on a plan but the payment is taken outside `Frontier`(using a wire transform) and we would still want `Frontier `to acknowledge the payment and continue to renew subscriptions. If that is the case, we can handle this by marking an invoice paid manually. Every time a subscription is due for payment, it generates an invoice. If a payment method is configured, stripe automatically handles the payment and marks the invoice as paid but even marking the invoice paid manually will do the same thing.
We can have another API that allows marking an invoice paid from Frontier. Of course, only platform superuser can only be able to mark it paid. 
```
POST /v1beta1/organizations/{org_id}/billing/{billing_id}/invoices/{invoice_id}/paid
```

surendratiwari3:
@kushsharma let me reiterate on this: while creating a billing account if stripe is not configured its throwing an error. usually if stripe is not configured it should not worry about this, it will be great if we devide this into 2 different concept if i want to configure stripe i can and if not still other part should be working as it is.

PR #411, opened 2023-11-07
feat: replace an existing policy

kushsharma:
- ReplacePolicy will take an id of the existing policy and replace it with the provided policy body.
- https://github.com/raystack/proton/pull/318


ravisuhag:
@kushsharma @anujk14 is this still valid?

kushsharma:
Not needed for now.

PR #460, opened 2024-01-15
Update admin ui with ui changes along with frontier sdk.

pyadav:


ui/vite.config.ts
```diff
@@ -5,18 +5,28 @@ import tsconfigPaths from "vite-tsconfig-paths";
...
-    outDir: "dist/ui",
-  },
-  server: {
-    proxy: {
-      "/v1beta1": {
-        target: process.env.SHILD_API_URL,
-        changeOrigin: true,
+export default defineConfig(({ command, mode }) => {
+  return {
+    base: "/console",
```
rsbh, line 10: why do we need `/console` base path if we are running ui on a separate port

ui/src/containers/groups.create/index.tsx
```diff
@@ -57,9 +61,17 @@ export default function NewGroup() {
...
-    await trigger(data);
-    navigate("/console/groups");
-    navigate(0);
+    if (!organisation) return;
+    try {
+      await client?.frontierServiceCreateGroup(organisation, data);
+      toast.success("members added");
+      navigate("/console/groups");
+      navigate(0);
+    } catch ({ error }: any) {
```
rsbh, line 70: it should be `catch ( error : any) ` here

ui/src/containers/users.create/index.tsx
```diff
@@ -35,9 +35,16 @@ export default function NewUser() {
...
   const onSubmit = async (data: any) => {
-    await trigger(data);
-    navigate("/console/users");
-    navigate(0);
+    try {
+      await client?.frontierServiceCreateUser(data);
+      toast.success("user added");
+      navigate("/console/users");
+      navigate(0);
+    } catch ({ error }: any) {
```
rsbh, line 43: `catch (error : any)`

rsbh (changes requested):

pkg/server/server.go
```diff
@@ -53,6 +52,22 @@ const (
...
+		logger.Error("ui server failed: no port specified")
+		return
+	}
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "ui/dist/ui/index.html")
+	})
+
+	logger.Info("ui server starting", "http-port", cfg.Port)
+	utils.Open(fmt.Sprintf("http://localhost:%d", cfg.Port))
```
rsbh, line 67: why do we need to open UI in browser, can we have config to disable this?

pkg/server/server.go
```diff
@@ -53,6 +52,22 @@ const (
...
 
+func ServeUI(ctx context.Context, logger log.Logger, cfg UIConfig) {
+	isUIPortNotExits := cfg.Port == 0
+	if isUIPortNotExits {
+		logger.Error("ui server failed: no port specified")
+		return
+	}
+
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "ui/dist/ui/index.html")
```
rsbh, line 63: I ran the `make build` command but `ui/dist` folder is empty. getting 404 error in browser.
`make ui` also gave error as packages are not installed.
[
Screenshot 2024-01-17 at 9 19 06 PM
](url)

pkg/server/server.go
```diff
@@ -53,6 +52,22 @@ const (
 	grpcDialTimeout = 5 * time.Second
 )
 
+func ServeUI(ctx context.Context, logger log.Logger, cfg UIConfig) {
+	isUIPortNotExits := cfg.Port == 0
+	if isUIPortNotExits {
+		logger.Error("ui server failed: no port specified")
```
kushsharma, line 58: setting port should not be logged with error, could be a warning like `ui port unspecified, keeping it disabled`

pkg/server/server.go
kushsharma: Yeah we should remove this

rsbh (approved):
LGTM

PR #471, opened 2024-01-24
feat(sdk): show plan upgrade downgrade

rsbh:


anujk14 (approved):
LGTM.

@pyadav I've had a look and it looks good to me. Do have a look later whenever possible.

PR #474, opened 2024-01-25
feat: admin console list pages

pyadav:


ui/src/containers/groups.list/header.tsx
```diff
@@ -1,35 +1,37 @@
 import { PlusIcon } from "@radix-ui/react-icons";
 
-import { Button, DataTable, Flex, Text, useTable } from "@raystack/apsara";
+import { Button, DataTable, Flex, useTable } from "@raystack/apsara";
 import { useNavigate } from "react-router-dom";
+import PageHeader from "~/components/page-header";
 
+const pageHeader = {
+  title: "Projects",
```
rsbh, line 8: title should be `Teams` or `Groups`

ui/src/containers/organisations.list/settings/index.tsx
```diff
@@ -0,0 +1,164 @@
...
+        // @ts-ignore
+        map[el.name] = el;
+        return map;
+      },
+      {}
+    );
+  }, [preferences]);
+
+  useEffect(() => {
+    if (preferencesMap["social_login"])
```
rsbh, line 43: The auth strategies are global level in the backend, not org level.
this section should be on a root level setting page

ui/types/organisation.d.ts
```diff
@@ -2,6 +2,7 @@ export type Organisation = {
   id: string;
```
rsbh, line 2: why we have custom Org interface, cant we use `V1Beta1Organization` ?

ui/src/containers/organisations.list/settings/index.tsx
pyadav: we are using same for SDK as well

ui/src/containers/groups.list/header.tsx
pyadav: Fixed

rsbh (approved):

PR #475, opened 2024-01-29
feat(sdk): add helper component to show formated amount

rsbh:


sdks/js/packages/core/react/components/helpers/Amount/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface AmountProps {
+  value: number;
+  currency?: string;
+  className?: string;
+  currencyClassName?: string;
+  decimalClassName?: string;
+  valueClassName?: string;
+}
+
+const currenySymbolMap: Record<string, string> = {
```
anujk14, line 12:
```suggestion
const currencySymbolMap: Record<string, string> = {
```

sdks/js/packages/core/react/components/helpers/Amount/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  usd: '$',
+  inr: '₹'
+} as const;
+
+const currenyDecimalMap: Record<string, number> = {
```
anujk14, line 17:
```suggestion
const currencyDecimalMap: Record<string, number> = {
```

PR #488, opened 2024-02-07
fix (admin-ui): set default page size of 200 when fetching user list

anujk14:
By default, backend sends 50 users only if no pageSize is set.


rsbh (approved):
LGTM

PR #489, opened 2024-02-08
chore: billing account and subscription management should be done by users with org delete permission

anujk14:
Currently, some subscription management actions can be performed by users with delete permission on the org, whereas some others allow users with update permission to make subscription changes. 
For example, currently, users with update permission can purchase a plan, but not change it (using ChangeSubscription API) or cancel it.

This PR tries to bring consistency to billing account and subscription management by allowing only users with org delete permission to manage both. 

Current:
1. Update subscription checks for `OrganizationUpdate` permission
2. Create checkout checks for `OrganizationUpdate` permission

Post this PR:
1. Update subscription checks for `OrganizationDelete` permission
2. Create checkout checks for `OrganizationDelete` permission


kushsharma (changes requested):
Billing related changes on organisation should be done by org admin only(having delete permission). Why do you think this change is needed?

PR #495, opened 2024-02-14
feat(admin-ui): add logout

rsbh:


ui/src/routes.tsx
```diff
@@ -31,24 +31,49 @@ import RoleDetails from "./containers/roles.list/details";
...
-    } catch (error) {
-      setIsAdmin(false);
     }
-  }, []);
+  }, [client, isUserEmpty]);
+
+  const isLoading = isOrgListLoading || isUserLoading;
 
-  return user && isAdmin ? (
+  return isLoading ? (
```
anujk14, line 67: [Optional] I think we should use `if..else` statements here instead of nested ternary operators. The current implementation is a bit difficult to read.

anujk14 (approved):
LGTM

PR #497, opened 2024-02-15
feat(admin-ui): add preferences pages to get list and update preference

rsbh:


ui/src/containers/preferences.list/details.tsx
```diff
@@ -0,0 +1,206 @@
...
+      },
+      {
+        href: `/preferences/${trait?.name}`,
+        name: `${trait?.title}`,
+      },
+    ],
+  };
+
+  useEffect(() => {
+    async function getPreference(name: string) {
```
pyadav, line 71: we can use useOutletContext for passing context to detail page.
For reference you can look into admin ui repo instead of fetching all preferences each time

ui/src/containers/preferences.list/details.tsx
```diff
@@ -0,0 +1,206 @@
...
+        // @ts-ignore
+        checked={checked}
+        onCheckedChange={(v: boolean) => onChange(v.toString())}
+      />
+    );
+  } else if (R.or(R.has("text"), R.has("textarea"))(trait)) {
+    return (
+      <TextField value={value} onChange={(e) => onChange(e.target.value)} />
+    );
+  } else {
```
pyadav, line 40: Not able to edit 
- `Invite Mail Body`
- `Full name`
etc. 


Maybe they have different type. Can you look into that






<br class="Apple-interchange-newline">

rsbh: `Full name` is fixed, there is no field type for `Invite Mail Body`. we will fix that in backend.

ui/src/containers/preferences.list/details.tsx
rsbh: Created a Layout component and fetched data in that. the list and details component are children routes and read data form outlet context.

pyadav (approved):
LGTM

PR #513, opened 2024-02-28
fix(sdk): use org projects api for all projects

rsbh:


sdks/js/packages/core/react/hooks/useOrganizationProjects.ts
```diff
@@ -1,47 +1,56 @@
...
-  useEffect(() => {
-    getProjects();
-  }, [client, getProjects, organization?.id]);
+        const newProjects = resp?.data?.projects || [];
+        // @ts-ignore
+        const access_pairs = resp?.data?.access_pairs || [];
+        setProjects(newProjects);
+        setAccessPairs(access_pairs);
+      } catch (err) {
+        console.error(err);
```
anujk14, line 47: Let's remove this?

sdks/js/packages/core/react/hooks/useOrganizationProjects.ts
```diff
@@ -1,47 +1,56 @@
...
+    async (org_id: string) => {
+      try {
+        setIsProjectsLoading(true);
+        const resp = allProjects
+          ? await client?.frontierServiceListOrganizationProjects(org_id, {
+              with_member_count: withMemberCount
+            })
+          : await client?.frontierServiceListProjectsByCurrentUser({
+              org_id,
+              with_permissions: ['update', 'delete'],
```
anujk14, line 36: If I am added as a project viewer somewhere, I should be able to view those projects. In that case, I won't have update and delete permissions. 
What happens if we add `get` permission check as well, and keep `non-inherited` as true? Will that help project viewers list their non-inherited projects in this screen?

sdks/js/packages/core/react/hooks/useOrganizationProjects.ts
rsbh: this is only for error logging, we will replace this when we will have support for injecting logger to sdk

sdks/js/packages/core/react/hooks/useOrganizationProjects.ts
rsbh: the projects list same for all permission. the `with_permissions` flag will send map of user permissions per projects in list response.
we use that to show edit and delete option in projects list page.

anujk14: Oh, got it!

anujk14 (approved):
LGTM

PR #548, opened 2024-03-19
feat(sdk): show interval with plan name in plan change banner and modal

rsbh:


sdks/js/packages/core/react/components/common/upcoming-plan-change-banner/index.tsx
```diff
@@ -113,15 +114,23 @@ export function UpcomingPlanChangeBanner({
     subscription?.id
   ]);
 
+  const currentPlanInterval =
```
anujk14, line 117: Can we have a function called `func getUserFriendlyPlanName(plan, intervalMap)` or `func getPrettyPlanName(plan, intervalMap)` which does this? That will ensure that the logic for figuring out plan name is not duplicated, and the code will be easier to manage as well.


Something like:
```
func getUserFriendlyPlanName(plan, intervalMap) {
    interval = intervalMap[plan?.interval as IntervalKeys];
    return `${plan?.title} (${interval})`;
}
```

anujk14 (changes requested):

anujk14 (approved):
LGTM

PR #550, opened 2024-03-20
chore(sdk): picking the metadata value based on product name from the feature

rsbh:


sdks/js/packages/core/react/components/organization/plans/index.tsx
```diff
@@ -283,7 +286,9 @@ const PlanPricingColumn = ({
               />
             ) : value ? (
               <Text>{value}</Text>
```
anujk14, line 288: Won't this mean that we show `false` here? Since values are now `false` not `-` in the metadata?

rsbh: if we explicitly define `false` then `false` will come. for empty string or no value means key is not defined in metadata, then `-` will be shown.

anujk14 (approved):
LGTM

PR #553, opened 2024-03-22
feat(admin-ui): add support to update user role and remove user

rsbh:


ui/src/containers/organisations.list/users/columns.tsx
```diff
@@ -30,19 +50,141 @@ export const getColumns: (
...
       footer: (props) => props.column.id,
     },
+    {
+      header: "Actions",
+      cell: ({ row, getValue }) => {
+        const userRoleIds = R.pipe(
+          R.pathOr([], [row?.original?.id || "", "roles"]),
+          R.map(R.path(["id"]))
+        )(userRolesMap);
+        const excluedRoles = R.filter(
```
anujk14, line 73:
```suggestion
        const excludedRoles = R.filter(
```

ui/src/containers/organisations.list/users/index.tsx
```diff
@@ -1,18 +1,58 @@
...
+        } = await client?.frontierServiceListOrganizationRoles(ordId, {
+          scopes: ["app/organization"],
+        });
+        const {
+          // @ts-ignore
+          data: { roles },
+        } = await client?.frontierServiceListRoles({
+          scopes: ["app/organization"],
+        });
+        setRoles([...roles, ...orgRoles]);
```
anujk14, line 47: Just confirming, this current PR adds functionality just to update org level roles, right?

rsbh: yes only org level roles

anujk14 (approved):
LGTM

PR #557, opened 2024-03-27
feat(sdk): trial integration in plans screen

rsbh:


sdks/js/packages/core/react/components/organization/plans/hooks/usePlans.tsx
```diff
@@ -59,7 +68,9 @@ export const usePlans = () => {
               cancel_url: cancel_url,
               success_url: success_url,
               subscription_body: {
-                plan: planId
+                plan: planId,
+                skip_trial: !isTrial,
+                cancel_after_trial: config?.billing?.cancelAfterTrial || false
```
anujk14, line 73: Should we set the default behaviour to cancel by default instead of setting `false`?
We've mainly built and tested frontier to work with a single paid plan. The default behaviour here might introduce untested scenarios.

sdks/js/packages/core/react/contexts/FrontierContext.tsx
```diff
@@ -274,8 +279,10 @@ export const FrontierContextProvider = ({
           orgId,
           billingId
         );
-        if (resp?.data?.subscriptions?.length) {
-          const activeSub = getActiveSubscription(resp?.data?.subscriptions);
+        const subscriptionsList = resp?.data?.subscriptions || [];
+        setSubscriptions(subscriptionsList);
```
anujk14, line 283: What does `setSubscriptions` do?

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -0,0 +1,354 @@
+import { useNavigate } from '@tanstack/react-router';
```
anujk14, line 1: [Nitpick] Should we try to set the linter to automatically sort imports alphabetically, and segment them according to whether they are external imports or internal ones?

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+const FeaturesList = ({ features, plan }: FeaturesListProps) => {
+  return features.map(feature => {
+    const planFeature = _.get(plan.features, feature, {
+      metadata: {}
+    });
+    const productMetaDataFeatureValues = plan.productNames
+      .map(name => _.get(planFeature.metadata, name))
+      .filter(value => value !== undefined);
+    // piciking the first value for feature metadata, in case of muliple products in a plan, there can be multiple metadata values.
```
anujk14, line 40:
```suggestion
    // picking the first value for feature metadata, in case of multiple products in a plan, there can be multiple metadata values.
```

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -0,0 +1,354 @@
...
+    trailSubscription
+  } = usePlans();
+
+  useEffect(() => {
+    if (planHasTrial) {
+      checkAlreadyTrialed(planIds);
+    }
+  }, [checkAlreadyTrialed, planHasTrial, planIds]);
+
+  const trailEndDate = planIds.includes(trailSubscription?.plan_id || '')
```
anujk14, line 128:
```suggestion
  const trialEndDate = planIds.includes(trialSubscription?.plan_id || '')
```

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -0,0 +1,354 @@
...
+    : '';
+
+  const showButton = isUpgrade && !hasAlreadyTrailed && planHasTrial;
+  return (
+    <Flex
+      className={plansStyles.trialWrapper}
+      justify={'center'}
+      align={'center'}
+    >
+      {isTrailCheckLoading ? (
```
anujk14, line 139: Is there a more readable way we can put this? Nested ternaries can get a bit difficult to understand later on

anujk14 (changes requested):

sdks/js/packages/core/react/contexts/FrontierContext.tsx
rsbh: we use to keep only 1 active subscription in the state. after trial user can have 2 subscription running.
one is in active state and other is in trialing state.
and we already fetch all subscriptions in order to set active one.
so instead of fetching all subscriptions again for trial, i just save this subscriptions list in the state.

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
rsbh: we can sort alphabetically, but sorting external and internal are difficult as we use alias and relative path for internal imports.

sdks/js/packages/core/react/components/organization/plans/hooks/usePlans.tsx
```diff
@@ -22,17 +24,24 @@ interface verifyPlanChangeOptions {
...
     fetchActiveSubsciption
   } = useFrontier();
 
+  const trailSubscription = subscriptions.find(
+    sub => sub.state === SUBSCRIPTION_STATES.TRIALING
+  );
+
   const checkoutPlan = useCallback(
-    async ({ planId, onSuccess }: checkoutPlanOptions) => {
+    async ({ planId, onSuccess, isTrial }: checkoutPlanOptions) => {
```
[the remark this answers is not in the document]
rsbh: we want default falsy value for `isTrial`. the checkout api default behaviour is to enable trial if trail_days are set.
so the we have to explicitly pass `skip_trial` `true` for normal checkouts  and `skip_trial` `false` for trial.

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
rsbh: we can change something when we do refactoring in future

sdks/js/packages/core/react/components/organization/plans/hooks/usePlans.tsx
anujk14: Bump

anujk14 (approved):
LGTM

PR #565, opened 2024-04-02
feat: collect address on checkout if it is required for taxation, and sync it

anujk14:
Currently, if `stripe_auto_tax` is enabled, and the user does not have an address associated with the billing account, we get an error from stripe as follows:
```
{
    "code": "customer_tax_location_invalid",
    "doc_url": "https://stripe.com/docs/error-codes/customer-tax-location-invalid",
    "status": 400,
    "message": "Automatic tax calculation in Checkout requires a valid address on the Customer. Add a valid address to the Customer or set either `customer_update[address]` to 'auto' or `customer_update[shipping]` to 'auto' to save the address entered in Checkout to the Customer.",
    "request_id": "",
    "request_log_url": "",
    "type": "invalid_request_error"
}
```

In order to fix this, the `customer_update[address]` field needs to be set to `auto` in order to collect the billing address for taxation during the checkout. This address is later synced back to Frontier's database in the `syncWithProvider` method in checkouts.

Stripe checkout screen when address is not pre-added:
Screenshot 2024-04-02 at 1 54 47 PM

Stripe checkout screen when address was already added earlier to stripe account:
Screenshot 2024-04-02 at 1 54 36 PM

### DIFF
--- a/billing/checkout/service.go
+++ b/billing/checkout/service.go
@@ -485,7 +485,6 @@ func (s *Service) SyncWithProvider(ctx context.Context, customerID string) error
 				}
 			}
 		}
-
 	}
 
 	// if payment is completed, create subscription for them in system
--- a/billing/checkout/service.go
+++ b/billing/checkout/service.go
@@ -50,6 +50,7 @@ type Repository interface {
 type CustomerService interface {
 	GetByID(ctx context.Context, id string) (customer.Customer, error)
 	List(ctx context.Context, filter customer.Filter) ([]customer.Customer, error)
+	Update(ctx context.Context, customer customer.Customer) (customer.Customer, error)
 }
 
 type PlanService interface {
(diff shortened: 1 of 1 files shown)


billing/checkout/service.go
```diff
@@ -456,6 +460,31 @@ func (s *Service) SyncWithProvider(ctx context.Context, customerID string) error
 		if checks[idx], err = s.repository.UpdateByID(ctx, ch); err != nil {
 			return fmt.Errorf("failed to update checkout session: %w", err)
 		}
+
+		if checkoutSession.CustomerDetails != nil && checkoutSession.CustomerDetails.Address != nil {
```
kushsharma, line 464: Instead of syncing these details here, I think we should add a `cron` syncer just like we have for checkouts/subscriptions if we want to keep customer details in sync always. This is not done for customers at the moment.

PR #567, opened 2024-04-03
fix: disable autotax on free billing plans

kushsharma:


billing/checkout/service.go
```diff
@@ -608,7 +615,7 @@ func (s *Service) ensureSubscription(ctx context.Context, ch Checkout) (string,
 	}
 
 	// if set to cancel after trial, schedule a phase to cancel the subscription
-	if ch.CancelAfterTrial {
+	if ch.CancelAfterTrial && plan.TrialDays > 0 {
```
rsbh, line 618: i think we should check subscription state as `trialing` here

rsbh (changes requested):

billing/checkout/service.go
anujk14: +1.

@kushsharma Also, we create the subscription here using checkout details in the `ensureSubscription` method. Should we instead fetch the actual Stripe subscription and then call our subscription service with those details?

kushsharma: How did that help?

PR #576, opened 2024-04-08
feat: add support for per month price in year plan

rsbh:


sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -24,6 +24,64 @@ import Amount from '~/react/components/helpers/Amount';
...
+  const perIntervalLabel = showPerMonthPrice
+    ? 'per seat/month'
+    : `per seat/${selectedInterval}`;
+
+  const amount = showPerMonthPrice
+    ? selectedIntervalPricing.amount / 12
+    : selectedIntervalPricing.amount;
+
+  const actualPerMonthAmount = plan.intervals['month']?.amount || 0;
+  const showDiscount = showPerMonthPrice && actualPerMonthAmount > 0;
```
anujk14, line 50: What does the second part here check for `actualPerMonthAmount > 0`?

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
```diff
@@ -24,6 +24,64 @@ import Amount from '~/react/components/helpers/Amount';
...
+
+  return (
+    <Flex gap="small" direction="column">
+      <Flex align={'center'} gap={'small'}>
+        <Text size={4} className={plansStyles.planTitle}>
+          {plan.title}
+        </Text>
+        {showDiscount && discount > 0 ? (
+          <Flex className={plansStyles.discountText}>
+            <Text weight={500}>{discount * -1}%</Text>
```
anujk14, line 63: Should we do an integer division to come up with discount percentage? We don't usually have decimal discounts like 10.12% off.

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
rsbh: it is needed to calculate discount, as we need to divide the price delta with `actualPerMonthAmount`. thus checking if value is greater than 0.
i have refactored to make it more readable

sdks/js/packages/core/react/components/organization/plans/pricing-column.tsx
rsbh: removed decimal from discount percentage

anujk14 (approved):

PR #578, opened 2024-04-10
feat(sdk): add plan interval cycle toggle

rsbh:
This PR will add support for plan interval cycle toggle.
users can switch plan interval from billing screen.
Screenshot 2024-04-10 at 12 36 22 PM
Screenshot 2024-04-10 at 12 35 19 PM

### DIFF
--- a/sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
+++ b/sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
@@ -7,19 +7,47 @@ import { useCallback, useEffect, useState } from 'react';
 import { useFrontier } from '~/react/contexts/FrontierContext';
 import { V1Beta1Plan } from '~/src';
 import { toast } from 'sonner';
-import { getPlanIntervalName } from '~/react/utils';
+import { getPlanIntervalName, getPlanPrice } from '~/react/utils';
+import * as _ from 'lodash';
+import { usePlans } from '../../plans/hooks/usePlans';
+import dayjs from 'dayjs';
+import { DEFAULT_DATE_FORMAT } from '~/react/utils/constants';
 
 export function ConfirmCycleSwitch() {
-  const { activePlan, client } = useFrontier();
-
+  const { activePlan, client, paymentMethod, config } = useFrontier();
   const navigate = useNavigate({ from: '/billing/cycle-switch/$planId' });
   const { planId } = useParams({ from: '/billing/cycle-switch/$planId' });
-  const cancel = useCallback(() => navigate({ to: '/billing' }), [navigate]);
+  const dateFormat = config?.dateFormat || DEFAULT_DATE_FORMAT;
 
   const [isPlanLoading, setIsPlanLoading] = useState(false);
   const [nextPlan, setNextPlan] = useState<V1Beta1Plan>();
-
   const [isCycleSwitching, setCycleSwitching] = useState(false);
+
+  const closeModal = useCallback(
+    () => navigate({ to: '/billing' }),
+    [navigate]
+  );
+
+  const {
+    checkoutPlan,
+    isLoading: isPlanActionLoading,
+    changePlan,
+    verifyPlanChange
+  } = usePlans();
+
+  const nextPlanPrice = nextPlan ? getPlanPrice(nextPlan) : { amount: 0 };
+  const isPaymentRequired =
+    _.isEmpty(paymentMethod) && nextPlanPrice.amount > 0;
+
+  const nextPlanIntervalName = getPlanIntervalName(nextPlan);
+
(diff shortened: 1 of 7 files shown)


sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const {
+    checkoutPlan,
+    isLoading: isPlanActionLoading,
+    changePlan,
+    verifyPlanChange
+  } = usePlans();
+
+  const nextPlanPrice = nextPlan ? getPlanPrice(nextPlan) : { amount: 0 };
+  const isPaymentRequired =
```
anujk14, line 39: Can we rename this to `isPaymentMethodRequired` ?

sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    _.isEmpty(paymentMethod) && nextPlanPrice.amount > 0;
+
+  const nextPlanIntervalName = getPlanIntervalName(nextPlan);
+
+  const nextPlanMetadata = nextPlan?.metadata as Record<string, number>;
+  const activePlanMetadata = activePlan?.metadata as Record<string, number>;
+
+  const isUpgrade =
```
anujk14, line 47: TODO for later: Let's move this to a common place?

```
func isUpgrade(plan1, plan2)
Return type: boolean
```

Our logic for determining whether a plan is an upgrade should be the same everywhere. A common method would help us enforce that.
Same for `isDowngrade`, `isPaymentMethodRequired` etc.

sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+      getNextPlan(planId);
+    }
+  }, [client, planId]);
+
+  const isLoading = isPlanLoading;
+
+  async function onConfirm() {
+    setCycleSwitching(true);
+    try {
+      if (nextPlan?.id) {
```
anujk14, line 78: In what scenario will we not have `nextPlan.id`? This method is on clicking "confirm" of the dialogue box, right?

sdks/js/packages/core/react/components/organization/billing/upcoming-billing-cycle.tsx
```diff
@@ -29,6 +34,45 @@ function LabeledBillingData({
...
+        className={billingStyles.linkBtn}
+        onClick={onClick}
+      >
+        Switch to {intervalName}
+      </Button>
+    </div>
+  );
+}
+
+function getSwitchablePlan(plans: V1Beta1Plan[], currentPlan: V1Beta1Plan) {
```
anujk14, line 63: In the original requirements, are we supposed to show the "Switch to ..." button only for upgrades, or for any cycle switch?

rsbh: i think we are showing switch for all cases. (upgrade and downgrade)

sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
rsbh: The schema type generated by (swagger-typescript-api)[https://github.com/acacode/swagger-typescript-api] has all keys as optional.
I will fix that in separate PR

sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
rsbh: yes, i tried to refactor it, but current PR got big due to that i removed the refactor changes.
right now the same logic is used at 2 places. i will refactor this in future

anujk14: Sounds good

sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
anujk14: Okay

sdks/js/packages/core/react/components/organization/billing/upcoming-billing-cycle.tsx
anujk14: Can you confirm on pe-design once? From the design, I also feel we need to show it for both. But based on recent experience, I think we should confirm it once. There is scope for misunderstanding there.

anujk14 (approved):
lgtm

PR #582, opened 2024-04-15
chore: allow transaction listing with get permission

anujk14:


kushsharma (approved):
Ideally, we should have `billingview` and `billingmanage` permission that can be assigned to roles to manage access better but it's fine to do this for now.

PR #587, opened 2024-04-16
feat(sdk): add modal for plan upgrade

rsbh:


sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
```diff
@@ -160,7 +160,11 @@ export function ConfirmCycleSwitch() {
                 New cycle:
               </Text>
               <Text size={2} style={{ color: 'var(--foreground-muted)' }}>
-                {nextPlanIntervalName}(effective from the next billing cycle)
+                {nextPlanIntervalName} (
+                {isUpgrade
+                  ? 'effective immediately'
+                  : 'effective from the next billing cycle'}
+                )
```
anujk14, line 167: Can we not have the same component for cycle-switch and confirm-change, and simply pass text values from the outside to it? Both the components seem to be very similar.

rsbh: I tried to combine both. But we need to refactor the logic first.
The changes were increasing the scope of current task. we can pick changes later

anujk14 (approved):

PR #599, opened 2024-04-29
feat: add api to get customer portal url

rsbh:
This PR will add support for customer portal setup in checkout api.
the api will return stripe's customer portal url as checkout url. customer can update payment methods, tax details, address in that portal.

ref: https://docs.stripe.com/customer-management, https://docs.stripe.com/customer-management/integrate-customer-portal


kushsharma:
Instead of creating a new API which is very specific to Stripe, how about we add this as a setup method in `CheckoutSetupBody`?

billing/checkout/service.go
```diff
@@ -728,6 +728,47 @@ func (s *Service) CreateSessionForPaymentMethod(ctx context.Context, ch Checkout
...
+
+	return Checkout{
+		ID:          checkoutID,
+		ProviderID:  session.ID,
+		CustomerID:  billingCustomer.ID,
+		CancelUrl:   ch.CancelUrl,
+		SuccessUrl:  ch.SuccessUrl,
+		CheckoutUrl: session.URL,
+		State:       "",
+		ExpireAt:    utils.AsTimeFromEpoch(0),
```
kushsharma, line 765: lets omit these vars if not setting them

billing/checkout/service.go
```diff
@@ -728,6 +728,47 @@ func (s *Service) CreateSessionForPaymentMethod(ctx context.Context, ch Checkout
...
+
+	sessionParams := &stripe.BillingPortalSessionParams{
+		Params: stripe.Params{
+			Context: ctx,
+		},
+		Customer: stripe.String(billingCustomer.ProviderID),
+	}
+
+	if ch.CancelUrl != "" {
+		sessionParams.ReturnURL = stripe.String(ch.CancelUrl)
```
kushsharma, line 748: does that mean setting return url is optional? Also how about success URL?

rsbh: billing session params only need return url. which is optional.
https://github.com/stripe/stripe-go/blob/v72.122.0/billingportal_session.go#L40

kushsharma (approved):

PR #602, opened 2024-04-30
feat(admin-ui): add page to generate token for service user

rsbh:


ui/src/containers/organisations.list/serviceusers/tokens/add.tsx
```diff
@@ -0,0 +1,153 @@
...
+                </Flex>
+              ) : null}
+
+              {token?.token ? (
+                <Text
+                  size={4}
+                  weight={500}
+                  style={{ color: "var(--foreground-danger)" }}
+                >
+                  For safety reasons, we can't show it again. Please copy it
```
anujk14, line 127: What is `it` here? Can we explicitly use the word `the token` here?

ui/src/containers/organisations.list/serviceusers/details.tsx
```diff
@@ -0,0 +1,120 @@
...
+      } = await client?.frontierServiceGetOrganization(organisationId ?? "");
+      setOrganisation(organization);
+    }
+
+    async function getServiceUser(userId: string) {
+      const resp = await client?.frontierServiceGetServiceUser(userId);
+      const user = resp?.data?.serviceuser;
+      setServiceUser(user);
+    }
+    if (organisationId) {
```
anujk14, line 58: Didn't quite get why we're fetching the `organizationId` here. Is it to show the org name in breadcrumbs?

ui/src/containers/organisations.list/serviceusers/tokens/add.tsx
rsbh: not just token, user has to copy id as well.

ui/src/containers/organisations.list/serviceusers/details.tsx
rsbh: yes, but this can also check if org exist or not.

anujk14: Do we need to check that though? 
In the absence of an org, the page navigation (using org ->service users->user details would not work. 
If someone directly opens this page, won't the service user fetch anyway fail because the user would be detached from org?

ui/src/containers/organisations.list/serviceusers/tokens/add.tsx
anujk14: In that case, can we mention it as "the credentials" instead of "it"?
Case in point: I thought it meant only the secret, not Id. Someone else might also misinterpret the text.

anujk14 (approved):
LGTM

PR #615, opened 2024-05-09
fix: send default payment method flag in metadata

anujk14:
Without any default payment method:

Screenshot 2024-05-09 at 1 27 26 PM

With default payment method:
image


billing/customer/service.go
```diff
@@ -206,6 +206,9 @@ func (s *Service) ListPaymentMethods(ctx context.Context, id string) ([]PaymentM
 		ListParams: stripe.ListParams{
 			Context: ctx,
 		},
+		Expand: []*string{
```
rsbh, line 209: why are we doing this?

rsbh (approved):

kushsharma (approved):

PR #616, opened 2024-05-10
Feat sdk add trial checks

rsbh:


sdks/js/packages/core/react/components/organization/plans/hooks/usePlans.tsx
```diff
@@ -155,40 +168,58 @@ export const usePlans = () => {
...
-            )
-          );
-          const value = resps.some(resp => resp?.data?.trialed);
-          setHasAlreadyTrialed(value);
+      const subscribedPlans = getSubscribedPlans();
+      const subscribedPlansIdsSet = new Set(
+        subscribedPlans.map(plan => plan.id)
+      );
+      const maxTrialPlanWeightage = getTrialedPlanMaxWeightage(subscribedPlans);
+      const trialedPlans = allPlans.reduce((acc, plan) => {
```
anujk14, line 195: For readability, can we use map here, and pass the result to `setHasAlreadyTrialed` as follows:

```
// overall flow remains here
const trialedPlans = allPlans.map(getPlansIneligibleForTrial).filter( plan => plan)
setHasAlreadyTrialed(trialedPlans)


// details of how we are filtering plans which cannot be trialed. In future if we have any change in trial logic, only this function needs to change
function getPlansIneligibleForTrial(plan) {
        const metadata = plan?.metadata as PlanMetadata;
        const weightage = metadata?.weightage || 0;
       if (plan?.id && subscribedPlansIdsSet.has(plan?.id)) || (maxTrialPlanWeightage > 0 && weightage < maxTrialPlanWeightage) {
        return plan
    }
}
```

anujk14 (approved):

PR #629, opened 2024-05-27
fix(sdk): hide interval switch if plan is already phased. show interval in resume button if plan is similar

rsbh:


sdks/js/packages/core/react/components/organization/billing/upcoming-billing-cycle.tsx
```diff
@@ -223,6 +223,10 @@ export const UpcomingBillingCycle = ({
     navigate({ to: planInfo.action.link });
   };
 
+  const alreadyPhased = activeSubscription?.phases?.find(
+    phase => phase.plan_id === switchablePlan?.id
+  );
```
anujk14, line 228: What is the expected value for `alreadyPhased` in the following scenario:
1. I have enterprise yearly
2. I downgrade to professional yearly
3. I come to billing page.
Will the user see switch to monthly for enterprise, or will it be skipped?

rsbh: user will still see the button for `switch to monthly` for the enterprise plan.

anujk14 (approved):

PR #640, opened 2024-06-11
feat: emit audit event+webhook when user is removed from a group

anujk14:
We require a webhook which notifies downstream services about removal of user from groups. This PR introduces that into the service.

The reason for doing this in service is so that this event is triggered even if different handlers are used to perform this action.

The idea behind adding this instead of adding more and more information into policy events is to provide an easy to understand event to downstream services so that appropriate action can be taken by them. A few more events such as this will be introduced in subsequent PRs.

### DIFF
--- a/core/group/service.go
+++ b/core/group/service.go
@@ -307,6 +307,12 @@ func (s Service) AddUsers(ctx context.Context, groupID string, userIDs []string)
 // RemoveUsers removes users from a group as members
 func (s Service) RemoveUsers(ctx context.Context, groupID string, userIDs []string) error {
 	var err error
+
+	group, err := s.repository.GetByID(ctx, groupID)
+	if err != nil {
+		return err
+	}
+
 	for _, userID := range userIDs {
 		// remove all access via policies
 		userPolicies, currentErr := s.policyService.List(ctx, policy.Filter{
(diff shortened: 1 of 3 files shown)


core/group/service.go
```diff
@@ -335,6 +337,13 @@ func (s Service) RemoveUsers(ctx context.Context, groupID string, userIDs []stri
 			err = errors.Join(err, currentErr)
 		}
 	}
+
+	if err == nil {
+		audit.GetAuditor(ctx, schema.PlatformOrgID.String()).LogWithAttrs(audit.GroupMemberRemovedEvent, audit.GroupTarget(groupID), map[string]string{
+			"userIDs": strings.Join(userIDs, ", "),
```
kushsharma, line 343: should we sort these ids to make it deterministic?

core/group/service.go
```diff
@@ -335,6 +337,13 @@ func (s Service) RemoveUsers(ctx context.Context, groupID string, userIDs []stri
 			err = errors.Join(err, currentErr)
 		}
 	}
+
+	if err == nil {
+		audit.GetAuditor(ctx, schema.PlatformOrgID.String()).LogWithAttrs(audit.GroupMemberRemovedEvent, audit.GroupTarget(groupID), map[string]string{
```
kushsharma, line 342: we should pass org id of the group id instead of platform id here?

core/group/service.go
anujk14: I thought of emitting events per user, to avoid losing any events in case of failure in any one user removal or infra crashes. Can you recheck once?

core/group/service.go
anujk14: Done

kushsharma (approved):

PR #642, opened 2024-06-12
chore: unit tests for credit service

anujk14:


billing/credit/service.go
```diff
@@ -32,15 +32,21 @@ func (s Service) Add(ctx context.Context, cred Credit) error {
 		return fmt.Errorf("credit amount is negative")
 	}
 	// check if already credited
-	if t, err := s.transactionRepository.GetByID(ctx, cred.ID); err == nil && t.ID != "" {
+	t, err := s.transactionRepository.GetByID(ctx, cred.ID)
+
+	if err != nil {
```
kushsharma, line 37: this will fail to add transaction if it returns not found from repository for this id. We want to do opposite of it.

anujk14: Yup, missed that. Added a check for it.

billing/credit/service.go
```diff
@@ -32,15 +32,21 @@ func (s Service) Add(ctx context.Context, cred Credit) error {
 		return fmt.Errorf("credit amount is negative")
 	}
 	// check if already credited
-	if t, err := s.transactionRepository.GetByID(ctx, cred.ID); err == nil && t.ID != "" {
+	t, err := s.transactionRepository.GetByID(ctx, cred.ID)
+
+	if err != nil && err != ErrNotFound {
```
kushsharma, line 37: better we use `errors.Is` instead of != check to make sure it unwraps correctly?

anujk14: Done

kushsharma (approved):

PR #643, opened 2024-06-12
feat(admin_ui): add filter to tables and use table loader property

rsbh:


ui/src/containers/invoices.list/columns.tsx
```diff
@@ -16,82 +13,68 @@ const currencyDecimal = 2;
...
       accessorKey: "due_date",
-      cell: isLoading
-        ? () => <Skeleton />
-        : ({ row, getValue }) => {
-            const date = getValue() || row.original.period_end_at;
-            return <Text>{dayjs(date).format("MMM DD, YYYY")}</Text>;
-          },
+      cell: ({ row, getValue }) => {
+        const date = getValue() || row.original.period_end_at;
+        return <Text>{dayjs(date).format("MMM DD, YYYY")}</Text>;
```
anujk14, line 24: I think we had created a constant somewhere for the date format, right? Can we use that here?

anujk14 (approved):

PR #646, opened 2024-06-14
feat: audit events for user removal from org, group and disabling org

anujk14:


core/group/service.go
```diff
@@ -369,5 +369,12 @@ func (s Service) Delete(ctx context.Context, id string) error {
 		return err
 	}
 
+	group, err := s.repository.GetByID(ctx, id)
```
kushsharma, line 372: better we fetch it before relation service? So if in case it doesn't exist, it fail early?

core/organization/service.go
```diff
@@ -288,6 +290,8 @@ func (s Service) RemoveUsers(ctx context.Context, orgID string, userIDs []string
 		}); err != nil {
 			err = errors.Join(err, currentErr)
 		}
+
+		audit.GetAuditor(ctx, orgID).Log(audit.OrgMemberDeletedEvent, audit.UserTarget(userID))
```
kushsharma, line 294: there is already a similar log in API package, should we remove that if we want to add it here?

kushsharma (approved):

core/group/service.go
anujk14: Changed the logic to fetch the group early, and emit the event only after both, relation as well as repository rows have been deleted.

PR #656, opened 2024-06-26
feat: create policy for projects

anujk14:
This PR adds an endpoint to create a policy for a project. 

Corresponding Proton PR: https://github.com/raystack/proton/pull/363


kushsharma (approved):

internal/api/v1beta1/policy.go
```diff
@@ -107,18 +107,47 @@ func (h Handler) CreatePolicy(ctx context.Context, request *frontierv1beta1.Crea
...
+
+	p := policy.Policy{
+		RoleID:        request.GetBody().GetRoleId(),
+		PrincipalType: principalType,
+		PrincipalID:   principalID,
+		ResourceID:    request.GetProjectId(),
+		ResourceType:  schema.ProjectNamespace,
+	}
+
+	newPolicy, err := h.policyService.Create(ctx, p)
```
kushsharma, line 134: Can you verify if this method converts project name to project id automatically? If not we should pass project id here if in case someone has passed project name?

anujk14: Updated

PR #658, opened 2024-07-02
feat(admin_ui): add support to invite users in a org

rsbh:


ui/src/containers/organisations.list/users/index.tsx
```diff
@@ -74,26 +75,48 @@ export default function OrganisationUsers() {
...
+        const [usersResp, invitationResp] = await Promise.all([
+          client?.frontierServiceListOrganizationUsers(orgId, {
+            with_roles: true,
+          }),
+          client?.frontierServiceListOrganizationInvitations(orgId),
+        ]);
+        const userList = usersResp?.data?.users || [];
+        const role_pairs = usersResp?.data?.role_pairs || [];
+        const invitedUsers =
+          invitationResp?.data?.invitations?.map((user) => ({
```
anujk14, line 89: We often have business logic of the kind written above where we fetch data and then filter it in some way on frontend.
Should we have helper functions for things like this? For example, a helper called `filterInvitedUsers(userList)` could do this, and we could reuse it. Since Frontier is under active development, having such helper methods would help us easily change things on frontend if something changes on the backend.

ui/src/containers/organisations.list/users/invite.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }, [organisationId]);
+
+  const onSubmit = async (data: InviteSchemaType) => {
+    try {
+      if (!organisationId) return;
+      await client?.frontierServiceCreateOrganizationInvitation(
+        organisationId,
+        {
+          user_ids: data?.emails,
+          group_ids: data?.team,
```
anujk14, line 57: The API supports adding someone to multiple groups? 😳

ui/src/containers/organisations.list/users/invite.tsx
```diff
@@ -0,0 +1,199 @@
...
+          role_ids: data?.type,
+        }
+      );
+      toast.success("members added");
```
anujk14, line 61: Can we capitalize the `M` here, and write it as `Member/s added`?

ui/src/containers/organisations.list/users/invite.tsx
```diff
@@ -0,0 +1,199 @@
...
+      navigate(`/organisations/${organisationId}/users`);
+    } catch ({ error }: any) {
+      toast.error("Something went wrong", {
```
anujk14, line 64: Can we handle 400s differently here, instead of throwing a generic `Something went wrong`?

ui/src/containers/organisations.list/users/index.tsx
rsbh: We are not filtering users here. We are adding an extra field to invited users and merging it will users

ui/src/containers/organisations.list/users/invite.tsx
rsbh: https://github.com/raystack/proton/blob/main/raystack/frontier/v1beta1/frontier.proto#L3063

anujk14 (approved):
lgtm ⚔️

PR #711, opened 2024-08-06
feat(admin-ui): add webhooks page

rsbh:


paanSinghCoder (approved):

ui/src/components/CustomField.tsx
```diff
@@ -42,82 +43,109 @@ export const CustomFieldName = ({
...
+          }}
+        >
+          <FormLabel>
+            <Text>{inputTitle}</Text>
+          </FormLabel>
+          <FormMessage match="valueMissing">
+            Please enter your {name}
+          </FormMessage>
+          <FormMessage match="typeMismatch">
+            Please provide a valid {title}
```
paanSinghCoder, line 63: `title` is marked as optional in `CustomFieldNameProps`. Should we have a fallback value here or default value in the args?

paanSinghCoder (approved):

ui/src/components/CustomField.tsx
```diff
@@ -5,19 +5,20 @@ import {
   FormLabel,
   FormMessage,
 } from "@radix-ui/react-form";
-import { Flex, Select, Text, TextField } from "@raystack/apsara";
-import { CSSProperties } from "react";
+import { Flex, Select, Switch, Text, TextField } from "@raystack/apsara";
+import React, { CSSProperties } from "react";
```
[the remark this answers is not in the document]
rsbh: we are using `React.RefAttributes<HTMLDivElement>` in this file. Thus, eslint gives an error if we don't import `React`.

ui/src/components/CustomField.tsx
rsbh: yes, we should do it. lets create a task for that

PR #713, opened 2024-08-06
fix: sdk docker build

rsbh:


docker-compose.yml
```diff
@@ -107,5 +107,5 @@ services:
       context: ./sdks/js/packages/sdk-demo
     ports:
       - "3000:3000"
-    depends_on:
```
kushsharma, line 110: why remove the dependency?

kushsharma (approved):

docker-compose.yml
rsbh: The network call will fail, UI will still work and API will start working once Frontier instance is Ready.

PR #715, opened 2024-08-08
fix: return error if user try to accept expired invite

AmanGIT07:
IDE-498

fix: return error if user is tying to accept expired invite. Currently, user can accept expired invite.

### Testing: 

API request:

```
`curl --location 'http://127.0.0.1:9080/v1beta1/organizations/
<org_id: UUID>/invitations/<invataiton_id: UUID>/accept' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Cookie: sid=' \
--data '{}'`
```

Response code:

`400 Bad Request`

### DIFF
--- a/internal/api/v1beta1/invitations.go
+++ b/internal/api/v1beta1/invitations.go
@@ -18,7 +18,10 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-var grpcInvitationNotFoundError = status.Error(codes.NotFound, "invitation not found")
+var (
+	grpcInvitationNotFoundError = status.Error(codes.NotFound, "invitation not found")
+	grpcInvitationExpiredError  = status.Error(codes.InvalidArgument, "invitation expired")
+)
 
 type InvitationService interface {
 	Get(ctx context.Context, id uuid.UUID) (invitation.Invitation, error)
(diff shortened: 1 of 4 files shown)


AmanGIT07:
https://linear.app/pixxel/issue/IDE-499/users-are-able-to-accept-expired-invitations

internal/api/v1beta1/invitations.go
```diff
@@ -246,6 +246,8 @@ func (h Handler) AcceptOrganizationInvitation(ctx context.Context, request *fron
 	if err := h.invitationService.Accept(ctx, inviteID); err != nil {
 		logger.Error(err.Error())
 		switch {
+		case errors.Is(err, invitation.InviteExpired):
+			return nil, grpcBadBodyError
```
anujk14, line 250: The `grpcBadBodyError` returns a very generic "invalid syntax in body" error. However, that does not fit this use case very well.
Instead, can we create and use a custom error here ([refer to grpcInvitationNotFoundError](https://github.com/raystack/frontier/blob/main/internal/api/v1beta1/invitations.go#L21)) that is more apt for the use case?

anujk14 (changes requested):

internal/api/v1beta1/invitations.go
AmanGIT07: Created a new grpc error code `grpcInvitationExpiredError` which returns http status code `400`.


### Testing:

API request:

`curl --location 'http://127.0.0.1:9080/v1beta1/organizations/<org_id>/invitations/<invite_id>/accept' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Cookie: sid= \
--data '{}'`

Response code
`400 Bad Request`

Response body:
`{"code":3,"message":"invitation expired","details":[]}`

anujk14:
Closing this as #716 has been raised instead of this.

PR #718, opened 2024-08-11
fix: context cancel errors are no longer internal errors

kushsharma:
- context cancellation will be treated as `status.Error(codes.Canceled, err.Error())`
- additionally added more histogram buckets for stripe syncs
- new vector `[.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 30, 90, 180, 300, 600]`

### DIFF
--- a/pkg/server/interceptors/errors.go
+++ b/pkg/server/interceptors/errors.go
@@ -1,0 +1,46 @@
+package interceptors
+
+import (
+	"context"
+	"errors"
+
+	grpczap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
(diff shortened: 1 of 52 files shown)


pkg/server/interceptors/errors.go
```diff
@@ -0,0 +1,46 @@
...
+		}
+		return resp, nil
+	}
+}
+
+func errWrapper(ctx context.Context, err error) error {
+	grpczap.Extract(ctx).Error(err.Error())
+
+	// check if grpc status error
+	if _, ok := status.FromError(err); ok {
```
anujk14, line 34: Doesn't `status.FromError` handle `context.canceled` as a known error and return `ok`?

kushsharma: It doesn't handle it, there is a test to verify it as well in `pkg/server/interceptors/errors_test.go`

anujk14 (approved):
LGTM

PR #722, opened 2024-08-13
refactor (client): Move client out of react build

paanSinghCoder:
[ID-518](https://linear.app/pixxel/issue/IDE-518)

**Problem statement:** 
The react package in sdks/package/core is imported even though only the API client is needed. This unnecessarily increases the bundle load on the importing package.

**Solution:**
Moving the API client to its own subpath similar to react sub-package.

- Rename client to api-client.
- Move api-client away from react package.
- Separate rollup config for the api-client.
- Update sdk-demo with api-client usage.
- Rename relative imports from api-client.
- Frontier version increase from v0.28.4 to v0.29.0.

**Testing**
**Ongoing**

**Breaking change**
**YES**

**TODO**
- ~~Rename `client` to something more meaningful.~~
- ~~Import and test `/react` and `/client` separately.~~


paanSinghCoder:
@rsbh Did a basic round of testing. Can you please go through on your end also as this might be a breaking change?

rsbh (approved):
LGTM

PR #724, opened 2024-08-14
feat: add pagination to ListAllOrganizations API

AmanGIT07:
### Overview:
A new package named `pagination` is introduced. It has methods like `Offset` and `SetTotalPages` which will be useful in API that needs pagination. 
`ListAllOrganizations` uses this and returns x amount of records by default. The page number and items per page can be modulated by params: `page_num` & `page_size` respectively. The API response includes `count` denoting the total number of records.

### Testing:

1. Default behaviour

Request:
```
`curl --location 'http://localhost:8400/frontier-api/v1beta1/admin/organizations' \
--header 'Cookie: sid=`
```

Response:
```
`{
    "organizations": [
        {
            "id": "",
            "name": "",
            "title": "",
            "metadata": {},
            "created_at": "",
            "updated_at": "",
            "state": "",
            "avatar": ""
        },
        {
            "id": "",
            "name": "",
            "title": "",
            "metadata": {},
            "created_at": "",
            "updated_at": "",
            "state": "",
            "avatar": ""
        }
    ],
    "count": 2
}`
```

2. Request a specific page with page size

Request:
```
curl --location 'http://localhost:8400/frontier-api/v1beta1/admin/organizations?page_num=1&page_size=10' \
--header 'Cookie: sid='
```

Response:
```
{
    "organizations": [
        {
            "id": "",
            "name": "",
            "title": "",
            "metadata": {},
            "created_at": "",
            "updated_at": "",
            "state": "",
            "avatar": ""
        }
    ],
    "count": 2
}
```

3. Use pagination with existing filter

Request:
```
curl --location 'http://localhost:8400/frontier-api/v1beta1/admin/organizations?state=disabled&page_num=1&page_size=1' \
--header 'Cookie: sid='
```

Response:
```
{
    "organizations": [
        {
            "id": "",
            "name": "",
            "title": "",
            "metadata": {},
            "created_at": "",
            "updated_at": "",
            "state": "disabled",
            "avatar": ""
        }
    ],
    "count": 1
}
```


core/pagination/pagination.go
```diff
@@ -0,0 +1,38 @@
+package pagination
+
```
anujk14, line 2: Pagination would be a better fit for `pkg` instead of `core` since it does not contain any logic specific to Frontier, and can be used fairly independently.

core/pagination/pagination.go
```diff
@@ -0,0 +1,38 @@
...
+	"math"
+)
+
+const (
+	DefaultPageSize = 50
+	DefaultPageNum  = 1
+)
+
+type Pagination struct {
+	PageNum    uint
```
anujk14, line 13: We could keep this to be in the same datatype as what we expect from proton. That way, we won't have to typecast it when using.

internal/store/postgres/organization_repository.go
```diff
@@ -201,7 +201,25 @@ func (r OrganizationRepository) List(ctx context.Context, flt organization.Filte
...
-	query, params, err := stmt.ToSQL()
+	offset := flt.Pagination.Offset()
+	limit := flt.Pagination.PageSize
+
+	totalCountStmt := stmt.Select(goqu.COUNT("*"))
+	totalCountQuery, _, err := totalCountStmt.ToSQL()
+	if err != nil {
+		return []organization.Organization{}, fmt.Errorf("%w: %s", queryErr, err)
+	}
+
```
anujk14, line 212: By default, does this method have any order by clause? If not, we should have one by `created_at DESC` since we're introducing pagination.
This is so that the API response is easier to consume and understand, otherwise we could get a different result each time since we're now returning paginated results.

anujk14 (changes requested):

core/pagination/pagination.go
AmanGIT07: Moved to pkg

core/pagination/pagination.go
AmanGIT07: Updated the datatype to be int32

internal/store/postgres/organization_repository.go
AmanGIT07: Added sorting on `created_at` DESC.

anujk14 (approved):

anujk14 (approved):

anujk14 (approved):

PR #737, opened 2024-08-23
fix: check if already deducted a transaction to avoid 5XX response.

AmanGIT07:
### Description 
When `CreateBillingUsage` API is called multiple times with same ID, it doesn't check if already debuted and tries to create transaction entry. This results in DB error `ERROR: duplicate key value violates unique constraint`,  HTTP status code  `500 `.
Applied a check to see if the ID is present. If no, it's BAU. If yes, it returns an error mentioning `credits already applied` with HTTP status code `409`.

### API Testing
1. Unique ID

Payload:

```
 "usages": [
    {
      "id": "cbaef2b4-eb91-44c0-bdc4-eed02666ea70",
      "customer_id": "",
      "source": "",
      "description": "",
      "type": "credit",
      "amount": "2",
      "user_id": "",
      "metadata": {},
      "created_at": "",
      "updated_at": ""
    }
```

Response status:  `200 OK`

2. Duplicate ID

Payload:

```
 "usages": [
    {
      "id": "cbaef2b4-eb91-44c0-bdc4-eed02666ea70",
      "customer_id": "",
      "source": "",
      "description": "",
      "type": "credit",
      "amount": "2",
      "user_id": "",
      "metadata": {},
      "created_at": "",
      "updated_at": ""
    }
```

Response:
status - `409 Conflict`
body -
```
{
    "code": 6,
    "message": "failed to deduct usage: credits already applied",
    "details": []
}

```

### DIFF
--- a/billing/credit/service.go
+++ b/billing/credit/service.go
@@ -80,6 +80,16 @@ func (s Service) Deduct(ctx context.Context, cred Credit) error {
 		return errors.New("credit amount is negative")
 	}
 
+	//check if already deducted
+	t, err := s.transactionRepository.GetByID(ctx, cred.ID)
+
+	if err != nil && !errors.Is(err, ErrNotFound) {
+		return err
+	}
+	if err == nil && t.ID != "" {
+		return ErrAlreadyApplied
+	}
+
 	// check balance, if enough, sub credits
 	currentBalance, err := s.GetBalance(ctx, cred.CustomerID)
 	if err != nil {
--- a/billing/credit/service.go
+++ b/billing/credit/service.go
@@ -33,22 +33,13 @@ func (s Service) Add(ctx context.Context, cred Credit) error {
 	if cred.Amount < 0 {
 		return errors.New("credit amount is negative")
 	}
-	// check if already credited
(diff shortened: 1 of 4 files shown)


billing/credit/service.go
```diff
@@ -80,6 +80,16 @@ func (s Service) Deduct(ctx context.Context, cred Credit) error {
 		return errors.New("credit amount is negative")
 	}
 
+	//check if already deducted
+	t, err := s.transactionRepository.GetByID(ctx, cred.ID)
```
anujk14, line 84: Instead of adding a DB call, can we try to handle the DB error itself, and return a 409 in that case?
That is, in the repository, when we see a failure, check to see if it is due to "already deduced transaction", and if so, bubble up an appropriate error.

anujk14 (changes requested):

billing/credit/service.go
AmanGIT07: Updated the code to handle the duplicate error at repository level and propagate errors upwards.

anujk14 (approved):

PR #739, opened 2024-08-24
feat: allow trial on default plan of billing account

kushsharma:
When the org is created, if a default billing account of the customer is registered to the billing provider, and if the plan is configured with trial days, the subscription will start as trialing automatically.


ravisuhag:
@kushsharma @anujk14 more context on this please ?

anujk14:
> @kushsharma @anujk14 more context on this please ?

@ravisuhag This will allow us to provide trials by default to users. For example, if we mark enterprise plan as the default one, and it has any `trial days` set, the user will be put on a free trial by default.

This is with reference to the decision doc that was shared earlier with us where we wanted to provide the enterprise trial by default to customers.

PR #752, opened 2024-08-28
feat(sdk): auto generate name field for team and project

rsbh:


anujk14:
@rsbh Do we still need this?

rsbh:
yes we can close this. I don't think we need this.

PR #754, opened 2024-08-28
test: authenticate principal unit tests

kushsharma:
Added a log to track successful syncer runs with billing provider


anujk14 (approved):
LGTM

PR #764, opened 2024-09-04
Test/billing unit tests for customer service

AmanGIT07:
Add unit tests for billing customer service.go


billing/customer/service_test.go
```diff
@@ -0,0 +1,362 @@
...
+	ctx := context.Background()
+	type args struct {
+		customer customer.Customer
+		offline  bool
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    customer.Customer
+		wantErr bool
```
anujk14, line 38: Can we have a field which also indicates what the error is? Currently, it is not very clear what error we would get in different scenarios.
For example, in the first one below, the list API does not return any error. The params sent are valid (barring the duplicacy in orgID). However, we still get an error. It would be better if we could even specify what error to expect in that situation.

Same for all other tests where we have not specified the error type.

anujk14: Here's why this is important:
Unit tests can often act as a documentation of sorts or help in debugging even, if someone completely new to the system tries to look into the code.

In our services even if we simply return back the error raised by repositories, even then, it is essential to mock a particular error and test that that exact error is being returned. This way, whenever anyone wants to debug such an error, they would know that they need to look in the repository.

billing/customer/service_test.go
```diff
@@ -0,0 +1,371 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestService_Update(t *testing.T) {
```
anujk14, line 158: Can we add an error scenario here?

billing/customer/service_test.go
```diff
@@ -0,0 +1,371 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestService_GetByID(t *testing.T) {
```
anujk14, line 251: Can we add an error scenario here?

billing/customer/service_test.go
```diff
@@ -0,0 +1,371 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestService_List(t *testing.T) {
```
anujk14, line 306: Same here.
I understand that we're only calling the repository and returning its error, but the unit tests will ensure that any time someone changes that implementation, they are aware that the contract and behaviour have changed.

anujk14 (changes requested):

anujk14 (approved):

PR #770, opened 2024-09-09
chore: identify APIs that can be moved to parent component

paanSinghCoder:
Identify API calls which can be potentially moved to top level.


anujk14:
@paanSinghCoder Do we still need this? Can we move relevant changes recommended here to a task and close this?

paanSinghCoder:
@anujk14 IMO these are not must have changes thus can be treated as very low priority and can even be discarded.

anujk14:
> @anujk14 IMO these are not must have changes thus can be treated as very low priority and can even be discarded.

@paanSinghCoder Since we had put in effort in identifying these instances, I believe we should add this to a task so that we don't end up spending time again on identifying the same thing again.

PR #771, opened 2024-09-10
fix: set page size 1000 till filter is implemented in org list API

paanSinghCoder:
Set organisation API page_size to 1000 till filters are supported on the BE.


ui/src/containers/projects.create/index.tsx
```diff
@@ -20,6 +20,10 @@ import { CustomFieldName } from "~/components/CustomField";
 import { SheetFooter } from "~/components/sheet/footer";
 import { SheetHeader } from "~/components/sheet/header";
 
+// Setting this to 1000 initially till APIs support filters and sorting.
```
anujk14, line 23: Let's rephrase the comment as a `TODO` so that we can find it easily later.
Same for the one above.

anujk14 (approved):

PR #780, opened 2024-09-19
chore: Added useTokens hook in package/core/react

singh-pk:


ui/src/containers/organisations.list/billingaccounts/details.tsx
```diff
@@ -2,8 +2,8 @@ import { Flex, Grid, Text } from "@raystack/apsara";
 import { NavLink, useParams } from "react-router-dom";
 import { useBillingAccount } from ".";
 import { BillingAccountAddress } from "@raystack/frontier";
+import { useTokens } from "@raystack/frontier/react";
```
rsbh, line 5: We are in the process of removing `@raystack/frontier/react` as a dependency in `/ui.`
can you keep the revert the changes in the `ui` folder back?

the `useTokens` from `@raystack/frontier/react` will not work here as there is no `activeOrganization, billingAccount` in the context.

rsbh (changes requested):

rsbh (approved):

PR #781, opened 2024-09-19
feat: set billing customer credit min limit

kushsharma:
- If a billing account credit min limit is set less then 0, it works as overdraft. Default is 0.
- If a org billing account is in overdraft state, it's not allowed to create new billing account unless the dues are settled.
- Setting this limit more then 0 works as min purchase requirement for the account.
- Only platform admins can change the account limits.
- https://github.com/raystack/proton/pull/370


internal/store/postgres/migrations/20240917183506_billing_customer_credit_min.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE billing_customers ADD COLUMN IF NOT EXISTS credit_min int8 DEFAULT 0;
```
anujk14, line 1: Can we rename `credit_min` to something more descriptive like `min_balance_allowed`?
`credit_min` does not seem very intuitive, IMO.

kushsharma: Open for suggestions but `min_balance_allowed` doesn't look better IMO, balance of what?

anujk14: Balance of the billing account (since it would be on the billing account table)

anujk14: Or `min_credit_balance_allowed`? (I was thinking `min_credit_allowed` , but that gives an impression like that's the minimum amount of credit you can transfer into the account at one time)

kushsharma: `credits` and `balance` means a similar things in this billing system so if we name `min_credit_balance`, one of these two words look redundant. Then the suffix of `_allowed` again I feel is redundant as we will not store a config which is not allowed and the `min` part makes it call out that it's the minimum we will allow. I agree, on the `min_credit` vs `credit_min` part though, I was also confused what should I go with, the reason I didn't go with `min_credit` is same as what you just described, it gives the impression as a minimum limit you can transact.

Any more suggestions?

kushsharma: After discussion on wire transfer, we might need to maintain another column for customer balance as well. So using credit and balance on one name would be conflicting.

anujk14 (approved):

PR #782, opened 2024-09-23
feat: reduce DB calls in plan list API

anujk14:
- Introduced a method called `ListWithProducts` in Plans repository, which fetches plans along with the products that belong to the plan
- Plan list method in plan service now uses `ListWithProducts`, to reduce the number of DB calls that are made to fetch plans+products+features.


billing/plan/service.go
```diff
@@ -84,22 +91,26 @@ func (s Service) GetByID(ctx context.Context, id string) (Plan, error) {
...
-		products, err := s.productService.List(ctx, product.Filter{
-			PlanID: listedPlan.ID,
-		})
-		if err != nil {
-			return nil, err
+
+	// Initialize a map of {productID: []feature.Feature}
+	productFeatures := initializeProductFeatureMap(plans)
+
+	features, _ := s.featureRepository.List(ctx, product.Filter{})
```
kushsharma, line 102: why err ignored?

billing/plan/service.go
```diff
@@ -327,3 +338,23 @@ func verifyDuplicatePlans(planFile File) error {
...
+	for _, pln := range p {
+		products := pln.Products
+		for _, prod := range products {
+			productFeatures[prod.ID] = []product.Feature{}
+		}
+	}
+	return productFeatures
+}
+
+func mapFeaturesToProducts(productFeatureMap map[string][]product.Feature, features []product.Feature) {
```
kushsharma, line 353: nit: IMO creating functions for 3-4 lines being used only at one place is not much of a use and code is easier to read when it's a single consistent flow. Also it's better to not modify the args, and instead return the new value.

internal/store/postgres/billing_plan_repository.go
```diff
@@ -273,3 +349,112 @@ func (r BillingPlanRepository) List(ctx context.Context, filter plan.Filter) ([]
...
+		prd.Col("state").As("product_state"),
+		prd.Col("metadata").As("product_metadata"),
+		prd.Col("created_at").As("product_created_at"),
+		prd.Col("updated_at").As("product_updated_at"),
+		prd.Col("deleted_at").As("product_deleted_at"),
+	)
+
+	var ids []string
+	var names []string
+	if len(filter.IDs) > 0 {
```
kushsharma, line 391: the filter is taking an array but we filter only on 0th index?

anujk14: We check the ids field to see if we need to filter by name or an id. I thought of making it work only with ids, but decided to keep it consistent with the `List` method in this repository. 
If it is safe to remove it from there also, then I think we could make both methods work with only ids.

billing/plan/service.go
anujk14: By mistake. Handled now.

billing/plan/service.go
anujk14: > Also it's better to not modify the args, and instead return the new value.
Agreed, updated to not do that

In terms of methods, I believe smaller methods which perform non-functional conversions/translations can sometimes distract from the actual logic of what the method is doing. Hence, separated out the mapping into a separate method. It can also help when debugging since there is a clear separation of parts in the method.

kushsharma (approved):

PR #784, opened 2024-09-26
test: add unit test for StartFlow in authenticate service

AmanGIT07:


core/authenticate/service_test.go
```diff
@@ -290,3 +297,129 @@ func TestService_GetPrincipal(t *testing.T) {
...
+					},
+					mockFlowRepo, mockDialer, nil, nil,
+					nil, nil, nil)
+				srv.Now = func() time.Time {
+					return timeNow
+				}
+				return srv
+			},
+		},
+	}
```
anujk14, line 411: Can we add a case for SendMail failure as well?

AmanGIT07: Added the failure case for SendMail

anujk14 (approved):
LGTM

PR #796, opened 2024-10-20
feat: generate invoice for overdraft credits

kushsharma:
It's necessary to configure the product name to calculate per unit price before the overdraft credits can be invoiced in `billing.customer.credit_overdraft_product`. If not set, invoice reconcilation and creation is skipped.

If there is already an invoice for overdraft credits for same time range, no new invoice will be created for that customer. Currently the system generates them on the first day of the month.
Reconciliation of these invoice happens on same cadence as invoices gets synced from the billing provider.

- https://github.com/raystack/proton/pull/373


billing/invoice/service.go
```diff
@@ -177,6 +256,37 @@ func (s *Service) SyncWithProvider(ctx context.Context, customr customer.Custome
 	return nil
 }
 
+func (s *Service) upsert(ctx context.Context, customerID string,
+	existingInvoice *Invoice, stripeInvoice *stripe.Invoice) error {
+	if existingInvoice != nil {
+		// already present in our system, update it if needed
```
anujk14, line 262: If month 1's invoice was not paid, are we updating the invoice of month 1 and adding balance of month 2?

internal/store/postgres/billing_transactions_repository.go
```diff
@@ -388,9 +422,28 @@ func (r BillingTransactionRepository) getBalanceInTx(ctx context.Context, tx *sq
...
+		var err error
+		amount, err = r.getBalanceInTx(ctx, tx, accountID, nil, nil)
+		return err
+	}); err != nil {
+		return 0, fmt.Errorf("failed to get balance: %w", err)
+	}
+	return amount, nil
+}
+
+// GetBalanceForRange returns the balance of the account in the given range.
```
anujk14, line 437: Per our discussion, we had decided to keep a balance for the start of the month, and calculate `end balance - start balance` to figure out if there is an overdraft (similar to a banking system).
If we don't keep the starting balance, won't we end up with incorrect overdrafts? For example, if I already had 1000 tokens in my account at the start of September (from token additions in previous months). And in September I spend 200 tokens. It would appear that my balance for September is `-200`, but it is actually `+800` due to earlier balance. 

Plainly going off the comment. Let me know in case I've understood it incorrectly.

anujk14 (changes requested):

internal/store/postgres/billing_transactions_repository.go
```diff
@@ -388,9 +422,28 @@ func (r BillingTransactionRepository) getBalanceInTx(ctx context.Context, tx *sq
 // in transaction table till now.
 func (r BillingTransactionRepository) GetBalance(ctx context.Context, accountID string) (int64, error) {
 	var amount int64
-	if err := r.dbc.WithTxn(ctx, sql.TxOptions{}, func(tx *sqlx.Tx) error {
+	if err := r.dbc.WithTxn(ctx, sql.TxOptions{
+		Isolation: sql.LevelSerializable,
```
AmanGIT07, line 426: This is done to prevent race condition?

internal/store/postgres/billing_transactions_repository.go
anujk14: Discussed offline

billing/invoice/service.go
anujk14: Discussed offline

internal/store/postgres/billing_transactions_repository.go
AmanGIT07: I feel it will make system slow once we start to have good amount of transactions. Also we calculate balance by summing over all rows instead which may also get performance issues.

kushsharma: > Is this done to prevent race condition?

Yes, this is to prevent race conditions. What other way do you propose here to prevent it?

> I feel it will make system slow once we start to have good amount of transactions. Also we calculate balance by summing over all rows instead which may also get performance issues.

Yes, one way I was thinking was to maintain a monthly statement in system and only accumulate records since last statement. But this introduces complexity quite a lot and was postponing it until we hit the performance degradation. I want to know what part of it is the bottleneck before going ahead with this approach. Do you have anything else in mind?

anujk14 (approved):

PR #797, opened 2024-10-21
fix: create org member event when the org is created

kushsharma:
`OrgMemberCreatedEvent` was not getting emitted in some of the situations, it should do now.


core/organization/service.go
```diff
@@ -184,6 +184,11 @@ func (s Service) AddMember(ctx context.Context, orgID, relationName string, prin
 	}); err != nil {
 		return err
 	}
+
+	audit.GetAuditor(ctx, orgID).Log(audit.OrgMemberCreatedEvent, audit.Target{
```
anujk14, line 188: Do we have events for whenever a user is added via SSO login? (Or any other way of adding a user)
We need to emit an event for all those scenarios.

kushsharma: All the usecases should be calling this method so should be covered.

anujk14, line 188: This event is currently emitted in the org handler as well in the `AddOrganizationUsers` method. Can we remove it from there to avoid duplicate events?

anujk14 (changes requested):

core/organization/service.go
kushsharma: Can't find any other place, can you share direct git link?

anujk14: https://github.com/raystack/frontier/blob/e913c1c9dbc545951344bdccbd00c0bc776c4b1f/internal/api/v1beta1/org.go#L386

kushsharma: I dont get it, why do you think it's not removed?

kushsharma: Screenshot 2024-10-28 at 3 01 52 PM

anujk14: Missed that earlier while reviewing. Looks good.

anujk14 (approved):

PR #804, opened 2024-10-24
chore: fix files to consider for test coverage

anujk14:


Makefile
```diff
@@ -30,7 +30,7 @@ lint-fix:
 	golangci-lint run --fix
 
 test: ## Run tests
-	@go test -race $(shell go list ./... | grep -v /ui | grep -v /vendor/ | grep -v /test/) -coverprofile=coverage.out -count 2 -timeout 150s
+	@go test -race $(shell go list ./... | grep -v /ui | grep -v /vendor/ | grep -v /test/ | grep -v mocks | grep -v postgres/migrations) -coverprofile=coverage.out -count 2 -timeout 150s
```
AmanGIT07, line 33: just specifying `mocks` works? or should it be `/mocks` ?

anujk14: Unless we have a file named `mocks`, we should be fine. But I'll change it to `/mocks` nonetheless for a stricter check.

AmanGIT07 (approved):
LGTM

PR #808, opened 2024-10-28
fix: check for resource.ErrNotExist error in check permission

rsbh:
Problem.
If a random ID is passed in the `CheckFederatedResourcePermission` API `resource` body. The API throws an `Internal` error. And in the server logs, the error is `resource doesn't exist`. 
After debugging, I found that we [fetch the resource from the db](https://github.com/raystack/frontier/blob/main/core/resource/service.go#L222), and it returns `ErrNotExist` error.
This error was not handled in the API handler function.

This PR adds a check for the `resource.ErrNotExist` error. It also refactors the switch statement.


anujk14 (approved):
lgtm

PR #810, opened 2024-10-31
feat: role filter while listing organization users

kushsharma:
- https://github.com/raystack/proton/pull/371
- a list of role name/ids are accepted as inputs, it can't be used when `with_roles` is toggled on.

### DIFF
--- a/internal/api/v1beta1/org.go
+++ b/internal/api/v1beta1/org.go
@@ -3,26 +3,20 @@ package v1beta1
 import (
 	"context"
 
-	"github.com/raystack/frontier/core/serviceuser"
-
-	"github.com/raystack/frontier/core/authenticate"
-
-	"go.uber.org/zap"
-
+	grpczap "github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap"
+	"github.com/pkg/errors"
 	"github.com/raystack/frontier/core/audit"
+	"github.com/raystack/frontier/core/authenticate"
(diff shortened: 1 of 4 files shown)


rsbh (approved):

internal/api/v1beta1/org.go
```diff
@@ -259,9 +261,67 @@ func (h Handler) ListOrganizationUsers(ctx context.Context, request *frontierv1b
...
 	}
 
-	users, err := h.userService.ListByOrg(ctx, orgResp.ID, request.GetPermissionFilter())
-	if err != nil {
-		return nil, err
+	var users []user.User
+	var rolePairPBs []*frontierv1beta1.ListOrganizationUsersResponse_RolePair
+
+	if len(request.GetRoleFilters()) > 0 {
+		// convert role names to ids if needed
```
anujk14, line 268: Can we make this compatible with `with_roles` field as well?

anujk14 (changes requested):

internal/api/v1beta1/org.go
anujk14: Because as an API contract, we should not have completely different behaviours depending on the filters being passed.

kushsharma: it doesn't make sense to return roles when you are querying by roles.

kushsharma: Added validation to fail if both are passed.

anujk14 (approved):

PR #824, opened 2024-11-29
fix: add isolation level in transaction repository

kushsharma:
Currently there were chances to have race conditions while writing to transaction repository. I have added a test to verify it doesn't happen. Database is using `Serializable` as the isolation level to avoid overlapping transactions.

### DIFF
--- a/internal/store/postgres/billing_transactions_repository.go
+++ b/internal/store/postgres/billing_transactions_repository.go
@@ -6,10 +6,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 	"strings"
 	"time"
 
 	"github.com/raystack/frontier/billing/customer"
+
 	"github.com/raystack/frontier/internal/bootstrap/schema"
 
 	"github.com/jackc/pgconn"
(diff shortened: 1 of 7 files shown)


internal/store/postgres/billing_transactions_repository.go
```diff
@@ -388,11 +404,17 @@ func (r BillingTransactionRepository) getBalanceInTx(ctx context.Context, tx *sq
...
 func (r BillingTransactionRepository) GetBalance(ctx context.Context, accountID string) (int64, error) {
 	var amount int64
-	if err := r.dbc.WithTxn(ctx, sql.TxOptions{}, func(tx *sqlx.Tx) error {
-		var err error
-		amount, err = r.getBalanceInTx(ctx, tx, accountID)
-		return err
-	}); err != nil {
+	err := r.withRetry(ctx, func() error {
+		return r.dbc.WithTxn(ctx, sql.TxOptions{
+			Isolation: sql.LevelRepeatableRead,
```
AmanGIT07, line 409: So if I understand correctly, we check balance by always looping over the table and calculating difference. We insert new rows for debit/credit.  How will `sql.LevelRepeatableRead` prevent race condition here? `Phantom reads` and `Write Skew` are allowed in `Repeatable Read` as far I can see. It won't prevent new rows being written and hence change of balance mid transaction. 
Can you test if this isolation is working fine? Maybe write a Python or Go code and hit the API concurrently 20-30 times for the same org. If let's say starting balance is 105, and we deduct 10 credit each call then only 10 calls should succeed and the balance should be 5 at the end.

kushsharma: Agree, Serializable would be best and safest(with some added latency), that's what I proposed in the last PR. Although I don't see the issue happening in the test https://github.com/raystack/frontier/pull/824/files#diff-aaeca6450e0dcd11871799ac2b14da4225d7a76f94c04a9d4636e44ae780773eR1115

AmanGIT07: Tested it against an external script that simulates concurrent transaction request. `LevelRepeatableRead` isloation failed at that case. `LevelSerializable` worked fine. Lets use that itself.

AmanGIT07 (approved):

PR #833, opened 2024-12-13
create model method to check default plan is free or not + add unit tests

AmanGIT07:


core/event/service.go
```diff
@@ -172,6 +170,17 @@ func (p *Service) EnsureDefaultPlan(ctx context.Context, orgID string) error {
 	return nil
 }
 
+func (p *Service) isDefaultPlanFree(defaultPlan plan.Plan) bool {
```
anujk14, line 173: @AmanGIT07 The idea behind the task was to create a general method which checks if the plan passed to it is a free plan or not. Not just for the default plan.
The code is alright, but can we do the following please:
1. Rename this to remove binding with "DefaultPlan"
2. Add the method on the plan model file. That way we could use it as `myPlan.isFree()`

anujk14 (approved):
lgtm

core/event/service.go
AmanGIT07: Done

PR #843, opened 2025-01-06
feat: cancel subscription after trial

rsbh:
This PR will update the `end_behavior` in the cancel upcoming phase 
If the user is on trial and config.subscription.cancel_after_trial is set to true, then the `end_behavior` in [subscription update call ](https://docs.stripe.com/api/subscription_schedules/object#subscription_schedule_object-end_behavior)will be set to `cancel`

### DIFF
--- a/billing/config.go
+++ b/billing/config.go
@@ -55,8 +55,8 @@ type PlanChangeConfig struct {
 }
 
 type SubscriptionConfig struct {
-	// CancelAfterTrial will cancel the subscription after trial end and will not generate invoice.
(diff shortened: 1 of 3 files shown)


billing/config.go
```diff
@@ -55,6 +55,8 @@ type PlanChangeConfig struct {
 }
 
 type SubscriptionConfig struct {
+	// CancelAfterTrial will cancel the subscription after trial end and will not generate invoice.
+	CancelAfterTrial bool `yaml:"cancel_after_trial" mapstructure:"cancel_after_trial" default:"false"`
```
kushsharma, line 59: instead should we call it `BehaviorAfterTrial`? and allow option as release/cancel?

kushsharma (approved):

PR #850, opened 2025-01-15
fix(catch): catch Authenticate api error

rsbh:


paanSinghCoder:
There is small typo. Should be `visible`

`const [visiable, setVisiable] = useState<boolean>(open);`

sdks/js/packages/core/react/components/onboarding/magiclink.tsx
```diff
@@ -73,11 +75,17 @@ export const MagicLink = ({ open = false, ...props }: MagicLinkProps) => {
         window.location = `${
           config.redirectMagicLinkVerify
         }?${searchParams.toString()}`;
+      } catch (err: unknown) {
```
paanSinghCoder, line 78: What will happen if control doesn't go in the if block? Should we add an else as well?

```
else {
    setError('email', { message: 'An unexpected error occurred' });
  }
```

paanSinghCoder (approved):

PR #877, opened 2025-02-26
feat: enable Org search aggregated joined with billing Details

whoAbhishekSah:
# Description

This PR adds an admin API, `POST /v1beta1/admin/organizations/search,` which returns an Org list matching the input parameters. The input parameters are specified inside the request body as per the RQL input schema defined [here](https://github.com/raystack/salt/tree/main/rql)

# Testing


rsbh (approved):
🚀  LGTM 🔥

PR #883, opened 2025-03-09
feat: export Org data with billing info

whoAbhishekSah:
## What does this PR do?
This PR adds a new admin endpoint that allows exporting organization data including billing information as a CSV file. The data is streamed in 200KB chunks using gRPC streaming to efficiently handle potentially large datasets.

It also fixes JSONB key value where an extra "\" was appearing in the value.
It also makes the search query case insensitive.

## Implementation Details
- Added new RPC method `ExportOrganizations` in the AdminService 
- Implements streaming response using `google.api.HttpBody` for chunked data transfer
- Generates CSV with comprehensive organization and billing data including:
  - Organization details (ID, name, title)
  - Billing information (plan, payment mode, subscription details)
  - Geographic info (country, state)
  - Temporal data (created/updated timestamps)

## Technical Details
- CSV generation handled by `Service.Export()` which:
  - Fetches org billing data from repository
  - Writes data to CSV buffer with appropriate headers
  - Returns byte array and content type
- gRPC handler streams the CSV in 200KB chunks via `AdminService_ExportOrganizationsServer`
- Uses `google.api.HttpBody` to maintain proper content type through streaming

## CSV Fields
The export includes the following columns:
```csv
Organization ID, Name, Title, Created By, Plan Name, Payment Mode, Country, State, Created At, Updated At, Subscription Cycle End, Subscription State, Plan Interval
```

## Usage
```bash
curl --location 'localhost:8000/v1beta1/admin/organizations/export' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: Basic blah' \
--header 'Cookie: sid=foobar'
```

## Testing
- Tested CSV generation with sample organization data
- Verified chunked streaming functionality
- Validated CSV format and content accuracy

Related PR: https://github.com/raystack/proton/pull/384


core/aggregates/orgbilling/service.go
```diff
@@ -64,3 +68,65 @@ type AggregatedOrganization struct {
 func (s Service) Search(ctx context.Context, query *rql.Query) (OrgBilling, error) {
 	return s.repository.Search(ctx, query)
 }
+
+func (s Service) Export(ctx context.Context) ([]byte, string, error) {
+	orgBillingData, err := s.repository.Search(ctx, &rql.Query{})
+	if err != nil {
+		return nil, "", err
+	}
+	if err != nil {
```
rsbh, line 77: duplicate code

core/aggregates/orgbilling/service.go
```diff
@@ -64,3 +68,65 @@ type AggregatedOrganization struct {
...
+		"Created By",
+		"Plan Name",
+		"Payment Mode",
+		"Country",
+		"State",
+		"Created At",
+		"Updated At",
+		"Subscription Cycle End",
+		"Subscription State",
+		"Plan Interval",
```
rohilsurana, line 98: Any way we can get this from struct field names itself?
This is basically hardcoding and seems like something can be missed in future.
It would be great if we can get it from json struct tags itself, even if the names do not have proper case.

core/aggregates/orgbilling/service.go
```diff
@@ -64,3 +68,65 @@ type AggregatedOrganization struct {
...
+			org.CreatedBy,
+			org.PlanName,
+			org.PaymentMode,
+			org.Country,
+			string(org.State),
+			org.CreatedAt.Format(time.RFC3339),
+			org.UpdatedAt.Format(time.RFC3339),
+			org.SubscriptionCycleEndAt.Format(time.RFC3339),
+			org.SubscriptionState,
+			org.PlanInterval,
```
rohilsurana, line 120: can we make this a struct instead?

core/aggregates/orgbilling/service.go
whoAbhishekSah: can you check now ?

PR #897, opened 2025-03-13
feat(admin-ui): add org details layout with sidepanel

rsbh:
This PR will add the layout for new org details pages.
It also add a side panel on right side, which show the following details.
1. Org details
2. KYC details
3. Plan details
4. Tokens
5. Billing Details


ui/package.json
```diff
@@ -14,7 +14,7 @@
     "@hookform/resolvers": "^3.0.1",
     "@radix-ui/react-form": "^0.0.2",
     "@radix-ui/react-icons": "^1.3.0",
-    "@raystack/apsara": "^0.34.2",
+    "@raystack/apsara": "^0.37.0",
```
paanSinghCoder, line 17: Can be updated to latest version. It has some fixes.

paanSinghCoder (changes requested):

ui/package.json
rsbh: i will do in next PR

PR #914, opened 2025-03-21
fix: validate preference value for newsletter

AmanGIT07:


core/preference/service.go
```diff
@@ -40,6 +42,9 @@ func (s *Service) Create(ctx context.Context, preference Preference) (Preference
 	if !allowCreate {
 		return Preference{}, ErrTraitNotFound
 	}
+	if preference.Name == UserNewsletter && !slices.Contains(newsletterPreferenceAllowedValues, preference.Value) {
```
rohilsurana, line 45: Let's not add a specific check for just the newsletter. Let's add a generic way to validate at the traits level itself?

rohilsurana (approved):

PR #924, opened 2025-03-25
feat: add project user search API

whoAbhishekSah:


internal/api/v1beta1/project_users.go
```diff
@@ -0,0 +1,91 @@
...
+		})
+	}
+
+	return &frontierv1beta1.SearchProjectUsersResponse{
+		ProjectUsers: projectUsers,
+		Pagination: &frontierv1beta1.RQLQueryPaginationResponse{
+			Offset: uint32(projectUsersData.Pagination.Offset),
+			Limit:  uint32(projectUsersData.Pagination.Limit),
+		},
+		Group: &frontierv1beta1.RQLQueryGroupResponse{
```
rohilsurana, line 72: do we need this if group by is disabled?

internal/api/v1beta1/project_users.go
```diff
@@ -0,0 +1,91 @@
...
+		projectUsers = append(projectUsers, transformAggregatedProjectUserToPB(v))
+	}
+
+	groupResponse := make([]*frontierv1beta1.RQLQueryGroupData, 0)
+	for _, groupItem := range projectUsersData.Group.Data {
+		groupResponse = append(groupResponse, &frontierv1beta1.RQLQueryGroupData{
+			Name:  groupItem.Name,
+			Count: uint32(groupItem.Count),
+		})
+	}
```
rohilsurana, line 64: do we need this if group by is disabled?

internal/api/v1beta1/project_users.go
whoAbhishekSah: we can remove it

rohilsurana (approved):

PR #928, opened 2025-04-02
fix: in and notin operator processing inside org billing search API 

whoAbhishekSah:
**Improve IN/NOT IN operator handling with explicit type casting**

Changes:
- Separated IN and NOT IN operator cases for better clarity
- Added explicit TEXT type casting for column values
- Replaced generic operator mapping with dedicated In/NotIn methods

This change ensures consistent text-based comparison for IN/NOT IN operations, preventing potential type mismatch issues and improving query reliability.


internal/store/postgres/org_billing_repository.go
```diff
@@ -403,11 +403,12 @@ func processStringDataType(filter rql.Filter, query *goqu.SelectDataset) *goqu.S
 		query = query.Where(goqu.L(fmt.Sprintf("coalesce(%s, '') = ''", filter.Name)))
 	case OPERATOR_NOT_EMPTY:
 		query = query.Where(goqu.L(fmt.Sprintf("coalesce(%s, '') != ''", filter.Name)))
-	case OPERATOR_IN, OPERATOR_NOT_IN:
-		// process the values of in and notin operators as comma separated list
-		query = query.Where(goqu.Ex{
-			filter.Name: goqu.Op{filter.Operator: strings.Split(filter.Value.(string), ",")},
-		})
+	case OPERATOR_IN:
```
punit-kulal, line 406: can we add these operator constants in rql since these are defined in RQL ?

internal/store/postgres/org_billing_repository.go
```diff
@@ -403,11 +403,12 @@ func processStringDataType(filter rql.Filter, query *goqu.SelectDataset) *goqu.S
...
+		// process the values of in operator as comma separated list
+		query = query.Where(goqu.Cast(goqu.I(filter.Name), "TEXT").In(strings.Split(filter.Value.(string), ",")))
```
punit-kulal, line 408: Are we supposed to give ,  separated strings and not an array ? Isn't array more semantically correct for an in operator (since we can't pass any value with `,` if we take as string.)

whoAbhishekSah: RQL doesn't support Arrays yet. For now, we are going with comma-separated strings until there is proper support for Arrays in RQL.

internal/store/postgres/org_billing_repository.go
whoAbhishekSah: These are already added: https://github.com/raystack/salt/blob/main/rql/parser.go#L12

punit-kulal: No, I meant as constants so that they can be referred instead of each rql user implementing them.

internal/store/postgres/org_billing_repository.go
punit-kulal: okay

punit-kulal (approved):

rsbh (approved):

PR #980, opened 2025-04-30
fix: make like and notlike case insensitive in User Search API

whoAbhishekSah:
## Description

- Make user search filter case insensitive by using postgres ilike/notilike operators                                                                                                                        
  - Improves UX by allowing case-insensitive text matching when filtering users                                                                                                                                  

## Testing                                                                                                                                                                             
  - Test user search with mixed case input matches users regardless of case                                                                                                                                      
  - Verify notlike operator also works case-insensitively                                                                                                                                                        
 - Check existing test cases pass with updated SQL queries


rohanchkrabrty (approved):
LGTM

PR #1036, opened 2025-06-24
implimented strategy pattern for start and finish flow

Mental-Block:
removed a bunch of if statements in flow in favour of implementing strategy pattern for different auth flow methods. all passes test.


rohilsurana:
Closing this for now, we might plan to try refactoring this later.

PR #1038, opened 2025-06-24
feat: implement buf connect server

whoAbhishekSah:
# Add Buf Connect Server Support

## Summary
- Added Buf Connect server support alongside the existing gRPC server
- The connect server runs on port 8002 (configurable)
- All API endpoints are exposed via the Connect protocol with unimplemented stubs
- Add Healthchecks
- **Migrate GolangCI lint to v2. It's a breaking change of lint configuration.**

## Changes
- Added Connect server configuration in pkg/server/config.go
- Created Connect server implementation in pkg/server/server.go
- Added dummy implementations for all Frontier and Admin service APIs
- Service runs in parallel with existing gRPC (8001) and HTTP (8080) servers
- Updated GolangCI Lint to fix the repeated failure of the Lint Job on GitHub Actions.

## Test Plan
- [x] Connect server starts successfully on port 8081
- [x] Verify server health with a simple API call:
```bash
curl \
  --header "Content-Type: application/json" \
  --header "Connect-Protocol-Version: 1" \
  --data '{}' \
  http://localhost:8002/raystack.frontier.v1beta1.FrontierService/ListUsers
```
- [x] Verify unimplemented error response is returned
- [x] Test concurrent operation with gRPC server
- [x] Healthcheck verification

```bash
curl --header "Content-Type: application/json" \
  --data '{"service":"raystack.frontier.v1beta1.AdminService"}' \
  http://localhost:8002/grpc.health.v1.Health/Check
```

## Future Work
- Implement actual API handlers for the Connect server
- Add authentication and authorisation middleware
- Add integration tests for Connect endpoints

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -61,6 +61,10 @@ import (
 
 const (
 	grpcDialTimeout = 5 * time.Second
+
+	// keeping it in sync with https://github.com/raystack/salt/blob/v0.3.8/mux/mux.go#L15
+	// which is being used in GRPC server shutdown
+	connectServerGracePeriod = 10 * time.Second
 )
 
 type UIConfigApiResponse struct {
@@ -155,7 +159,7 @@ func ServeConnect(ctx context.Context, logger log.Logger, cfg ConnectConfig, dep
 	go func() {
 		<-ctx.Done()
 
-		ctxShutdown, cancel := context.WithTimeout(context.Background(), time.Minute)
+		ctxShutdown, cancel := context.WithTimeout(context.Background(), connectServerGracePeriod)
 		defer cancel()
 
 		if err := server.Shutdown(ctxShutdown); err != nil {


internal/api/v1beta1connect/frontier_service_impl.go
```diff
@@ -0,0 +1,672 @@
...
+import (
+	"context"
+
+	"github.com/bufbuild/connect-go"
+	frontierv1beta1 "github.com/raystack/frontier/proto/v1beta1"
+)
+
+func (h *ConnectHandler) ListUsers(context.Context, *connect.Request[frontierv1beta1.ListUsersRequest]) (*connect.Response[frontierv1beta1.ListUsersResponse], error) {
+	return nil, connect.NewError(connect.CodeUnimplemented, nil)
+}
```
whoAbhishekSah, line 12: We will move these functions to separate files as we implement them individually.

whoAbhishekSah: Example implementation

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM alpine:3.18
 
+RUN apk --no-cache add curl
```
whoAbhishekSah: `curl` is required to run the K8S healthcheck to make an HTTP Post call to the Buf connect server.
```yaml
livenessProbe:
  exec:
    command:
      - curl
      - -X
      - POST
      - -H
      - "Content-Type: application/json"
      - -d
      - '{"service":"raystack.frontier.v1beta1.FrontierService"}'
      - http://localhost:8002/grpc.health.v1.Health/Check
  initialDelaySeconds: 15
  periodSeconds: 10
```

rsbh: Why can't we directly use k8s gRPC liveness probe ?

whoAbhishekSah: I wasn't aware Kube supported gRPC health checks. We won't need curl then. Let me remove this change.

rsbh (approved):

cmd/serve.go
```diff
@@ -305,7 +305,14 @@ func StartServer(logger *log.Zap, cfg *config.Frontier) error {
 
 	go server.ServeUI(ctx, logger, cfg.UI, cfg.App)
 
-	// serving server
+	// start connect server
+	go func() {
+		if err := server.ServeConnect(ctx, logger, cfg.App.Connect, deps); err != nil {
```
rohilsurana, line 310: it does not seem like we are performing a graceful shutdown here. Can you please check on this?

whoAbhishekSah: Yes, great catch. Thanks. Added the shutdown part.

pkg/server/server.go
```diff
@@ -115,6 +121,58 @@ func ServeUI(ctx context.Context, logger log.Logger, uiConfig UIConfig, apiServe
...
+		Addr:    fmt.Sprintf(":%d", cfg.Port),
+		Handler: h2c.NewHandler(mux, &http2.Server{}),
+	}
+
+	logger.Info("connect server starting", "port", cfg.Port)
+
+	go func() {
+		<-ctx.Done()
+
+		ctxShutdown, cancel := context.WithTimeout(context.Background(), time.Minute)
```
rohilsurana: graceful shutdown period is hardcoded for now, do we want to use similar config as mux?

PR #1049, opened 2025-07-02
feat: add payment method config to only allow specified payment methods in checkouts and invoices

rohilsurana:
Introduces payment method configs, adjustments to checkout create logic.

### Payment method and currency handling:

* **`billing/config.go`:** Added a new `PaymentMethodConfig` type to define payment method constraints based on minimum and maximum amounts. Updated the `Config` struct to include `PaymentMethodConfig` and `DefaultCurrency`. 
* **`billing/checkout/service.go`:** Incorporated `defaultCurrency` and `paymentMethodConfig` into the `Service` struct and updated checkout creation logic to filter payment methods based on subtotal amounts.
* **`billing/invoice/service.go`:** Added similar support for `defaultCurrency` and `paymentMethodConfig` in invoice creation, including logic to determine allowed payment methods based on subtotal amounts.

### API improvements:

* **`internal/api/v1beta1/billing_checkout.go`:** Enhanced the `CreateCheckout` handler to support product quantity in requests, improving API flexibility for checkout creation.


internal/api/v1beta1/billing_checkout.go
```diff
@@ -117,16 +117,20 @@ func (h Handler) CreateCheckout(ctx context.Context, request *frontierv1beta1.Cr
 		skipTrial = request.GetSubscriptionBody().GetSkipTrial()
 		cancelAfterTrail = request.GetSubscriptionBody().GetCancelAfterTrial()
 	}
-	featureID := ""
+
+	var featureID string
+	var quantity int64
```
whoAbhishekSah, line 122: Since it should always be greater than or equal to 0, how about we use "unint"?

billing/invoice/service.go
```diff
@@ -618,6 +636,17 @@ func (s *Service) CreateInProvider(ctx context.Context, custmr customer.Customer
 			"org_id":     custmr.OrgID,
 			"managed_by": "frontier",
 		},
+		PaymentSettings: &stripe.InvoicePaymentSettingsParams{
+			PaymentMethodTypes: paymentMethodTypes,
+			PaymentMethodOptions: &stripe.InvoicePaymentSettingsPaymentMethodOptionsParams{
+				CustomerBalance: &stripe.InvoicePaymentSettingsPaymentMethodOptionsCustomerBalanceParams{
+					FundingType: stripe.String(string(stripe.InvoicePaymentSettingsPaymentMethodOptionsCustomerBalanceFundingTypeBankTransfer)),
+					BankTransfer: &stripe.InvoicePaymentSettingsPaymentMethodOptionsCustomerBalanceBankTransferParams{
+						Type: stripe.String("us_bank_transfer"),
```
whoAbhishekSah, line 645: nit: let's use const string?

billing/checkout/service.go
```diff
@@ -407,7 +428,7 @@ func (s *Service) Create(ctx context.Context, ch Checkout) (Checkout, error) {
 			AutomaticTax: &stripe.CheckoutSessionAutomaticTaxParams{
 				Enabled: stripe.Bool(s.stripeAutoTax),
 			},
-			Currency: stripe.String(billingCustomer.Currency),
+			Currency: stripe.String(s.defaultCurrency),
```
whoAbhishekSah, line 431: This makes billing customers' currency property obsolete and makes it a Frontier app property. Is it intended? 
If yes, we should consider removing this property from the Billing Customers.

billing/invoice/service.go
whoAbhishekSah: "stripe.CheckoutSessionPaymentMethodOptionsCustomerBalanceBankTransferTypeUSBankTransfer"

billing/config.go
```diff
@@ -68,3 +69,23 @@ type ProductConfig struct {
...
 }
+
+type PaymentMethodConfig struct {
+	// Type is the type of the payment method.
+	// possible values: card, customer_balance
+	Type string `yaml:"type" mapstructure:"type"`
+	// Max is the maximum amount in default currency that should be allowed to be charged to this payment method
+	Max int64 `yaml:"max" mapstructure:"max" default:"0"`
+	// Min is the minimum amount in default currency that should be allowed to be charged to this payment method
+	Min int64 `yaml:"min" mapstructure:"min" default:"0"`
```
whoAbhishekSah, line 80: What is the use of min and max config?

billing/invoice/service.go
rohilsurana: not available in our version of stripe sdk

billing/config.go
rohilsurana: min max will help adding conditions for selecting payment methods on the basis of amount.
For ex - 
If amount is greater than $100,000 then card payment method won't be supported.
or
If amount is less than $40,000 then Bank transfer method won't be allowed.

internal/api/v1beta1/billing_checkout.go
rohilsurana: Currently the API uses int64, hence kept it the same.

billing/checkout/service.go
rohilsurana: Yes, it makes it a frontier property but it only helps in enabling payment method type feature, hence if there is a product with different currency it would just skip that product for calculations.

PR #1061, opened 2025-07-10
feat: add banner for postpaid users

rohilsurana:
This pull request introduces a banner to inform overdraft users, i.e. users with credit_min set to negative, of the behaviour of how adding tokens and paying invoices works.
Screenshot 2025-07-10 at 17 50 44

### DIFF
--- a/sdks/js/packages/core/react/components/organization/tokens/index.tsx
+++ b/sdks/js/packages/core/react/components/organization/tokens/index.tsx
@@ -13,7 +13,6 @@ import { DEFAULT_TOKEN_PRODUCT_NAME } from '~/react/utils/constants';
 import { useBillingPermission } from '~/react/hooks/useBillingPermission';
 import { useTokens } from '~/react/hooks/useTokens';
 import coin from '~/react/assets/coin.svg';
-import billingStyles from '../billing/billing.module.css';
 
 interface TokenHeaderProps {
   billingSupportEmail?: string;
--- a/sdks/js/packages/core/react/components/organization/tokens/index.tsx
+++ b/sdks/js/packages/core/react/components/organization/tokens/index.tsx
@@ -1,4 +1,5 @@
-import { Button, Tooltip, Skeleton, Text, Headline, Flex, Image, toast, Link } from '@raystack/apsara/v1';
+import { Button, Tooltip, Skeleton, Text, Headline, Flex, Image, toast, Link, Callout } from '@raystack/apsara/v1';
+import { InfoCircledIcon } from '@radix-ui/react-icons';
 import { styles } from '../styles';
 import tokenStyles from './token.module.css';
 import { useFrontier } from '~/react/contexts/FrontierContext';
@@ -12,6 +13,7 @@ import { DEFAULT_TOKEN_PRODUCT_NAME } from '~/react/utils/constants';
 import { useBillingPermission } from '~/react/hooks/useBillingPermission';
 import { useTokens } from '~/react/hooks/useTokens';
 import coin from '~/react/assets/coin.svg';
+import billingStyles from '../billing/billing.module.css';
 
 interface TokenHeaderProps {
   billingSupportEmail?: string;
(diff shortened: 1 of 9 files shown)


sdks/js/packages/core/react/components/organization/tokens/index.tsx
```diff
@@ -12,6 +13,7 @@ import { DEFAULT_TOKEN_PRODUCT_NAME } from '~/react/utils/constants';
 import { useBillingPermission } from '~/react/hooks/useBillingPermission';
 import { useTokens } from '~/react/hooks/useTokens';
 import coin from '~/react/assets/coin.svg';
+import billingStyles from '../billing/billing.module.css';
```
rohanchkrabrty: Unused import, can be removed

sdks/js/packages/core/react/components/organization/tokens/index.tsx
```diff
@@ -114,6 +116,29 @@ function BalancePanel({
...
+      <Callout
+        type="accent"
+        icon={<InfoCircledIcon className={tokenStyles.tokenInfoText} />}
+        className={tokenStyles.tokenInfoBox}
+      >You can now add tokens anytime to reduce next month’s invoice. But this won’t settle any existing or overdue invoices.
+      </Callout>) : null}
+    </>
+  );
+}
+
```
rohanchkrabrty: Lets not create a separate component for info callout if it's not used elsewhere

sdks/js/packages/core/react/components/organization/tokens/index.tsx
```diff
@@ -226,6 +251,12 @@ export default function Tokens() {
             billingSupportEmail={config.billing?.supportEmail}
             isLoading={isLoading}
           />
+          <TokenInfoBox
+            balance={tokenBalance}
+            isLoading={isLoading}
+            isCheckoutLoading={isCheckoutLoading}
+            canUpdateWorkspace={isAllowed}
+          />
```
rohanchkrabrty, line 255: `isPostpaid` logic can be defined in this parent component

and the token info can be written as 

```jsx
{isPostpaid && isAllowed ? (
<Callout
  type="accent"
  icon={<InfoCircledIcon className={tokenStyles.tokenInfoText} />}
  className={tokenStyles.tokenInfoBox}
>You can now add tokens anytime to reduce next month’s invoice. But this won’t settle any existing or overdue invoices.
</Callout> : null}
```

rohanchkrabrty (changes requested):

sdks/js/packages/core/react/components/organization/tokens/index.tsx
rohilsurana: The thought was that in future we could have more such callouts for errors etc, hence wanted to keep these segregated.

sdks/js/packages/core/react/components/organization/tokens/index.tsx
rohilsurana: removing

sdks/js/packages/core/react/components/organization/tokens/index.tsx
rohilsurana: Did not get this? Is this related to the previous comment of inlining the component?
If not can you elaborate or suggest changes?

rohanchkrabrty: Yes, related to the previous comment of inlining the component. Missed to mention that.

If we go inline, the code block would be similar to what I mentioned in this comment

sdks/js/packages/core/react/components/organization/tokens/index.tsx
rohanchkrabrty: Understood. In that case please remove the unused props from type like `balance`, `isLoading` etc

And the return statement can be written as
```jsx
return isPostpaid && canUpdateWorkspace ? (
<Callout
  type="accent"
  icon={<InfoCircledIcon className={tokenStyles.tokenInfoText} />}
  className={tokenStyles.tokenInfoBox}
>You can now add tokens anytime to reduce next month’s invoice. But this won’t settle any existing or overdue invoices.
</Callout>) : null );
  ```
  
  `<> </>` will create a React fragment which is not needed in this case

rohilsurana: updated

rohanchkrabrty (approved):

PR #1065, opened 2025-07-14
fix(admin_ui): send negative value for postpaid min credit

rsbh:


ui/src/pages/organizations/details/edit/billing.tsx
```diff
@@ -26,7 +26,10 @@ interface EditBillingPanelProps {
 const billingDetailsUpdateSchema = z
   .object({
     token_payment_type: z.enum(["prepaid", "postpaid"]),
-    credit_min: z.number(),
+    credit_min: z.number().transform((val) => {
+      // Convert negative API values to positive UI values
+      return Math.abs(val);
+    }),
```
paanSinghCoder, line 32: // Convert negative API values to positive UI values
credit_min: z.number().transform((val) => Math.abs(val))

paanSinghCoder (approved):

PR #1075, opened 2025-07-22
feat(sdk): add i18n support

rsbh:
This PR adds i18n support in React SDK using https://github.com/i18next/react-i18next.
The existing app name and plan change messages have been migrated from config to translation data.
This also introduces terminology support, so the user can easily change the names of features in SDK.
eg. user can change `organization` to `Workspace`, `group` to `team`, etc.

Only `organization` terminology has been added in this PR.


rsbh:
Closin

PR #1086, opened 2025-07-30
feat(sdk): setup connect query and migrate authn calls

rsbh:
This PR sets up the connect-rpc client with [connect-query](https://github.com/connectrpc/connect-query-es) (wrapper of react-query).
The method definitions are coming from the `@raystack/proton/frontier` package.
The  `@raystack/proton/frontier` package is TypeScript code generated from the https://github.com/raystack/proton repo.

Following methods are migrated to connect-rpc method instead of fetch calls.
  1. ListAuthStrategies
  2. Authenticate
  3. AuthCallback
  4. GetOrganization
  5. AuthLogout

This PR also added `/hooks` entry path in `@raystack/frontier`, which exports the connect-query hooks and frontier service methods, so the consumer apps don't need to install `@raystack/proton` and `@connectrpc/connect-query` packages and just use the export from `@raystack/frontier/hooks`

The SDK demo app is also updated to use this. A proxy from `/frontier-connect` to the Frontier Connect server is also added in the Vite config.

### DIFF
--- a/sdks/js/packages/sdk-demo/src/pages/Callback.tsx
+++ b/sdks/js/packages/sdk-demo/src/pages/Callback.tsx
@@ -1,37 +1,40 @@
 import { useEffect, useCallback, Suspense } from 'react';
-import frontierClient from '@/api/frontier';
 import { Flex } from '@raystack/apsara/v1';
 import { useSearchParams, useNavigate } from 'react-router-dom';
 import useAuthRedirect from '@/hooks/useAuthRedirect';
+import { useMutation, FrontierServiceQueries } from '@raystack/frontier/hooks';
 
 function CallbackComponent() {
   const [searchParams] = useSearchParams();
   const navigate = useNavigate();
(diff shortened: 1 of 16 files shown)


sdks/js/packages/sdk-demo/src/pages/Callback.tsx
```diff
@@ -1,37 +1,40 @@
...
+import { useMutation, FrontierServiceQueries } from '@raystack/frontier/hooks';
 
 function CallbackComponent() {
   const [searchParams] = useSearchParams();
   const navigate = useNavigate();
 
-  const callFrontierCallback = useCallback(async () => {
-    const state = searchParams?.get('state');
-    const code = searchParams?.get('code');
+  const { mutateAsync: authCallback } = useMutation(
```
rohanchkrabrty: should this be used with useQuery here? since it's getting called on mount

PR #1088, opened 2025-07-31
feat(sdk): add customization provider to update terminology

rsbh:
This PR adds a customization provider in the SDK, allowing the consumer app to update the entity names according to their needs.
eg change `organization` to `workspace`, `team` to `group`, etc.
The customization provider also adds support for the consumer app to pass the app name in the SDK.

2 new hooks are added: `useTerminology` and `useMessages`.
- `useTerminology`: it will be used to display the entity name as per the consumer app. Options are available to change the word to lowercase, uppercase, and capital case. 
The consumer app must provide both singular and plural forms of the entity.
- `useMessages`- it will be used to show custom messages in the UI. like plan change messages, etc.

Only the `organization` entity has been updated to use the terminology.

In the future, we will update the other entities to use the new hook.


sdks/js/packages/core/react/hooks/useMessages.ts
```diff
@@ -0,0 +1,19 @@
+import { useContext } from 'react';
+import { get } from 'lodash';
+import {
+  CustomizationContext,
+  defaultCustomization
+} from '../contexts/CustomizationContext';
+
+export const useMessages = () => {
+  const context = useContext(CustomizationContext);
+  const config = context || defaultCustomization;
```
rohanchkrabrty, line 10: make this into a hook and export it from `CustomizationContext` file

```ts
useCustomizationContext() => {
 const context = useContext(CustomizationContext);
 return context || defaultCustomization;
}
```

sdks/js/packages/core/react/hooks/useTerminology.ts
```diff
@@ -0,0 +1,58 @@
...
+      terminology.project!.singular,
+      terminology.project!.plural
+    ),
+    team: createEntity(terminology.team!.singular, terminology.team!.plural),
+    member: createEntity(
+      terminology.member!.singular,
+      terminology.member!.plural
+    ),
+    user: createEntity(terminology.user!.singular, terminology.user!.plural),
+    appName: createEntity(terminology.appName!, terminology.appName!)
```
rohanchkrabrty: can you convert this to reducer function which will loop over entities

sdks/js/packages/core/shared/types.ts
```diff
@@ -15,8 +15,21 @@ export interface FrontierClientBillingOptions {
   basePlan?: BasePlan;
 }
 
-export interface FrontierClientAPIPlatformOptions {
-  appName?: string;
+export interface FrontierClientCustomizationOptions {
+  terminology?: {
+    organization?: { singular: string; plural: string };
+    project?: { singular: string; plural: string };
+    team?: { singular: string; plural: string };
```
rohanchkrabrty: `{ singular: string; plural: string }`

this can be made into a reusable type

sdks/js/packages/core/react/hooks/useTerminology.ts
```diff
@@ -0,0 +1,58 @@
...
+      return text.toUpperCase();
+    case 'capital':
+      return text.charAt(0).toUpperCase() + text.slice(1).toLowerCase();
+    default:
+      return text;
+  }
+};
+
+const createEntity = (singular: string, plural: string): TerminologyEntity => {
+  return ({ plural: isPlural = false, case: caseType }: TerminologyOptions = {}) => {
```
rohanchkrabrty: can we put a default case for the `caseType` like lower or etc

rsbh: This is to return the entity name as it is

rohanchkrabrty (approved):

PR #1106, opened 2025-08-20
feat: migrate org export RPCs to Connect Server

whoAbhishekSah:


internal/api/v1beta1connect/organization_tokens.go
```diff
@@ -66,3 +66,29 @@ func transformAggregatedTokenToPB(v svc.AggregatedToken) *frontierv1beta1.Search
...
+		chunk := orgTokensDataBytes[i:end]
+		msg := &httpbody.HttpBody{
+			ContentType: contentType,
+			Data:        chunk,
+		}
+
+		if err := stream.Send(msg); err != nil {
+			return connect.NewError(connect.CodeInternal, ErrInternalServerError)
+		}
+	}
```
rsbh: This is common in all the export functions, we can move this to a common helper function

whoAbhishekSah: Done.

rsbh (approved):

PR #1118, opened 2025-09-01
feat: add `audit_records` table with indexes and UUIDv7 generation function

AmanGIT07:


internal/store/postgres/migrations/20250901054744_create_audits_table.down.sql
```diff
@@ -0,0 +1,22 @@
+-- Drop indexes first
+DROP INDEX IF EXISTS idx_audit_records_idempotency_key;
+DROP INDEX IF EXISTS idx_audit_records_target;
+DROP INDEX IF EXISTS idx_audit_records_deleted;
+DROP INDEX IF EXISTS idx_audit_records_resource_metadata_gin;
+DROP INDEX IF EXISTS idx_audit_records_actor_metadata_gin;
+DROP INDEX IF EXISTS idx_audit_records_metadata_gin;
+DROP INDEX IF EXISTS idx_audit_records_request_id;
+DROP INDEX IF EXISTS idx_audit_records_org_resource;
+DROP INDEX IF EXISTS idx_audit_records_org_actor;
```
rohilsurana: lets remove ununsed indexes from here too

internal/store/postgres/migrations/20250901054744_create_audits_table.up.sql
```diff
@@ -0,0 +1,73 @@
...
+
+CREATE TABLE audit_records (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v7(),
+    idempotency_key UUID, -- Same key for bulk requests
+    event VARCHAR(255) NOT NULL,
+    actor_id UUID NOT NULL,
+    actor_type VARCHAR(50) NOT NULL,
+    actor_name VARCHAR(255) NOT NULL,
+    actor_metadata JSONB DEFAULT '{}',
+    resource_id UUID NOT NULL,
```
rohilsurana: lets make this as a varchar field as it might not necessarily be uuid

rohilsurana: same for target_id as well.

Since actor is only defined by frontier itself we can keep it as uuid

internal/store/postgres/migrations/20250901054744_create_audits_table.down.sql
AmanGIT07: Ahh. missed it. removed

internal/store/postgres/migrations/20250901054744_create_audits_table.up.sql
AmanGIT07: updated

rohilsurana (approved):

PR #1122, opened 2025-09-03
feat(Session): Fetch metadata in create session

paanSinghCoder:
This PR creates a util which fetches metadata from req.header and adds it to the DB during create session call.
1. Session Metadata Structure
- IP Address: Client's IP address (extracted from configurable header)
- Location: Country and city information (from CloudFront headers)
- Operating System: Detected from User-Agent string
- Browser: Detected from User-Agent string
2. Metadata Extraction Utility (pkg/utils/metadata.go)
- ExtractSessionMetadata() function that parses HTTP headers
- Configurable header names for IP, country, and city
- User-Agent parsing for OS and browser detection
- Support for major browsers (Chrome, Safari, Firefox, Edge, etc.)
- Support for major operating systems (Windows, macOS, Linux, iOS, Android, etc.)
Cloudfront docs: [https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-device-type](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-device-type)


pkg/server/config.go
```diff
@@ -34,9 +34,17 @@ type UIConfig struct {
 	OrganizationTypes []string `yaml:"organization_types" mapstructure:"organization_types"`
 }
 
+type ConnectHeader struct {
```
rohilsurana: should we move these to session configurations? As these do not affect the connect protocol/server in any way.

pkg/server/config.go
```diff
@@ -34,9 +34,17 @@ type UIConfig struct {
...
+	ViewerAddress string `yaml:"viewer_address" mapstructure:"viewer_address" default:"CloudFront-Viewer-Address"`
+	ViewerCountry string `yaml:"viewer_country" mapstructure:"viewer_country" default:"CloudFront-Viewer-Country"`
+	ViewerCity    string `yaml:"viewer_city" mapstructure:"viewer_city" default:"CloudFront-Viewer-City"`
```
paanSinghCoder: Rename default to generic values instead of Cloudfront values. X-FRONTIER-IP, X-FRONTIER-COUNTRY, X-FRONTIER-CITY (in small).

paanSinghCoder: Rename `viewer_` to `client_`

pkg/server/connect_interceptors/metadata.go
```diff
@@ -0,0 +1,108 @@
...
+	return next
+}
+
+func (s *SessionMetadataInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+	return next
+}
+
+func (s *SessionMetadataInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
+	return connect.UnaryFunc(func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		metadata := make(map[string]any)
```
paanSinghCoder: Make metadata a struct.

pkg/server/connect_interceptors/metadata.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		ctx = consts.WithSessionMetadata(ctx, metadata)
+		return next(ctx, req)
+	})
+}
+
+func extractBrowser(userAgent string) string {
+	userAgent = strings.ToLower(userAgent)
+
```
paanSinghCoder: Make these functions support more OS/browsers.

internal/api/v1beta1connect/authenticate.go
```diff
@@ -112,8 +113,11 @@ func (h *ConnectHandler) AuthCallback(ctx context.Context, request *connect.Requ
 		return nil, connect.NewError(connect.CodeInternal, err)
 	}
 
+	// Extract session metadata from request headers
+	metadata := utils.ExtractSessionMetadata(ctx, request, h.metadataConfig)
```
AmanGIT07: `metadata` is an imported package. Will cause issues when we try to reference it later. we should rename this variable name

core/authenticate/config.go
```diff
@@ -54,6 +54,14 @@ type SessionConfig struct {
 	// Validity is the duration for which the session is valid
 	Validity time.Duration `mapstructure:"validity" yaml:"validity" default:"720h"`
 	Secure   bool          `mapstructure:"secure" yaml:"secure" default:"false"`
+	// Headers configuration for session metadata collection
+	Headers ConnectHeader `yaml:"headers" mapstructure:"headers"`
+}
+
+type ConnectHeader struct {
```
AmanGIT07: Can we rename it to something related to session like SessionMetadataHeaders?

internal/api/v1beta1connect/authenticate.go
```diff
@@ -112,8 +113,11 @@ func (h *ConnectHandler) AuthCallback(ctx context.Context, request *connect.Requ
...
+
 	// registration/login complete, build a session
-	session, err := h.sessionService.Create(ctx, response.User.ID)
+	session, err := h.sessionService.Create(ctx, response.User.ID, metadata)
```
AmanGIT07: Instead of accepting metadata as a param, can we not get it from context or header directly in the service?

internal/api/v1beta1connect/v1beta1connect.go
```diff
@@ -17,6 +18,7 @@ type ConnectHandler struct {
 	frontierv1beta1connect.UnimplementedFrontierServiceHandler
 
 	authConfig                       authenticate.Config
+	metadataConfig                   utils.MetadataConfig
```
AmanGIT07: why do we need this in the handler? Can we pass it to the session service during initialization?

pkg/utils/metadata.go
```diff
@@ -0,0 +1,137 @@
...
+	"connectrpc.com/connect"
+	"github.com/raystack/frontier/core/authenticate/session"
+)
+
+// MetadataConfig holds configuration for header names
+type MetadataConfig struct {
+	ClientIP      string
+	ClientCountry string
+	ClientCity    string
+}
```
AmanGIT07: can name it better like SessionMetadataConfig

pkg/utils/metadata.go
```diff
@@ -0,0 +1,137 @@
...
+	if city := req.Header().Get(config.ClientCity); city != "" {
+		metadata.Location.City = city
+	}
+
+	// OS and Browser (from User-Agent)
+	userAgent := req.Header().Get("User-Agent")
+	if userAgent != "" {
+		metadata.OS = extractOS(userAgent)
+		metadata.Browser = extractBrowser(userAgent)
+	}
```
AmanGIT07: can we take a look at this library to extract these data cleanly - https://github.com/mssola/useragent ?

core/authenticate/session/service.go
```diff
@@ -60,9 +61,14 @@ func (s Service) Create(ctx context.Context, userID string) (*Session, error) {
...
-		Metadata:        nil,
+		Metadata:        metadata,
+	}
+	err := s.repo.Set(ctx, sess)
+	if err != nil {
+		s.log.Warn("failed to create session", "err", err)
+		return nil, err
 	}
-	return sess, s.repo.Set(ctx, sess)
+	return sess, nil
```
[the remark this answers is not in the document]
paanSinghCoder: it was concise but there was no error handling present.

pkg/utils/metadata.go
paanSinghCoder: Should we install an external dependency just for this? In frontend, we usually avoid external packages unless very necessary. Do you see any added advantage to install an external lib?

internal/api/v1beta1connect/authenticate.go
paanSinghCoder: I think it would tightly couple the HTTP specific logic and make it difficult to test.

PR #1137, opened 2025-09-18
feat: Implement export audit records rpc with streaming

AmanGIT07:
This PR implements the ExportAuditRecords RPC for streaming audit record data as CSV files.

  Implementation Details

  - Repository Layer: Uses PostgreSQL server-side cursors to handle large datasets without loading all records into memory. Fetches data in configurable batches (1000 records) and streams via io.Pipe to the handler.
  - Handler Layer: Receives the data stream from repository and forwards it to the client using Connect's server-side streaming with chunked HTTP responses (200KB chunks).
  - RQL Support: Supports filtering, searching, and sorting through RQL query parameters, consistent with the existing ListAuditRecords endpoint.
  - Memory Efficiency: The cursor-based approach ensures constant memory usage regardless of result set size, making it suitable for exporting large number of records

### DIFF
--- a/internal/store/postgres/audit_record_repository.go
+++ b/internal/store/postgres/audit_record_repository.go
@@ -257,7 +257,6 @@ func (r AuditRecordRepository) Export(ctx context.Context, rqlQuery *rql.Query) 
 		goqu.L(`COALESCE(metadata::text, '{}') AS "Metadata"`),
 	)
 
-	// Apply sorting after selecting columns for export
 	csvQuery, err = utils.AddRQLSortInQuery(csvQuery, rqlQuery)
 	if err != nil {
 		return nil, "", wrapValidationError(err)
@@ -288,7 +287,7 @@ func (r AuditRecordRepository) executeCursorQuery(ctx context.Context, selectQue
 			if err != nil {
 				return fmt.Errorf("failed to begin transaction: %w", err)
 			}
-			defer tx.Rollback() // Always rollback
+			defer tx.Rollback()
 
 			// Generate unique cursor name
 			cursorName, err := r.generateCursorName()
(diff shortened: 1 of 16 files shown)


internal/store/postgres/audit_record_repository.go
```diff
@@ -230,6 +228,114 @@ func (r AuditRecordRepository) List(ctx context.Context, rqlQuery *rql.Query) (a
...
+		goqu.L(`COALESCE(target_name, '')`),
+		goqu.L(`COALESCE(target_metadata::text, '{}')`),
+		goqu.L(`COALESCE(org_id::text, '')`),
+		goqu.L(`COALESCE(request_id, '')`),
+		goqu.L(`to_char(occurred_at AT TIME ZONE 'UTC', 'YYYY-MM-DD HH24:MI:SS.MS TZ')`),
+		goqu.L(`to_char(created_at AT TIME ZONE 'UTC', 'YYYY-MM-DD HH24:MI:SS.MS TZ')`),
+		goqu.L(`COALESCE(metadata::text, '{}')`),
+	)
+
+	csvQuery, err = utils.AddRQLSortInQuery(csvQuery, rqlQuery)
```
whoAbhishekSah, line 260: Why not add sort in the `buildFilteredQuery` function itself?

whoAbhishekSah (approved):

internal/store/postgres/audit_record_repository.go
AmanGIT07: We are using `buildFilteredQuery` in List method as well. List supports grouping which in turn uses ordering under the hood.  It was interfering and causing issues - column "x" must appear in the GROUP BY clause or be used in an aggregate function (SQLSTATE 42803).

PR #1167, opened 2025-09-30
feat: enrich context with audit required info

AmanGIT07:


core/auditrecord/service.go
```diff
@@ -164,3 +166,42 @@ func computeHash(auditRecord AuditRecord) string {
 	// Convert to hex string
 	return hex.EncodeToString(hashBytes)
 }
+
+type auditContextToSet struct {
+	Actor           Actor
+	IsSuperUser     bool
+	SessionMetadata frontiersession.SessionMetadata
+}
```
rohilsurana: These should not be part of audit context, instead they should be their own , like session context to set, authcontext to set etc, audit service is just a consumer for it.

The namespacing should be dependent on Producer of the information, as usually there is only one producer but can have multiple consumers.

pkg/server/consts/gateway.go
```diff
@@ -13,10 +13,16 @@ func (c *contextKey) String() string { return "context value " + c.name }
...
 
-	AuditActorContextKey    = contextKey{name: "audit-actor"}
-	AuditMetadataContextKey = contextKey{name: "audit-metadata"}
-	AuditServiceContextKey  = contextKey{name: "audit-service"}
+	// SessionContextKey is context key that contains session metadata
+	SessionContextKey = contextKey{name: "session-context"}
+
+	AuditActorContextKey       = contextKey{name: "audit-actor"}
+	AuditRecordActorContextKey = contextKey{name: "audit-record-actor"}
+	AuditMetadataContextKey    = contextKey{name: "audit-metadata"}
```
rohilsurana: can we get rid of this?

pkg/server/consts/gateway.go
```diff
@@ -13,10 +13,16 @@ func (c *contextKey) String() string { return "context value " + c.name }
...
+	AuthSuperUserContextKey = contextKey{name: "auth-superuser"}
 
-	AuditActorContextKey    = contextKey{name: "audit-actor"}
-	AuditMetadataContextKey = contextKey{name: "audit-metadata"}
-	AuditServiceContextKey  = contextKey{name: "audit-service"}
+	// SessionContextKey is context key that contains session metadata
+	SessionContextKey = contextKey{name: "session-context"}
+
+	AuditActorContextKey       = contextKey{name: "audit-actor"}
+	AuditRecordActorContextKey = contextKey{name: "audit-record-actor"}
```
rohilsurana, line 23: lets move AuditRecord* apart from Audit* ones.

rohilsurana (approved):

pkg/server/consts/gateway.go
AmanGIT07: done

core/auditrecord/service.go
AmanGIT07: done

pkg/server/consts/gateway.go
AmanGIT07: will remove existing log after new implementation is complete

PR #1177, opened 2025-10-07
feat: migrate relations, permissions and namespaces RPC to ConnectRPC

whoAbhishekSah:
## Migrate Relations, Permissions, and Namespaces RPCs to ConnectRPC server

The following RPCs are migrated. 

- [x] CreateRelation
- [x] GetRelation 
- [x] DeleteRelation
- [x] ListPermissions
- [x] GetPermission
- [x] CheckResourcePermission
- [x] BatchCheckPermission  
- [x] ListNamespaces
- [x] GetNamespace

The implementation of the handlers is similar to that of the gRPC handlers. 
I have just made it compatible with the ConnectRPC signature. 
We have used connect errors instead of gRPC errors.

**One crucial change** made is that the following APIs have been put behind an Authentication check, which was not implemented in the GPRC handler. 

- [x] ListPermissions
- [x] GetPermission
- [x] ListNamespaces
- [x] GetNamespace

The reason is that these APIs store Frontier resource-level permission names. They should at least be checked for authentication.

### DIFF
--- a/internal/api/v1beta1connect/permission_check.go
+++ b/internal/api/v1beta1connect/permission_check.go
@@ -124,7 +124,7 @@ func (h *ConnectHandler) CheckResourcePermission(ctx context.Context, req *conne
 
 	permissionName, err := h.getPermissionName(ctx, objectNamespace, req.Msg.GetPermission())
 	if err != nil {
-		return nil, connect.NewError(connect.CodeInvalidArgument, ErrInternalServerError)
+		return nil, err
 	}
 	result, err := h.resourceService.CheckAuthz(ctx, resource.Check{
 		Object: relation.Object{
@@ -154,7 +154,7 @@ func (h *ConnectHandler) BatchCheckPermission(ctx context.Context, req *connect.
 
 		permissionName, err := h.getPermissionName(ctx, objectNamespace, body.GetPermission())
 		if err != nil {
-			return nil, connect.NewError(connect.CodeInvalidArgument, ErrInternalServerError)
+			return nil, err
 		}
 		checks = append(checks, resource.Check{
 			Object: relation.Object{
--- a/internal/api/v1beta1connect/permission_check.go
+++ b/internal/api/v1beta1connect/permission_check.go
@@ -111,3 +111,75 @@ func (h *ConnectHandler) fetchAccessPairsOnResource(ctx context.Context, objectN
 		return pair.Status
 	}), nil
 }
+
+func (h *ConnectHandler) CheckResourcePermission(ctx context.Context, req *connect.Request[frontierv1beta1.CheckResourcePermissionRequest]) (*connect.Response[frontierv1beta1.CheckResourcePermissionResponse], error) {
+	objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetResource())
+	if len(req.Msg.GetResource()) == 0 || err != nil {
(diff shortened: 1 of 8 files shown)


rohilsurana (commented):
## Error Code Inconsistency Issue

Found an error handling inconsistency in `internal/api/v1beta1connect/permission_check.go` at line 127:

```go
permissionName, err := h.getPermissionName(ctx, objectNamespace, req.Msg.GetPermission())
if err != nil {
    return nil, connect.NewError(connect.CodeInvalidArgument, ErrInternalServerError)
}
```

**Issue**: When `getPermissionName` returns an error, it's wrapped as `InvalidArgument` but uses the error constant `ErrInternalServerError`. This is semantically incorrect since `getPermissionName` already returns properly formatted Connect errors.

**Fix**: Should be:
```go
if err != nil {
    return nil, err // getPermissionName already returns properly formatted Connect errors
}
```

The `getPermissionName` function (lines 30-49) already handles error mapping correctly:
- `permission.ErrNotExist` → `connect.CodeNotFound`
- Other errors → `connect.CodeInternal`

So we should propagate its errors directly rather than re-wrapping them incorrectly.

internal/api/v1beta1connect/permission_check.go
```diff
@@ -111,3 +111,75 @@ func (h *ConnectHandler) fetchAccessPairsOnResource(ctx context.Context, objectN
...
+		objectNamespace = schema.ParseNamespaceAliasIfRequired(req.Msg.GetObjectNamespace())
+		objectID = req.Msg.GetObjectId()
+	}
+	if objectNamespace == "" || objectID == "" {
+		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+	}
+
+	permissionName, err := h.getPermissionName(ctx, objectNamespace, req.Msg.GetPermission())
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, ErrInternalServerError)
```
rohilsurana: **Issue**: When `getPermissionName` returns an error, it is wrapped as `InvalidArgument` but uses the error constant `ErrInternalServerError`. This is semantically incorrect since `getPermissionName` already returns properly formatted Connect errors.

**Fix**: Should be:
```go
if err != nil {
    return nil, err // getPermissionName already returns properly formatted Connect errors
}
```

The `getPermissionName` function already handles error mapping correctly, so we should propagate its errors directly.

rohilsurana (changes requested):
Found error handling inconsistency that needs to be fixed.

internal/api/v1beta1connect/permission_check.go
```diff
@@ -111,3 +111,75 @@ func (h *ConnectHandler) fetchAccessPairsOnResource(ctx context.Context, objectN
...
+	checks := make([]resource.Check, 0, len(req.Msg.GetBodies()))
+	for _, body := range req.Msg.GetBodies() {
+		objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(body.GetResource())
+		if len(body.GetResource()) == 0 || err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		}
+
+		permissionName, err := h.getPermissionName(ctx, objectNamespace, body.GetPermission())
+		if err != nil {
+			return nil, connect.NewError(connect.CodeInvalidArgument, ErrInternalServerError)
```
rohilsurana: Same here as well

rohilsurana (approved):

PR #1178, opened 2025-10-07
feat: add audit logging for KYC upsert operations

AmanGIT07:


internal/store/postgres/kyc_repository.go
```diff
@@ -76,25 +77,53 @@ func (r OrgKycRepository) Upsert(ctx context.Context, input kyc.KYC) (kyc.KYC, e
...
 	if err == nil {
-		//kyc for org exists, prepare UPDATE query
-		query, params, err = dialect.Update(TABLE_ORGANIZATIONS_KYC).Set(goqu.Record{
-			"status": input.Status,
-			"link":   input.Link,
-		}).Where(goqu.Ex{"org_id": input.OrgID}).
-			Returning(&kycModel).ToSQL()
+		// kyc for org exists, prepare UPDATE query with JOIN
+		query, params, err = dialect.Update(TABLE_ORGANIZATIONS_KYC).
+			From(TABLE_ORGANIZATIONS).
```
rohilsurana: (nit) lets chain From after the Set method, just to stay similar to how we usually write in SQL.

rohilsurana (approved):

internal/store/postgres/kyc_repository.go
AmanGIT07: done

PR #1189, opened 2025-10-09
feat: add audit records for multiple events

AmanGIT07:


internal/store/postgres/billing_checkout_repository.go
```diff
@@ -154,19 +160,89 @@ func (r BillingCheckoutRepository) Create(ctx context.Context, toCreate checkout
...
+				planID = *result.PlanID
+			}
+			featureID := ""
+			if result.FeatureID != nil {
+				featureID = *result.FeatureID
+			}
+
+			auditRecord := BuildAuditRecord(
+				ctx,
+				"billing_checkout.created",
```
rohilsurana: lets make these constants in one single file, so all audit events can be tracked from one place.

internal/store/postgres/audit_record.go
```diff
@@ -214,7 +214,14 @@ func BuildAuditRecord(ctx context.Context, event string, resource AuditResource,
 		actorMetadata[consts.AuditSessionMetadataKey] = sessionMetadata
 	}
 
-	actorUUID, _ := uuid.Parse(actorID)
+	var actorUUID uuid.UUID
+	if actorID == "" { // cron jobs
+		actorUUID = uuid.Nil
+		actorType = "system"
+		actorName = "system"
```
whoAbhishekSah: nit: let: put them as "Consts" after import statements.

internal/store/postgres/audit_record.go
```diff
@@ -214,7 +214,14 @@ func BuildAuditRecord(ctx context.Context, event string, resource AuditResource,
...
+	} else {
+		actorUUID, _ = uuid.Parse(actorID)
```
whoAbhishekSah, line 224: Is there a missed error handling here?

internal/store/postgres/billing_checkout_repository.go
```diff
@@ -238,29 +314,103 @@ func (r BillingCheckoutRepository) UpdateByID(ctx context.Context, toUpdate chec
...
+						"customer_id":        result.CustomerID,
+						"checkout_url":       result.CheckoutUrl,
+						"skip_trial":         result.SubscriptionConfig.SkipTrial,
+						"cancel_after_trial": result.SubscriptionConfig.CancelAfterTrial,
+					},
+				},
+				orgID,
+				nil,
+				result.UpdatedAt,
+			)
```
whoAbhishekSah: Let's move the object creation into a separate function so that it reads better -- A private function. This applies to other methods as well.

internal/store/postgres/billing_customer_repository.go
```diff
@@ -391,23 +521,71 @@ func (r BillingCustomerRepository) UpdateDetailsByID(ctx context.Context, custom
...
 
 func (r BillingCustomerRepository) Delete(ctx context.Context, id string) error {
-	stmt := dialect.Delete(TABLE_BILLING_CUSTOMERS).Where(goqu.Ex{
-		"id": id,
-	})
-	query, params, err := stmt.ToSQL()
+	type customerWithContext struct {
+		Customer
+		OrgName string `db:"org_name"`
+	}
```
whoAbhishekSah: Let's move this struct outside the function scope into package scope so that we don't need to redeclare each time?

internal/store/postgres/billing_checkout_repository.go
```diff
@@ -238,29 +314,103 @@ func (r BillingCheckoutRepository) UpdateByID(ctx context.Context, toUpdate chec
 	if toUpdate.PaymentStatus != "" {
 		updateRecord["payment_status"] = toUpdate.PaymentStatus
 	}
+
+	var result checkoutWithContext
+
+	// Fetch org name via billing_customers -> organizations
+	orgNameSubquery := dialect.From(TABLE_BILLING_CUSTOMERS).
```
whoAbhishekSah: Why don't we pass orgName (or Org Object) to this function from the service function itself by calling the Org Service Get: [ref](https://github.com/raystack/frontier/blob/main/core/organization/service.go#L87). 

This can simplify the repository layer code.

internal/store/postgres/audit_record.go
AmanGIT07: it will default to zero UUID

internal/store/postgres/audit_record.go
AmanGIT07: Done

PR #1196, opened 2025-10-11
feat: migrate preferences list to ConnectRPC

rohilsurana:
This pull request updates the preferences list UI to use the new `@raystack/proton/frontier` API and modernizes data fetching and mutation logic. It replaces legacy types and API calls with the latest ConnectRPC-based queries and mutations, improves cache updating after preference changes, and refactors components for better reliability and maintainability.

- API migration and type updates
- Data fetching and error handling improvements
- Preference update logic to connect and cache management
- UI and prop naming consistency

### DIFF
--- a/ui/src/containers/preferences.list/details.tsx
+++ b/ui/src/containers/preferences.list/details.tsx
@@ -82,7 +82,7 @@ export default function PreferenceDetails() {
           cardinality: "finite",
         });
 
-        queryClient.setQueryData(queryKey, (oldData: ListPreferencesResponse | undefined) => {
+        queryClient.setQueryData(queryKey, (oldData?: ListPreferencesResponse) => {
           if (!oldData || !data.preference || !Array.isArray(data.preference)) return oldData;
 
           const updatedPreferences = data.preference;


ui/src/containers/preferences.list/details.tsx
```diff
@@ -60,11 +63,56 @@ function PreferenceValue({ value, trait, onChange }: PreferenceValueProps) {
...
+      onSuccess: (data) => {
+        // Update the cache with the new preference from mutation response
+        const queryKey = createConnectQueryKey({
+          schema: AdminServiceQueries.listPreferences,
+          transport,
+          input: {},
+          cardinality: "finite",
+        });
+
+        queryClient.setQueryData(queryKey, (oldData: ListPreferencesResponse | undefined) => {
```
paanSinghCoder: Any reason we have written it like `oldData: ListPreferencesResponse | undefined` and not this `oldData?: ListPreferencesResponse`?

rohilsurana: oldData can be undefined in case it gets cleared due to gc or other reasons.

paanSinghCoder: Both practically means the same thing. Normally, the suggested one is used instead of setting the type as undefined.

rohilsurana: let me try using it that way

rohilsurana: fixed

PR #1201, opened 2025-10-14
feat: migrate product list and prices to ConnectRPC

rohilsurana:
Migrated Admin Product List and Product Prices pages from REST API to ConnectRPC.

Changes:
- Replaced manual state management with `useQuery` from ConnectRPC
- Updated types from `@raystack/frontier` to `@raystack/proton/frontier`
- Fixed timestamp handling using `timestampToDate` utility
- Added error handling with EmptyState components
- Changed field names from snake_case to camelCase
- Fixed bigint amount conversion to string for Price component
- Updated product icon size to 24x24 px
- Added `onRowClick` handler to open product details sidebar

### DIFF
--- a/ui/src/containers/products.list/columns.tsx
+++ b/ui/src/containers/products.list/columns.tsx
@@ -38,7 +38,7 @@ export const getColumns: () => DataTableColumnDef<
 
         const priceComp =
           prices?.length === 1 ? (
-            <Price value={prices[0].amount?.toString()} currency={prices[0].currency} />
+            <Price value={prices[0].amount?.toString() ?? '0'} currency={prices[0].currency} />
           ) : (
             <NavLink
               to={`/products/${row?.original?.id}/prices`}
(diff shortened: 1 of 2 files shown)


ui/src/containers/products.list/prices/columns.tsx
```diff
@@ -28,46 +29,54 @@ export const getColumns: (
...
       filterVariant: "text",
     },
 
     {
       header: "Amount",
       accessorKey: "amount",
-      cell: ({ row, getValue }) => (
-        <Price value={row.original.amount} currency={row.original.currency} />
+      cell: ({ row }) => (
+        <Price value={row.original.amount?.toString()} currency={row.original.currency} />
```
paanSinghCoder: We can pass a fallback value as we are using the null check else it will return undefined to the value prop.

ui/src/containers/products.list/columns.tsx
```diff
@@ -38,9 +38,12 @@ export const getColumns: () => DataTableColumnDef<
 
         const priceComp =
           prices?.length === 1 ? (
-            <Price value={prices[0].amount} currency={prices[0].currency} />
+            <Price value={prices[0].amount?.toString()} currency={prices[0].currency} />
```
paanSinghCoder: Same here

PR #1207, opened 2025-10-15
feat: Create audit records for invite and role operations

AmanGIT07:


core/auditrecord/models/models.go
```diff
@@ -0,0 +1,70 @@
+package models
```
AmanGIT07, line 1: Moved the types from core/auditrecord/auditrecord.go to this new package to avoid circular dependency:
  - core/auditrecord (service + types) → user, serviceuser, session
  - core/user → core/role
  - core/role → core/auditrecord

core/invitation/service.go
```diff
@@ -359,6 +368,35 @@ func (s Service) Accept(ctx context.Context, id uuid.UUID) error {
...
+	if err != nil {
+		return err
+	}
+
+	// create audit record for invitation acceptance
+	_, err = s.auditRecordRepository.Create(ctx, models.AuditRecord{
+		Event: pkgAuditRecord.OrganizationInvitationAcceptedEvent.String(),
+		Resource: models.Resource{
+			ID:   org.ID,
+			Type: pkgAuditRecord.OrganizationType.String(),
```
rohilsurana: Lets change the type of Type to a proper type instead so we don't have to do .String() everytime, else we loose the benefit of creating a type.

rohilsurana: same for Event as well

AmanGIT07: Done

core/auditrecord/auditrecord.go
```diff
@@ -1,70 +1,12 @@
...
+import "github.com/raystack/frontier/core/auditrecord/models"
+
+// Models moved to a new package to avoid circular dependency with other packages.
+// Re-assigned for backward compatibility
+type AuditRecord = models.AuditRecord
+type Actor = models.Actor
+type Resource = models.Resource
+type Target = models.Target
+type AuditRecordsList = models.AuditRecordsList
+type AuditRecordRQLSchema = models.AuditRecordRQLSchema
```
whoAbhishekSah, line 12: Why don't we simply import the types from models wherever these types are used?

core/organization/service.go
```diff
@@ -61,25 +64,58 @@ type PreferencesService interface {
...
+
+// mapPrincipalTypeToAuditType maps schema principal types to audit record type constants
+func mapPrincipalTypeToAuditType(principalType string) pkgAuditRecord.EntityType {
+	switch principalType {
+	case schema.UserPrincipal:
+		return pkgAuditRecord.UserType
+	case schema.ServiceUserPrincipal:
+		return pkgAuditRecord.ServiceUserType
+	default:
+		return pkgAuditRecord.UserType
```
whoAbhishekSah: can be simplified to 
```
	case schema.ServiceUserPrincipal:
		return pkgAuditRecord.ServiceUserType
	default:
         return pkgAuditRecord.UserType
```

internal/api/v1beta1connect/role.go
```diff
@@ -197,6 +197,13 @@ func (h *ConnectHandler) CreateOrganizationRole(ctx context.Context, request *co
 		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadBodyMetaSchemaError)
 	}
 
+	// Fetch organization to get name for audit record
+	org, err := h.orgService.Get(ctx, request.Msg.GetOrgId())
+	if err != nil {
+		return nil, connect.NewError(connect.CodeInternal, err)
+	}
+	metaDataMap["org_name"] = org.Title
```
whoAbhishekSah: Let's hardcode this metadatamap key "org_name" as a const as it is being used at multiple places.

internal/store/postgres/audit_record.go
```diff
@@ -185,54 +185,68 @@ func extractSuperUserFromContext(ctx context.Context) bool {
...
-	var actorUUID uuid.UUID
-	if actorID == "" { // cron jobs
-		actorUUID = uuid.Nil
-		actorType = pkgAuditRecord.SystemActor
-		actorName = pkgAuditRecord.SystemActor
-	} else {
-		actorUUID, _ = uuid.Parse(actorID)
-	}
+	actorUUID, _ := uuid.Parse(actor.ID)
 	orgUUID, _ := uuid.Parse(orgID)
```
whoAbhishekSah, line 240: It seems like we are missing error handling here.

core/auditrecord/auditrecord.go
AmanGIT07: These are used at lot of places, this PR would have become large. Will refactor in a separate one

internal/store/postgres/audit_record.go
AmanGIT07: It will default to zero uuid

PR #1208, opened 2025-10-16
refactor: replace hardcoded audit entity types with constants

AmanGIT07:


internal/store/postgres/billing_checkout_repository.go
```diff
@@ -180,12 +180,12 @@ func (r BillingCheckoutRepository) Create(ctx context.Context, toCreate checkout
 				auditrecord.BillingCheckoutCreatedEvent.String(),
 				AuditResource{
 					ID:   checkoutModel.CustomerID,
-					Type: "billing_customer",
+					Type: auditrecord.BillingCustomerType.String(),
```
rohilsurana, line 183: lets change the type of Type to Type? 
Also propbably a good reason to call type Type something else like ResourceType to avoid too many tongue twisters.

AmanGIT07: Wont that need us to store all possible events and entities that we expect from other services if we want type safety?

AmanGIT07: Used type, changes made in https://github.com/raystack/frontier/pull/1207

PR #1209, opened 2025-10-16
fix: add nil checks for audit repository

AmanGIT07:


core/audit/service.go
```diff
@@ -78,9 +79,15 @@ func (s *Service) Create(ctx context.Context, l *Log) error {
...
 		l.ID = uuid.NewString()
 	}
-	err := s.repository.Create(ctx, l)
-	if err != nil {
-		return err
+	if l.CreatedAt.IsZero() {
+		l.CreatedAt = time.Now()
+	}
+
+	if s.repository != nil {
```
rohilsurana, line 86: should we handle this here or in the function that initializes it, in this case `NewService` ?

PR #1214, opened 2025-10-17
migrate(sdk): user preferences to ConnectRPC

rohilsurana:
## Summary
Migrate `usePreferences` hook from REST client to Connect RPC

### Fetching Behavior

**Before (REST client):**
- Used `useEffect` with `fetchPreferences` in dependency array - could cause infinite loops if not memoized correctly
- Manual state management with `useState` for preferences
- `fetchPreferences()` returned `Promise<V1Beta1Preference[] | undefined>` - required await
- No caching - every component mount triggered a new network request
- Errors were caught and logged, but returned empty array `[]` (no error state exposed)

**After (Connect RPC):**
- Uses React Query's `useQuery` with `enabled: autoFetch` - prevents infinite loops
- Automatic caching - subsequent mounts use cached data (no redundant requests)
- `fetchPreferences()` returns `void` - triggers refetch via React Query
- Cache is shared across all components using this hook
- React Query provides built-in error state (available via `error` if exposed)

### Update Behavior

**Before (REST client):**
- After successful update, directly updated local state with `setPreferences(getFormattedData(data))`
- No automatic refetch - components using the same hook elsewhere wouldn't see the update
- Returned updated preferences array `Promise<V1Beta1Preference[] | undefined>`
- On error, logged to console and returned empty array `[]` (silently failed)

**After (Connect RPC):**
- After successful update, invalidates React Query cache for `listCurrentUserPreferences`
- Automatic background refetch triggers - **all components** using this hook see the update
- Returns `Promise<void>` - doesn't return the data (fetched automatically)
- On error, logs to console and **re-throws error** for caller to handle
- Runtime type validation added via `create(CreateCurrentUserPreferencesRequestSchema, { bodies })`

### Caching & Synchronization

**Before:**
- ❌ No caching - same preferences fetched multiple times
- ❌ No synchronization - updating in one component didn't update others
- ❌ Manual state management across components

**After:**  
- ✅ Automatic caching via React Query
- ✅ Global synchronization - updates propagate to all components
- ✅ Shared cache reduces network requests

### Status Management

**Before:**
```typescript
status: 'idle' | 'fetching' | 'loading'  // manually managed with setState
isLoading: status === 'loading'          // derived from status
isFetching: status === 'fetching'        // derived from status
```

**After:**
```typescript
isFetching: from useQuery                // tracks fetch operations
isLoading: from useMutation             // tracks update operations  
status: computed from both              // 'loading' | 'fetching' | 'idle'
```

### Key Improvements
1. **Eliminates race conditions**: React Query handles concurrent requests
2. **Better performance**: Caching reduces unnecessary network calls
3. **Global state sync**: Updates propagate automatically
4. **Type safety**: Runtime validation with protobuf schemas
5. **Cleaner code**: No manual state/effect management

### DIFF
--- a/sdks/js/packages/core/react/hooks/usePreferences.ts
+++ b/sdks/js/packages/core/react/hooks/usePreferences.ts
@@ -73,7 +73,7 @@ export function usePreferences({
     }
   });
 
-  const updatePreferences = async (preferences: Preference[]) => {
+  const updatePreferences = useCallback(async (preferences: Preference[]) => {
     try {
       const req = create(CreateCurrentUserPreferencesRequestSchema, {
         bodies: preferences
@@ -86,7 +86,7 @@ export function usePreferences({
       console.error(err);
       throw err;
     }
-  };
+  }, [updatePreferencesMutation]);
 
   const status: UsePreferences['status'] = isUpdatingPreferences
     ? 'loading'


sdks/js/packages/core/react/hooks/usePreferences.ts
```diff
@@ -27,64 +35,76 @@ export function usePreferences({
...
+  } = useQuery(
+    FrontierServiceQueries.listCurrentUserPreferences,
+    {},
+    { enabled: autoFetch }
+  );
+
+  const preferences = useMemo(
+    () => getFormattedData(data?.preferences ?? []),
+    [data?.preferences]
+  );
```
rohanchkrabrty: `useQuery` has an option `select` which is a function that can transform the returned data. Let's use that to format the data instead of using a `useMemo`

https://tanstack.com/query/latest/docs/framework/react/reference/useQuery

sdks/js/packages/core/react/hooks/usePreferences.ts
```diff
@@ -27,64 +35,76 @@ export function usePreferences({
...
-      } finally {
-        setStatus('idle');
-      }
-      return [];
-    },
-    [client]
-  );
+  const fetchPreferences = () => {
+    refetch();
+  };
```
rohanchkrabrty: wrap both the functions in `useCallback`

rohilsurana: We don't need to wrap refetch at all, have remove this since refetch from React Query already has a stable reference so it doesn't change between renders.

rohanchkrabrty: Yeah `refetch` was overkill. But the `updatePreferences` needs to be wrapped. On each render a new function instance will be created. `useCallback` will prevent that

rohilsurana: ahh missed that, only looked at this one. making the change now.

PR #1217, opened 2025-10-23
feat: add `org_name` to audit records

AmanGIT07:
Added organization name (org_name) to audit records. Refactored billing transaction audit logging to avoid serialization conflicts in high-concurrency scenarios.

 Why:
  - org_name field: Eliminates need to JOIN with organizations table for display/filtering, improving query performance and user experience
  - Refactoring: SERIALIZABLE transactions were hitting serialization conflicts when fetching org_name from organizations table under high concurrency. Moving audit creation to service layer (outside transaction) solves this.

### DIFF
--- a/internal/store/postgres/audit_record_repository.go
+++ b/internal/store/postgres/audit_record_repository.go
@@ -67,7 +67,7 @@ var (
 func buildOrgNameQuery(orgID interface{}) (string, []interface{}, error) {
 	return dialect.Select("name").
 		From(TABLE_ORGANIZATIONS).
-		Where(goqu.Ex{"id": orgID, "deleted_at": nil}).
+		Where(goqu.Ex{"id": orgID}).
 		ToSQL()
 }
 
--- a/billing/credit/service.go
+++ b/billing/credit/service.go
@@ -6,8 +6,10 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-
+	"github.com/raystack/frontier/billing/customer"
+	"github.com/raystack/frontier/core/auditrecord"
 	"github.com/raystack/frontier/internal/bootstrap/schema"
+	pkgAuditRecord "github.com/raystack/frontier/pkg/auditrecord"
 )
 
 type TransactionRepository interface {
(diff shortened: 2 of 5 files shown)


internal/store/postgres/audit_record_repository.go
```diff
@@ -76,6 +83,14 @@ func (r AuditRecordRepository) Create(ctx context.Context, auditRecord auditreco
 		return auditrecord.AuditRecord{}, err
 	}
 
+	// Enrich organization name from DB
+	if auditRecord.OrgID != "" {
+		query, params, err := buildOrgNameQuery(auditRecord.OrgID)
+		if err == nil {
+			_ = r.dbc.QueryRowxContext(ctx, query, params...).Scan(&dbRecord.OrganizationName)
```
rohilsurana, line 90: in case of any error, we would loose have empty org name, are we ok with that?

internal/store/postgres/audit_record.go
```diff
@@ -267,6 +269,16 @@ func BuildAuditRecord(ctx context.Context, event pkgAuditRecord.Event, resource
 
 // InsertAuditRecordInTx inserts an audit record within a transaction
 func InsertAuditRecordInTx(ctx context.Context, tx *sqlx.Tx, record AuditRecord) error {
+	// Enrich the organization name from DB
+	if record.OrganizationID != uuid.Nil {
+		var orgName string
+		query, params, err := buildOrgNameQuery(record.OrganizationID)
+		if err == nil {
+			_ = tx.QueryRowContext(ctx, query, params...).Scan(&orgName)
```
rohilsurana, line 277: in case of any error, we would loose have empty org name, are we ok with that?

rohilsurana (approved):

internal/store/postgres/audit_record_repository.go
AmanGIT07: yes, we are ok with blank org name as it will possible for a few cases

internal/store/postgres/audit_record.go
AmanGIT07: yes, we are ok with blank org name as it will possible for a few cases

PR #1220, opened 2025-10-24
feat: add audit logging for session revoke

AmanGIT07:


internal/store/postgres/session_repository.go
```diff
@@ -93,29 +97,96 @@ func (s *SessionRepository) Get(ctx context.Context, id uuid.UUID) (*frontierses
...
+		},
+		"", // sessions are not org-specific, will be stored as zero uuid
+		nil,
+		deletedAt,
+	)
+
+	return InsertAuditRecordInTx(ctx, tx, auditRecord)
+}
+
+func extractActorIDFromContext(ctx context.Context) string {
```
rohilsurana: is this something we are doing for the first time? Do we need to instead extract this function out?

AmanGIT07: Used an existing method

PR #1228, opened 2025-10-29
feat: migrate roles page to connect

rohanchkrabrty:
This PR updates the Roles page in IAM Admin to use connect


ui/src/containers/roles.list/index.tsx
```diff
@@ -1,59 +1,51 @@
...
-    }
-    getRoles();
-  }, []);
-  let { roleId } = useParams();
+  const { data: roles = [], isLoading } = useQuery(
+    FrontierServiceQueries.listRoles,
+    {},
+    {
+      select: data => data?.roles ?? [],
+    },
```
rohilsurana, line 34: lets log the error in case query fails?

rohanchkrabrty: Yeah good catch

rohilsurana (approved):

PR #1229, opened 2025-10-29
feat: migrate invoice list to connect

rohanchkrabrty:
This PR migrates the Invoice page in IAM Admin to use connect


ui/src/pages/invoices/list/index.tsx
```diff
@@ -29,50 +35,117 @@ const NoInvoices = () => {
...
+    data: infiniteData,
+    isLoading,
+    isFetchingNextPage,
+    fetchNextPage,
+    error,
+    isError,
+    hasNextPage,
+  } = useInfiniteQuery(
+    AdminServiceQueries.searchInvoices,
+    { query: query },
```
rohilsurana: lets use create() from protobuf package here

ui/src/pages/invoices/list/index.tsx
```diff
@@ -29,50 +35,117 @@ const NoInvoices = () => {
...
+      await fetchNextPage();
+    } catch (error) {
+      console.error("Error loading more invoices:", error);
+    }
+  };
+
+  const columns = useMemo(
+    () =>
+      getColumns({
+        groupCountMap: infiniteData
```
rohilsurana: is this different from `const groupCountMap` ? can we just reuse that here?

rohilsurana (changes requested):

ui/src/pages/invoices/list/index.tsx
rohilsurana: also looks like there is no grouping in invoices list page, is that expected? and in that case do we need this?

ui/src/pages/invoices/list/index.tsx
rohanchkrabrty: is `create()` usage needed here for query request params?  connect-query handles the type right?

ui/src/pages/invoices/list/index.tsx
rohanchkrabrty: Yeah good catch, removed this as there is no grouping needed

ui/src/pages/invoices/list/index.tsx
rohilsurana: it would be good to have runtime checks

rohanchkrabrty: The `transformDataTableQueryToRQLRequest` function already makes the query using `create(RQLRequestSchema)`

rohilsurana: Ahh ok missed that

PR #1259, opened 2025-11-05
feat(sdk): migrate api tokens screen to connect query

rsbh:
This PR migrates the APIs for the API Tokens screen (Service Account and Service Account Tokens) to connect rpc.
This also simplifies the logic of the service user token list by using the connect query state as shared state and a new hook to manage it.

When we create service user token, then we get token in the response of create call. but in list call response then token will not come as it is visible for 1 time and we dont store in backend.

So when we open the list service user token list, we should show existing token details, and when a new token is added, we have to show it in the same list with the token visible. 
When a user creates a new service user, we have to automatically generate an initial token for them and display it in the list.

So after creating service user, we generate token and manually set it as query data for list api. and we dont refetch the api. all new token will append to the same query data and revoke will remove them.

### DIFF
--- a/sdks/js/packages/core/react/components/organization/api-keys/add.tsx
+++ b/sdks/js/packages/core/react/components/organization/api-keys/add.tsx
@@ -16,6 +16,7 @@ import { useFrontier } from '~/react/contexts/FrontierContext';
 import * as yup from 'yup';
 import { yupResolver } from '@hookform/resolvers/yup';
 import { useCallback } from 'react';
+import { orderBy } from 'lodash';
 import {
   FrontierServiceQueries,
   CreateServiceUserRequestSchema,
@@ -82,9 +83,7 @@ export const AddServiceAccount = () => {
       enabled: Boolean(orgId),
       select: data => {
         const list = data?.projects ?? [];
-        return list.sort((a, b) =>
-          (a?.title || '') > (b?.title || '') ? 1 : -1
-        );
+        return orderBy(list, ['title'], ['asc']);
       }
     }
   );


sdks/js/packages/core/react/components/organization/api-keys/add.tsx
```diff
@@ -56,82 +71,125 @@ export const AddServiceAccount = () => {
...
+    {
+      enabled: Boolean(orgId),
+      select: data => {
+        const list = data?.projects ?? [];
+        return list.sort((a, b) =>
+          (a?.title || '') > (b?.title || '') ? 1 : -1
+        );
+      }
+    }
+  );
```
rohanchkrabrty, line 89: can we put this logic inside a util or something? Getting reused more than once

sdks/js/packages/core/react/components/organization/api-keys/add.tsx
```diff
@@ -56,82 +71,125 @@ export const AddServiceAccount = () => {
...
+            orgId
+          }),
+          cardinality: 'finite'
+        });
+
+        queryClient.setQueryData(
+          listTokensQueryKey,
+          create(ListServiceUserTokensResponseSchema, {
+            tokens: tokenResponse.token ? [tokenResponse.token] : []
+          })
```
rohanchkrabrty, line 167: since we have created a hook, it would be better to use that here.

rohanchkrabrty (changes requested):

sdks/js/packages/core/react/components/organization/api-keys/add.tsx
rsbh: The hooks has the dependency on `serviceUserId`. In this component, the serviceUser has not been created yet so we cant initialize the hook. We get the `serviceUserId` in form `onSumit` callback just before the naviation logic.

sdks/js/packages/core/react/components/organization/api-keys/add.tsx
rsbh: We dont need logic in utils. we can use `orderBy` from lodash

sdks/js/packages/core/react/components/organization/api-keys/add.tsx
rohanchkrabrty: Understood

sdks/js/packages/core/react/components/organization/api-keys/add.tsx
rohanchkrabrty: Let's use it then

rsbh: i already did
https://github.com/raystack/frontier/pull/1259/commits/cb2e2576d3ce797abff72bbde7af3716227bbbd8

rohanchkrabrty (approved):

PR #1263, opened 2025-11-06
fix: preserve old values in audit records for org and billing customer updates

AmanGIT07:
## Problem

When updating organizations or billing customers, audit records were capturing the **NEW** values instead of **OLD** values. This happened because we were reading the resource data after the UPDATE statement within the same transaction, which returns the updated values due to "read your own writes" behavior.

This broke audit trail continuity - we couldn't see what the resource was called *before* the update, making it difficult to trace changes over time.

 ### Example of the issue:
  Update org title: "Acme Corp" → "Acme Corporation"
  Audit record showed: resource_name="Acme Corporation" (NEW)
  Should have shown: resource_name="Acme Corp" (OLD)

  ## Solution

  ### 1. Fetch old values BEFORE the update
  - **Organization updates** (`UpdateByID`, `UpdateByName`): Fetch title before UPDATE statement
  - **Billing customer updates** (`UpdateByID`): Fetch customer name before UPDATE statement

  ### 2. Use old values in audit records
  - Store old title/name as the resource name in audit records
  - Add both old and new values to metadata only when they actually change
  - Follow existing naming convention: `title`/`updated_title`, `name`/`updated_name`

  ### 3. Pre-populate OrganizationName
  - Modified `InsertAuditRecordInTx` to only fetch org name if not already set
  - Pre-populate `AuditRecord.OrganizationName` with old title to prevent fetching updated value

  ### 4. Code quality improvements
  - Extracted common logic into `buildOrgUpdateAuditRecord` helper function
  - Eliminated duplication between UpdateByID and UpdateByName
  - Optimized to avoid copying metadata when values don't change

  ## Files Changed

  - `internal/store/postgres/organization_repository.go`
    - Added `buildOrgUpdateAuditRecord` helper
    - Modified `UpdateByID` and `UpdateByName` to fetch and preserve old title

  - `internal/store/postgres/billing_customer_repository.go`
    - Modified `UpdateByID` to fetch and preserve old customer name

  - `internal/store/postgres/audit_record.go`
    - Updated `InsertAuditRecordInTx` to respect pre-populated OrganizationName

### DIFF
--- a/internal/store/postgres/organization_repository.go
+++ b/internal/store/postgres/organization_repository.go
@@ -279,6 +279,43 @@ func (r OrganizationRepository) List(ctx context.Context, flt organization.Filte
 	return transformedOrgs, nil
 }
 
+// buildOrgUpdateAuditRecord creates an audit record for organization updates
+func buildOrgUpdateAuditRecord(ctx context.Context, orgBeforeUpdate, orgAfterUpdate Organization, metadata map[string]interface{}) AuditRecord {
+	title := nullStringToString(orgBeforeUpdate.Title)
+	updatedTitle := nullStringToString(orgAfterUpdate.Title)
+
+	auditMetadata := metadata
+	if title != updatedTitle {
+		// Create a new one to avoid mutating the original
+		auditMetadata = make(map[string]interface{})
+		if metadata != nil {
+			for k, v := range metadata {
+				auditMetadata[k] = v
+			}
+		}
+		auditMetadata["title"] = title
+		auditMetadata["updated_title"] = updatedTitle
+	}
+
+	auditRecord := BuildAuditRecord(
+		ctx,
+		auditrecord.OrganizationUpdateEvent,
+		AuditResource{
+			ID:       orgAfterUpdate.ID,
+			Type:     auditrecord.OrganizationType,
+			Name:     title,
+			Metadata: auditMetadata,
+		},
+		nil,
+		orgAfterUpdate.ID,
+		nil,
+		orgAfterUpdate.UpdatedAt,
+	)
+	// Pre-populate OrganizationName with title to avoid fetching the updated title from DB
+	auditRecord.OrganizationName = title
+	return auditRecord
+}
+
 func (r OrganizationRepository) UpdateByID(ctx context.Context, org organization.Organization) (organization.Organization, error) {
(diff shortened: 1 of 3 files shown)


internal/store/postgres/organization_repository.go
```diff
@@ -309,25 +354,19 @@ func (r OrganizationRepository) UpdateByID(ctx context.Context, org organization
 	var orgModel Organization
 	if err = r.dbc.WithTxn(ctx, sql.TxOptions{}, func(tx *sqlx.Tx) error {
 		return r.dbc.WithTimeout(ctx, TABLE_ORGANIZATIONS, "Update", func(ctx context.Context) error {
-			if err := tx.QueryRowxContext(ctx, query, params...).StructScan(&orgModel); err != nil {
+			// Fetch title before update
+			var orgBeforeUpdate Organization
+			if err := tx.QueryRowxContext(ctx, getQuery, getParams...).StructScan(&orgBeforeUpdate); err != nil {
```
rohilsurana, line 359: seems like the query will only get title field but we are reading the whole org struct, can we instead just read into a variable for title?

internal/store/postgres/organization_repository.go
```diff
@@ -279,6 +279,43 @@ func (r OrganizationRepository) List(ctx context.Context, flt organization.Filte
...
+			Name:     title,
+			Metadata: auditMetadata,
+		},
+		nil,
+		orgAfterUpdate.ID,
+		nil,
+		orgAfterUpdate.UpdatedAt,
+	)
+	// Pre-populate OrganizationName with title to avoid fetching the updated title from DB
+	auditRecord.OrganizationName = title
```
rohilsurana, line 315: name and title inconsistency?

internal/store/postgres/organization_repository.go
AmanGIT07: This variable makes it easy to pass the params to `buildOrgUpdateAuditRecord` method which is a generic method that accepts the org struct. If we just read into variable and pass the same in that method, we will have to keep updating the method for any new field.

internal/store/postgres/organization_repository.go
AmanGIT07: For org, this is correct.

rohilsurana (approved):

PR #1276, opened 2025-11-11
feat(sdk): migrate billing screens to connect rpc

rsbh:
This PR migrates the Billing screen APIs in the Frontier SDK to use RPC calls.

Following APIs have been migrates to connect-query and connect rpc

1. ListInvoices
2. CreateCheckout
3. GetUpcomingInvoice
4. ListOrganizationUsers

The types and interfaces are also updated to use types from Proton instead of OpenAPI contract.
Object keys have been changed to camelCase instead of snake_case as per protobuf definition.

### DIFF
--- a/sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
+++ b/sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
@@ -67,36 +67,35 @@ export function ConfirmCycleSwitch() {
   const isLoading = isAllPlansLoading;
 
   async function onConfirm() {
-    if (nextPlan?.id) {
-      const nextPlanId = nextPlan?.id;
-      if (isPaymentMethodRequired) {
-        checkoutPlan({
-          planId: nextPlanId,
-          isTrial: false,
-          onSuccess: data => {
-            window.location.href = data?.checkoutUrl as string;
-          }
-        });
-      } else
-        changePlan({
-          planId: nextPlanId,
-          onSuccess: async () => {
-            const planPhase = await verifyPlanChange({
-              planId: nextPlanId
+    const nextPlanId = nextPlan?.id;
+    if (!nextPlanId) return;
+    if (isPaymentMethodRequired) {
+      checkoutPlan({
+        planId: nextPlanId,
+        isTrial: false,
+        onSuccess: data => {
+          window.location.href = data?.checkoutUrl as string;
+        }
+      });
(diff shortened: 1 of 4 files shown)


sdks/js/packages/core/react/components/organization/billing/cycle-switch/index.tsx
```diff
@@ -71,45 +67,35 @@ export function ConfirmCycleSwitch() {
...
-              window.location.href = data?.checkout_url as string;
-            }
-          });
-        } else
-          changePlan({
-            planId: nextPlanId,
-            onSuccess: async () => {
-              const planPhase = await verifyPlanChange({
-                planId: nextPlanId
+    if (nextPlan?.id) {
```
rohanchkrabrty: can we do early return for this? `if (!nextPlan.id) return;`

we can avoid nesting with this

sdks/js/packages/core/react/components/organization/billing/index.tsx
```diff
@@ -140,35 +141,32 @@ export default function Billing() {
...
-        setInvoices(newInvoices);
-      } catch (err) {
-        console.error(err);
-      } finally {
-        setIsInvoicesLoading(false);
-      }
-    },
-    [client]
+  const { data: invoices = [], isLoading: isInvoicesLoading } = useConnectQuery(
+    FrontierServiceQueries.listInvoices,
```
rohanchkrabrty, line 151: need to handle errors for listInvoices

sdks/js/packages/core/react/components/organization/billing/index.tsx
rohanchkrabrty: can we do early return for this too

sdks/js/packages/core/react/components/organization/billing/payment-method.tsx
```diff
@@ -40,46 +53,37 @@ export const PaymentMethod = ({
     const orgId = billingAccount?.orgId || '';
     const billingAccountId = billingAccount?.id || '';
     if (billingAccountId && orgId) {
```
rohanchkrabrty: same for this

sdks/js/packages/core/react/components/organization/billing/upcoming-billing-cycle.tsx
rohanchkrabrty, line 1: we should handle errors for both the queries

PR #1277, opened 2025-11-11
feat: add location new fields and change structure

paanSinghCoder:
1. Change location mapping as API structure has changed.
2. Add metadata to `frontierContext` as `sessionMetadata`
3. Update proto

### DIFF
--- a/ui/src/pages/organizations/details/contexts/organization-context.tsx
+++ b/ui/src/pages/organizations/details/contexts/organization-context.tsx
@@ -1,13 +1,14 @@
-import { Organization, OrganizationSchema } from "@raystack/proton/frontier";
+import {
+  Organization,
+  OrganizationSchema,
+  type Role,
+  type BillingAccount,
+  type User,
+  type OrganizationKyc,
+  type BillingAccountDetails,
+} from "@raystack/proton/frontier";
 import { create } from "@bufbuild/protobuf";
 import { createContext } from "react";
-import type {
-  V1Beta1Role,
-  V1Beta1BillingAccount,
-  V1Beta1User,
-  V1Beta1OrganizationKyc,
-  V1Beta1BillingAccountDetails,
-} from "~/api/frontier";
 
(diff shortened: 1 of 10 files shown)


ui/src/pages/users/details/security/sessions/index.tsx
```diff
@@ -11,6 +11,37 @@ import relativeTime from "dayjs/plugin/relativeTime";
...
+ * Note: This function also exists in the SDK utils/index.ts file. 
+ * If you make any changes here, please update the SDK utils/index.ts file as well.
+ */
+const formatLocation = (location?: {
+  city?: string;
+  country?: string;
+  latitude?: string;
+  longitude?: string;
+}): string => {
+  if (!location) return 'Unknown location';
```
rohilsurana: The one in sdk just says `Unknown`, can we keep these same?

paanSinghCoder: Good catch.

sdks/js/packages/core/react/hooks/useLastActiveTracker.ts
```diff
@@ -15,4 +16,5 @@ export const useLastActiveTracker = ({ enabled = false }) => {
       retry: false
     }
   );
+  return { data: data, isLoading, error: error ? new Error(error.message) : undefined };
```
rohilsurana: Why are we converting the error here?

ui/src/pages/users/details/security/sessions/index.tsx
paanSinghCoder:
```suggestion
  if (!location) return 'Unknown';
```

sdks/js/packages/core/react/hooks/useLastActiveTracker.ts
paanSinghCoder: Not needed. We can just return error directly. Done

PR #1304, opened 2025-11-28
fix(sdk): remove `select` from `useQuery` to prevent rerender

rsbh:
## Summary

This PR fixes the issue of the rerender of the Service user Projects screen in IAM SDK.
The issue is that select returns an array, which is a dependency of `useEffect`, and in `useEffect`, we call setState, which rerenders the page, and again select returns a new array. 

This is a common issue with useQuery select https://github.com/TanStack/query/discussions/3769

## Changes

This PR replaces that with useMemo and uses data from the useQuery return value. 
The issue is happening only here due to `useEffect` and setting state, but we replace the select in other pages to prevent this in future

## Test Plan

### DIFF
--- a/sdks/js/packages/core/react/components/organization/api-keys/service-user/projects.tsx
+++ b/sdks/js/packages/core/react/components/organization/api-keys/service-user/projects.tsx
@@ -28,6 +28,7 @@ import {
   Project,
   Policy
 } from '@raystack/proton/frontier';
+import { orderBy } from 'lodash';
 
 type ProjectAccessMap = Record<string, { value: boolean; isLoading: boolean }>;
 
@@ -105,10 +106,7 @@ export default function ManageServiceUserProjects() {
 
   const orgId = organization?.id || '';
 
-  const {
-    data: projectsData,
-    isLoading: isProjectsLoading
-  } = useQuery(
+  const { data: projectsData, isLoading: isProjectsLoading } = useQuery(
     FrontierServiceQueries.listOrganizationProjects,
     create(ListOrganizationProjectsRequestSchema, {
       id: orgId,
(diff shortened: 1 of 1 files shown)


sdks/js/packages/core/react/components/organization/api-keys/service-user/projects.tsx
```diff
@@ -116,20 +116,21 @@ export default function ManageServiceUserProjects() {
...
-            : -1
-        );
-      }
+      enabled: Boolean(orgId)
     }
   );
 
+  const projects = useMemo(() => {
+    const list = projectsData?.projects ?? [];
+    return list.sort((a, b) =>
```
paanSinghCoder: I think we should avoid mutating the original data. Instead, we can create a new array copy, run .sort() on it and return.

`return [...list].sort()`

sdks/js/packages/core/react/components/organization/api-keys/hooks/useServiceUserTokens.ts
```diff
@@ -26,18 +27,19 @@ export const useServiceUserTokens = ({
...
       orgId
     }),
     {
-      enabled: Boolean(id) && Boolean(orgId) && Boolean(enableFetch),
-      select: data => data?.tokens ?? []
+      enabled: Boolean(id) && Boolean(orgId) && Boolean(enableFetch)
     }
   );
 
+  const tokens = useMemo(() => tokensData?.tokens ?? [], [tokensData]);
```
paanSinghCoder, line 41: I think we should add a comment with each useMemo so that in future dev knows why select was replaced with useMemo pattern and they don't add it back.

sdks/js/packages/core/react/components/organization/api-keys/service-user/projects.tsx
rsbh: updated with lodash `orderBy`

rsbh: update it with `orderBy`

PR #1314, opened 2025-12-04
feat(sdk): remove name from project create

rsbh:
## Summary

This PR removes the name field from the project creation form.

## Changes

The name will be auto-generated from the title, and a hash from the org ID will be added as a suffix.
Due to this.
1. Projects with the same title can be created in different organizations.
2. Projects with the same title can't be created in the same organization.

## Technical Details

We are using the `djb2` hashing algorithm to generate a hash from the string.
The error capture has been moved from `onError` to catch block https://github.com/TanStack/query/discussions/1784 

## Test Plan

### DIFF
--- a/sdks/js/packages/core/hooks/index.ts
+++ b/sdks/js/packages/core/hooks/index.ts
@@ -7,6 +7,9 @@ export {
   createConnectQueryKey
 } from '@connectrpc/connect-query';
 
+// Re-export Connect utilities
+export { ConnectError, Code } from '@connectrpc/connect';
+
 // Re-export Frontier service queries for convenience
 export { FrontierServiceQueries } from '@raystack/proton/frontier';
 


sdks/js/packages/core/react/components/organization/project/add.tsx
```diff
@@ -61,32 +58,38 @@ export const AddProject = () => {
...
-        body: {
-          title: data.title,
-          name: data.name,
-          orgId: organization.id
-        }
-      })
-    );
+    const slug = slugify(data.title, { lower: true, strict: true });
+    const suffix = generateHashFromString(organization.id);
+    const name = `${slug}-${suffix}`;
```
paanSinghCoder, line 69: I think we should name the var slug instead of name.

sdks/js/packages/core/react/components/organization/project/add.tsx
```diff
@@ -61,32 +58,38 @@ export const AddProject = () => {
...
+            name,
+            orgId: organization.id
+          }
+        })
+      );
+    } catch (error) {
+      if (error instanceof ConnectError && error.code === Code.AlreadyExists) {
+        setError('title', {
+          message:
+            'A project with a similar title already exist. Please tweak the title and try again.'
```
paanSinghCoder, line 84: similar vs same?

I think `same` should be the copy. `A Project with the same title already exists.` or  `A Project with the title already exists.` Thoughts?

rsbh: The slug is in lower case.
So Title and title will create same slug

paanSinghCoder (approved):

PR #1356, opened 2026-01-29
fix: formatting issue for negative tokens

paanSinghCoder:
Screenshot 2026-01-29 at 4 49 52 PM

## Summary
**Issue**: 
- Values like -595 were shown as -,595 because the minus sign was included in the length used for comma placement, so a comma was inserted after the sign.
- Very big values loose accuracy as `Number()` is used.

**Fix**: 

- Use `BigInt()` to convert the string to a bigInt without loosing the accuracy.
- Implement `num.toLocaleString();` to format the number.
- Get locale from frontier config. Defaults to undefined.

## Test Plan

- [x] Manually tested on local

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -326,7 +326,12 @@ func buildAPIDependencies(
 	resourceBlobRepository *blob.ResourcesRepository,
 	planBlobRepository *blob.PlanRepository,
 ) (api.Deps, error) {
-	preferenceService := preference.NewService(postgres.NewPreferenceRepository(dbc))
+	// Load custom traits from config file if specified
+	traits, err := preference.LoadTraitsFromFile(cfg.App.CustomTraitsPath)
+	if err != nil {
+		return api.Deps{}, fmt.Errorf("failed to load custom traits: %w", err)
+	}
+	preferenceService := preference.NewService(postgres.NewPreferenceRepository(dbc), traits)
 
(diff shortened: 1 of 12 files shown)


rohanchkrabrty (changes requested):
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

Let's use this to format the number.

```js
const n = 34523453.345;
console.log(n.toLocaleString());    // "34,523,453.345"
```

web/lib/react/utils/index.ts
```diff
@@ -124,17 +124,15 @@ export const checkSimilarPlans = (
...
-  }, '');
+/**
+ * Formats a number or bigint as a locale-aware string (e.g. with thousand separators).
+ * Uses the runtime's default locale (e.g. the user's browser/system language).
+ *
+ * @param num - The number or bigint to format. Defaults to 0.
+ * @returns The formatted string (e.g. "1,234,567" in user's locale).
+ */
+export function getFormattedNumberString(num: number | bigint = 0) {
+  return num.toLocaleString();
```
rsbh: Can we add a locale in the Frontier config and use it here?
we already does it for date. 
If the consumer wants to hard-code the currency format, then this will not work.

rohanchkrabrty (approved):

PR #1383, opened 2026-02-16
feat: export Webhooks page as a component and consume it in apps/admin

paanSinghCoder:
### Summary
- **Moved** webhooks from `apps/admin/src/containers/webhooks` to `lib/admin/views/webhooks` using `git mv`. Supporting pieces (SheetFooter, CustomField, multiselect, webhook-events) were also moved into lib via `git mv`.
- **Lib:** Single **WebhooksView** (same pattern as RolesView/ProductsView): list + create/update sheets driven by `selectedWebhookId`, `createOpen`, `onCloseDetail`, `onSelectWebhook`, `onOpenCreate`, `enableDelete`. No `<Outlet />`; create/update live inside the view.
- **App:** **WebhooksPage** matches RolesPage/ProductsPage: one page for `/webhooks`, `/webhooks/create`, `/webhooks/:webhookId`, passing URL (params + `useMatch`) into WebhooksView. 
- **Build:** Admin tsup externals updated (e.g. `react-dom`, `zod`, `sonner`). Relative imports in lib fixed for the admin build. CustomField multiselect import path fixed.
- **Result:** Webhooks lives in lib, is exported as `WebhooksView`, and is consumed from the app in the same way as roles and products.


web/lib/admin/views/webhooks/webhooks/create/index.tsx
```diff
@@ -88,7 +93,7 @@ export default function CreateWebhooks() {
             <SheetHeader
               title="Add new Webhook"
               onClick={onOpenChange}
-              data-test-id="admin-add-new-webhook-btn"
+              data-testid="admin-add-new-webhook-btn"
```
rsbh: we dont need to change it. We use the `data-test-id` attribute in Playwright.

web/lib/tsup.config.ts
```diff
@@ -43,7 +43,16 @@ export default defineConfig(() => [
       js: "'use client'"
     },
     format: ['cjs', 'esm'],
-    external: ['react', 'react-router-dom'],
+    external: [
+      'react',
```
rsbh: We can import  `devDependencies` from package.json here

web/lib/admin/views/webhooks/webhooks/create/index.tsx
paanSinghCoder: Done

web/lib/tsup.config.ts
paanSinghCoder: Done

PR #1390, opened 2026-02-18
feat: export Users page as a component and consume it in apps/admin

paanSinghCoder:
### Summary
- Users UI moved from `apps/admin/src/pages/users` to `lib/admin/views/users` and exposed as a reusable view.

### Lib (`web/lib/admin/views/users`)
- **`UsersView`** with props: `selectedUserId?`, `onCloseDetail?`, `onExportUsers?`, `onNavigateToUser?`.
- List vs detail driven by `selectedUserId`; no `Outlet` in the lib.
- List columns use `<Link>` for user routes; app controls navigation via `onNavigateToUser`.

### App (`apps/admin`)
- **`UsersPage`** uses `UsersView` with:
  - `selectedUserId` from route params
  - `onExportUsers` for CSV export
  - `onNavigateToUser` → `navigate(\`/users/${id}/security\`)`
- Nested routes (`users`, `users/:userId`, `users/:userId/security`) all render `UsersPage`, which delegates to `UsersView`.

### Why
- Reuse the same Users UI across apps; routing and navigation stay in the host app.

### DIFF
--- a/web/sdk/admin/components/AssignRole.tsx
+++ b/web/sdk/admin/components/AssignRole.tsx
@@ -26,6 +26,7 @@ import { createClient } from "@connectrpc/connect";
 import { useForm } from "react-hook-form";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { z } from "zod";
+import { SCOPES } from "../utils/constants";
 
 interface AssignRoleProps {
   organizationId: string;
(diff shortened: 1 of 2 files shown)


web/sdk/admin/components/AssignRole.tsx
```diff
@@ -0,0 +1,213 @@
...
+      );
+      await Promise.all(
+        removedRolesPolicies.map((policy: Policy) =>
+          deletePolicy(
+            create(DeletePolicyRequestSchema, { id: policy.id || "" }),
+          ),
+        ),
+      );
+
+      const resource = `app/organization:${organizationId}`;
```
rsbh: We can use the `SCOPES` from constant file here

paanSinghCoder: Done

rsbh (approved):

PR #1402, opened 2026-02-23
fix: set connect query to 2.1.1

rsbh:
## Summary

This PR hardcodes connect-query to version 2.1.1
We were using 2.1.1 before, but recently we started using 2.2.0 because of the `^` in the package.json and the need to update the pnpm lock file. 

The issue is that our pageQuery is nested inside the `query`. But we were passing `query` as the key, since we can't pass `query.offset`, as it doesn't support that and it will throw TS issue.
But in the [recent changes to the library](https://github.com/connectrpc/connect-query-es/pull/559), it ignores the page query when generating the infinite query key. 
We were passing `query` as the page query and search, and the filter was nested inside it, and they stopped working.
Either we move the `offset` ouside query or add support for nested keys in connect-query. 
 
 An issue has been raised in connect-query as well https://github.com/connectrpc/connect-query-es/issues/569

### DIFF
--- a/web/apps/admin/package.json
+++ b/web/apps/admin/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@bufbuild/protobuf": "^2.11.0",
     "@connectrpc/connect": "^2.1.0",
-    "@connectrpc/connect-query": "^2.1.1",
+    "@connectrpc/connect-query": "2.1.1",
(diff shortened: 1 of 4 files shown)


paanSinghCoder (approved):
Looking at this scenario, I think we should consider removing`^` for other crucial packages as well.

It will add an overhead for us to manually upgrade packages but will keep things predictable.

PR #1442, opened 2026-03-09
feat: PAT authentication chain, token validation, and error handling

AmanGIT07:
-  Add PAT client assertion, token validator (prefix → decode → hash → DB lookup → expiry), and wire into authenticate service.
- Refactor GetPrincipal into composable AuthenticatorFunc chain with errSkip sentinel for fall-through.
-  Handle PAT-specific errors: malformed/expired/disabled PATs return CodeUnauthenticated instead of CodeInternal. SpiceDB write rejections for PAT subjects on disallowed relations return CodePermissionDenied instead of 500.
-   Restructure userpat package into models/ and errors/ subpackages.
-   Add GetBySecretHash and UpdateLastUsedAt repository methods. Support
-   PAT prefix tokens in session interceptor.


core/authenticate/authenticators.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/raystack/frontier/pkg/utils"
+)
+
+// AuthenticatorFunc attempts to authenticate a request.
+// Returns (Principal, nil) on success, errSkip if not applicable (try next),
+// or any other error for a terminal authentication failure.
+type AuthenticatorFunc func(ctx context.Context, s *Service) (Principal, error)
+
+// errSkip signals that this authenticator doesn't apply to the request.
+var errSkip = errors.New("skip authenticator")
```
whoAbhishekSah: This line can go inside errors.go of this directory.

core/userpat/models/pat.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+type PAT struct {
+	ID         string `rql:"name=id,type=string"`
+	UserID     string `rql:"name=user_id,type=string"`
+	OrgID      string `rql:"name=org_id,type=string"`
+	Title      string `rql:"name=title,type=string"`
+	SecretHash string `json:"-"`
+	Metadata   metadata.Metadata
+	LastUsedAt *time.Time `rql:"name=last_used_at,type=datetime"`
```
whoAbhishekSah: let's use one type, either `*time.Time` or `time.time`?

core/userpat/validator.go
```diff
@@ -0,0 +1,74 @@
...
+
+	paterrors "github.com/raystack/frontier/core/userpat/errors"
+	"github.com/raystack/frontier/core/userpat/models"
+	"github.com/raystack/salt/log"
+	"golang.org/x/crypto/sha3"
+)
+
+// Validator validates PAT values during authentication.
+// It is separate from Service to avoid circular dependencies in wiring
+// (authenticate.Service → userpat.Service → orgService → authenticate.Service).
```
whoAbhishekSah: Looks like a redundant comment.

core/userpat/validator.go
```diff
@@ -0,0 +1,74 @@
...
+
+	if pat.ExpiresAt.Before(time.Now()) {
+		return models.PAT{}, paterrors.ErrExpired
+	}
+
+	// async last_used_at update — don't block the auth path
+	go func() {
+		if err := v.repo.UpdateLastUsedAt(context.Background(), pat.ID, time.Now()); err != nil {
+			v.logger.Error("failed to update PAT last_used_at", "pat_id", pat.ID, "error", err)
+		}
```
whoAbhishekSah: Why goroutine? Why not do the DB write in the same flow and return an error if it fails?

core/authenticate/authenticators.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// authenticateWithPAT validates a personal access token.
+func authenticateWithPAT(ctx context.Context, s *Service) (Principal, error) {
+	value, ok := GetTokenFromContext(ctx)
+	if !ok {
+		return Principal{}, errSkip
+	}
+
+	pat, err := s.userPATService.Validate(ctx, value)
```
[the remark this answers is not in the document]
AmanGIT07: No nil guard needed — it's always wired in cmd/serve.go

core/userpat/validator_test.go
```diff
@@ -0,0 +1,136 @@
...
+		require.NoError(t, err)
+		assert.Equal(t, expectedPAT.ID, pat.ID)
+		assert.Equal(t, expectedPAT.UserID, pat.UserID)
+		assert.Equal(t, expectedPAT.OrgID, pat.OrgID)
+		assert.Equal(t, expectedPAT.Title, pat.Title)
+
+		// wait briefly for the async goroutine to complete
+		time.Sleep(50 * time.Millisecond)
+		repo.AssertCalled(t, "UpdateLastUsedAt", mock.Anything, "pat-1", mock.AnythingOfType("time.Time"))
+	})
```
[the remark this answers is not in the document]
AmanGIT07: Fixed. Using a channel + select with timeout now

core/userpat/validator.go
```diff
@@ -0,0 +1,74 @@
...
+	}()
```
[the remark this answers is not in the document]
AmanGIT07: Will revisit with coalescing if write volume becomes a concern

core/userpat/models/pat.go
AmanGIT07: It is intentional — LastUsedAt is nullable (nil until first use), while ExpiresAt/CreatedAt/UpdatedAt are always set at creation time.

core/userpat/validator.go
AmanGIT07: Trimmed to one line.

core/userpat/validator.go
AmanGIT07: last_used_at is just observability metadata, not critical to the auth decision. Didn't want to add latency or block the original flow. Hence kept it fire and forget

core/userpat/models/pat.go
whoAbhishekSah: Can we add a comment for the same?

core/userpat/validator.go
whoAbhishekSah: Don't we need it for audit purposes?

AmanGIT07: This field is not critical to Audit. We will be creating audit record for the CRUD operations of PAT, and this field is not touched in any of these operations.  (This field just tells when this PAT was last used, helpful when user goes to revoke screen, they can revoke the one that seems dormant)

core/userpat/models/pat.go
AmanGIT07: will do

AmanGIT07: Done

core/userpat/validator.go
AmanGIT07: For now we will be going ahead with synchronous call, we can revisit it once this starts causing issue.

core/userpat/service.go
```diff
@@ -203,11 +205,11 @@ func (s *Service) resolveAndValidateRoles(ctx context.Context, roleIDs []string)
...
-			return nil, fmt.Errorf("role %s has scopes %v: %w", r.Name, r.Scopes, ErrUnsupportedScope)
+			return nil, fmt.Errorf("role %s has no scopes defined: %w", r.Name, paterrors.ErrUnsupportedScope)
 		}
 		for _, scope := range r.Scopes {
 			if scope != schema.ProjectNamespace && scope != schema.OrganizationNamespace {
-				return nil, fmt.Errorf("role %s has scopes %v: %w", r.Name, r.Scopes, ErrUnsupportedScope)
+				return nil, fmt.Errorf("role %s has scopes %v: %w", r.Name, r.Scopes, paterrors.ErrUnsupportedScope)
 			}
 		}
 	}
```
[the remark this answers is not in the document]
AmanGIT07: This needs a transaction control between app db and auth db, which is currently a limitation

core/userpat/service.go
```diff
@@ -125,11 +127,11 @@ func (s *Service) Create(ctx context.Context, req CreateRequest) (PAT, string, e
...
 	created, err := s.repo.Create(ctx, pat)
 	if err != nil {
-		return PAT{}, "", err
+		return patmodels.PAT{}, "", err
 	}
 
 	if err := s.createPolicies(ctx, created.ID, req.OrgID, roles, req.ProjectIDs); err != nil {
-		return PAT{}, "", fmt.Errorf("creating policies: %w", err)
+		return patmodels.PAT{}, "", fmt.Errorf("creating policies: %w", err)
 	}
```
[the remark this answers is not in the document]
AmanGIT07: This needs a transaction control between App DB and Auth DB(spiceDB) which is currently a limitation

PR #1451, opened 2026-03-17
fix: add tooltip for row text

paanSinghCoder:
## Summary
- Show full description on hover via `Tooltip`. Cell text truncates with `ellipsis`, tooltip uses apsara `Tooltip` with `delayDuration={500}`.
- `Date` and `Tokens` columns constrained with `styles: { header: { width: "180px" } }`

https://github.com/user-attachments/assets/95867595-9c68-46fb-9df7-379f6d4efefb

## **Future enhancement: Tooltip shows always, irrespective of the text is truncated or not.**
- Detecting truncation (e.g. `scrollWidth` vs `clientWidth`) needs a `ResizeObserver` (or similar) per cell. Doing that in each column/cell is not scalable (many observers, repeated logic).
- A better approach is to support this in DataTable: e.g. a column option like `showTooltipWhenTruncated` and optional tooltip content, with a single `ResizeObserver` (or shared logic) inside the table. That keeps the feature in one place and avoids N observers per table. Left for a follow-up.

## Test Plan

### DIFF
--- a/web/sdk/admin/views/organizations/details/tokens/columns.tsx
+++ b/web/sdk/admin/views/organizations/details/tokens/columns.tsx
@@ -58,15 +58,14 @@ export const getColumns = (): DataTableColumnDef<
     {
       accessorKey: "description",
       header: "Events",
+      classNames: { cell: styles["truncate-tooltip-wrapper"] },
       cell: ({ getValue }) => {
         const text = (getValue() as string) ?? "";
         if (!text) return text;
         return (
-          <span className={styles["truncate-tooltip-wrapper"]}>
-            <Tooltip message={text} delayDuration={500}>
-              <span className={styles["truncate-text"]}>{text}</span>
-            </Tooltip>
-          </span>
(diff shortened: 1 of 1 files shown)


web/sdk/admin/views/organizations/details/tokens/columns.tsx
```diff
@@ -56,7 +59,15 @@ export const getColumns = (): DataTableColumnDef<
       accessorKey: "description",
       header: "Events",
       cell: ({ getValue }) => {
-        return getValue();
+        const text = (getValue() as string) ?? "";
+        if (!text) return text;
+        return (
+          <span className={styles["truncate-tooltip-wrapper"]}>
```
rsbh: We can pass this className to `cell` as well.
column defintions support `classNames` prop for cell and header

paanSinghCoder: Done

PR #1472, opened 2026-03-23
fix: filter by text is case sensitive

paanSinghCoder:
## Summary
Make organization like filters case-insensitive in Admin SearchOrganizations backend query handling.

Related PR - [https://github.com/raystack/salt/pull/83](https://github.com/raystack/salt/pull/83)

**Changes**
Updated frontier/internal/store/postgres/org_billing_repository.go

- like now uses ILIKE
- notlike now uses NOT ILIKE

Added regression coverage in frontier/internal/store/postgres/org_billing_repository_test.go to assert title like generates ILIKE.

**Why**
Filtering organizations by title from /organizations returned different results for different casing (e.g. fah vs Fah). This aligns filter behavior with expected case-insensitive search.

## Test Plan


internal/store/postgres/org_billing_repository.go
```diff
@@ -411,10 +411,10 @@ func processStringDataType(filter rql.Filter, query *goqu.SelectDataset) *goqu.S
 		query = query.Where(goqu.Cast(goqu.I(filter.Name), "TEXT").NotIn(strings.Split(filter.Value.(string), ",")))
 	case OPERATOR_LIKE:
 		// some semi string sql types like UUID require casting to text to support like operator
-		query = query.Where(goqu.L(fmt.Sprintf(`"%s"::TEXT LIKE '%s'`, filter.Name, filter.Value.(string))))
+		query = query.Where(goqu.L(fmt.Sprintf(`"%s"::TEXT ILIKE '%s'`, filter.Name, filter.Value.(string))))
```
rsbh: Instead of changing the existing query, add new cases in switch conditions for `OPERATOR_ILIKE` and `OPERATOR_NOT_ILIKE`.

Current solution will break the exsiting logic and misleading as the operator is `LIKE` and in query we are using `ILIKE`

rsbh (changes requested):

web/sdk/admin/utils/transform-query.ts
```diff
@@ -57,18 +57,11 @@ function transformFilter(
     value = convertFilterValue(filter.value);
   }
 
-  // TODO: add support for ilike in RQL and backend
-  // Transform ilike operator to like only for string values
-  const operator =
-    filter.operator === "ilike" && value.case === "stringValue"
```
paanSinghCoder, line 63: @rsbh Any idea why is there `value.case === "stringValue"` check?

rsbh: I dont remeber it, but i think it is redundant. `ilike` should be only in string value.

rsbh (approved):

rsbh (approved):

PR #1531, opened 2026-04-10
fix: skip redundant policy delete+create when role is unchanged

whoAbhishekSah:
## Summary
- `SetProjectMemberRole` and `SetOrganizationMemberRole` now return early if the principal already has exactly the requested role
- Avoids unnecessary SpiceDB writes (delete + create) when the role hasn't changed

## Changes
- `core/organization/service.go` — early return when single existing policy matches new role
- `core/project/service.go` — same
- Unit tests for both covering the skip path (mock verifies no `Delete` or `Create` calls)

## Test plan
- [x] Unit test: org SetMemberRole skips when role unchanged (no delete/create called)
- [x] Unit test: project SetMemberRole skips when role unchanged (no delete/create called)
- [x] Existing tests still pass (role change, promotion, demotion paths unaffected)

### DIFF
--- a/core/organization/service.go
+++ b/core/organization/service.go
@@ -393,6 +393,11 @@ func (s Service) SetMemberRole(ctx context.Context, orgID, userID, newRoleID str
 		return ErrNotMember
 	}
 
+	// skip if the user already has exactly this role
+	if len(existingPolicies) == 1 && existingPolicies[0].RoleID == newRoleID {
+		return nil
+	}
+
 	// check minimum owner constraint
 	err = s.validateMinOwnerConstraint(ctx, orgID, newRoleID, existingPolicies)
 	if err != nil {
--- a/core/organization/service_test.go
+++ b/core/organization/service_test.go
@@ -598,6 +598,25 @@ func TestService_SetMemberRole(t *testing.T) {
 			newRoleID: ownerRoleID,
 			wantErr:   nil,
 		},
+		{
+			name: "should skip delete+create when role is unchanged",
+			setup: func(repo *mocks.Repository, userSvc *mocks.UserService, roleSvc *mocks.RoleService, policySvc *mocks.PolicyService, _ *mocks.AuditRecordRepository) {
+				repo.EXPECT().GetByID(ctx, orgID).Return(organization.Organization{ID: orgID, State: organization.Enabled}, nil)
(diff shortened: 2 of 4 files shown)


whoAbhishekSah:
## Test Results

Setup: Created org `pr1531-test-1775811649` with project `pr1531-test-project`. Alice (owner), Bob (member).

| # | Test | RPC | Result |
|---|---|---|---|
| 1 | Set bob as org viewer (initial) | SetOrganizationMemberRole | `{}` |
| 2 | Set bob as org viewer again (same role — early return path) | SetOrganizationMemberRole | `{}` |
| 3 | Set bob as project viewer (initial) | SetProjectMemberRole | `{}` |
| 4 | Set bob as project viewer again (same role — early return path) | SetProjectMemberRole | `{}` |
| 5 | Change bob to project manager (role change — full delete+create) | SetProjectMemberRole | `{}` |
| 6 | Set bob as project manager again (same role — early return path) | SetProjectMemberRole | `{}` |
| 7 | Change bob to org manager (role change — full delete+create) | SetOrganizationMemberRole | `{}` |
| 8 | Set bob as org manager again (same role — early return path) | SetOrganizationMemberRole | `{}` |

All 8 tests return `{}` (success) with no errors. The "set same role" calls (tests 2, 4, 6, 8) hit the early-return optimization — they succeed without performing redundant SpiceDB delete+create operations. The role-change calls (tests 5, 7) confirm the normal path still works correctly.

core/organization/service.go
```diff
@@ -393,6 +393,11 @@ func (s Service) SetMemberRole(ctx context.Context, orgID, userID, newRoleID str
 		return ErrNotMember
 	}
 
+	// skip if the user already has exactly this role
+	if len(existingPolicies) == 1 && existingPolicies[0].RoleID == newRoleID {
```
AmanGIT07, line 397: Are we intentionally checking `len(existingPolicies) == 1`  so that if it has multiple roles it becomes just one after update?

AmanGIT07: I meant we could just do an `in` operation otherwise

whoAbhishekSah: There will always be 1 policy per user per org. Hence, doing that check,

AmanGIT07 (approved):

PR #1536, opened 2026-04-13
feat(billing): add SearchOrgInvoices RQL endpoint

paanSinghCoder:
## Summary
- add a new `SearchOrgInvoices` RPC in FrontierService that supports `RQLRequest` and pagination response
Proton: [PR-471](https://github.com/raystack/proton/pull/471)

## SearchOrgInvoices RQL Support Matrix

| Category | Supported | Notes |
|---|---|---|
| `org_id` | Yes (required) | Request is org-scoped; validation enforced. |
| `nonzero_amount_only` | Yes | `true` applies `amount > 0` filter. |
| `query.limit` | Yes | Defaults to `50` when missing/invalid. |
| `query.offset` | Yes | Defaults to `0` when missing/invalid. |
| `query.search` | Yes | Applied to searchable invoice fields. |
| `query.sort` | Yes | Works on supported fields below. |
| `query.filters` | Yes | Works on supported fields below. |
| `query.group_by` | No | Explicitly rejected (`group_by is not supported`). |

### Filterable / Sortable Fields

| Field | Type | Filter | Sort |
|---|---|---:|---:|
| `id` | string | Yes | Yes |
| `provider_id` | string | Yes | Yes |
| `customer_id` | string | Yes | Yes |
| `state` | string | Yes | Yes |
| `currency` | string | Yes | Yes |
| `amount` | number | Yes | Yes |
| `hosted_url` | string | Yes | Yes |
| `due_at` | datetime | Yes | Yes |
| `effective_at` | datetime | Yes | Yes |
| `created_at` | datetime | Yes | Yes |
| `period_start_at` | datetime | Yes | Yes |
| `period_end_at` | datetime | Yes | Yes |

### Searchable Fields

| Field |
|---|
| `id` |
| `provider_id` |
| `state` |
| `currency` |
| `hosted_url` |

### Operator Support (practical)

| Type | Common operators |
|---|---|
| `string` | `eq`, `neq`, `like`, `notlike`, `in`, `notin`, `empty`, `notempty` |
| `number` | `eq`, `neq`, `gt`, `gte`, `lt`, `lte` |
| `datetime` | `eq`, `neq`, `gt`, `gte`, `lt`, `lte` |

## Test plan
- [ ] run `make proto` and verify no generated-file diffs remain
- [ ] run billing API tests for connect handler and repository paths
- [ ] verify existing `ListInvoices` consumers still behave unchanged
- [ ] exercise `SearchOrgInvoices` with filters/sort/search/limit/offset


billing/invoice/invoice.go
```diff
@@ -113,3 +113,18 @@ type Filter struct {
 
 	Pagination *pagination.Pagination
 }
+
+type SearchOrgInvoice struct {
```
AmanGIT07, line 118: Check `InvoiceWithOrganization`, it appears to have RQL tags. Can we reuse it or we need `SearchOrgInvoice` struct separately

billing/invoice/service.go
```diff
@@ -311,6 +312,19 @@ func (s *Service) List(ctx context.Context, filter Filter) ([]Invoice, error) {
...
+func (s *Service) SearchOrgInvoices(ctx context.Context, customerID string, nonzeroOnly bool, rqlQuery *rql.Query) ([]Invoice, int64, error) {
+	if customerID == "" {
+		return nil, 0, errors.New("customer id is required")
+	}
+	if rqlQuery == nil {
+		return nil, 0, fmt.Errorf("%w: query is required", ErrBadInput)
+	}
+	if len(rqlQuery.GroupBy) > 0 {
+		return nil, 0, fmt.Errorf("%w: group_by is not supported", ErrBadInput)
+	}
```
AmanGIT07, line 324: Validations can go in proto. (Caution: check if `rqlQuery` is validated there)

internal/api/v1beta1connect/billing_invoice.go
```diff
@@ -99,6 +99,77 @@ func (h *ConnectHandler) ListInvoices(ctx context.Context, request *connect.Requ
...
+	}
+	if err = rql.ValidateQuery(rqlQuery, invoice.SearchOrgInvoice{}); err != nil {
+		return nil, connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("failed to validate rql query: %v", err))
+	}
+	if rqlQuery.Limit <= 0 {
+		rqlQuery.Limit = utils.DefaultLimit
+	}
+	if rqlQuery.Offset < 0 {
+		rqlQuery.Offset = utils.DefaultOffset
+	}
```
AmanGIT07: check if this happens automatically in utils

internal/store/postgres/billing_invoice_repository.go
```diff
@@ -273,6 +274,83 @@ func (r BillingInvoiceRepository) List(ctx context.Context, flt invoice.Filter)
...
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", invoice.ErrBadInput, err)
+	}
+
+	withSort, err := frontierutils.AddRQLSortInQuery(withSearch, rqlQuery)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", invoice.ErrBadInput, err)
+	}
+
+	countQuery, countParams, err := withSort.Select(goqu.COUNT("*")).ToSQL()
```
AmanGIT07: To get total count, we can use the filter/search applied query

billing/invoice/invoice.go
paanSinghCoder: I think we should keep them separate as they are of different use case - `InvoiceWithOrganization` is global search and `SearchOrgInvoice` is scoped for a single org.

Also, Some fields are not shared across both.

internal/api/v1beta1connect/billing_invoice.go
paanSinghCoder: Replaced the manual check and added utils.AddRQLPaginationInQuery(...)

internal/store/postgres/billing_invoice_repository.go
paanSinghCoder: Done. replaced with `withSearch`

billing/invoice/service.go
```diff
@@ -306,11 +307,24 @@ func (s *Service) ListAll(ctx context.Context, filter Filter) ([]Invoice, error)
 // List currently queries stripe for invoices, but it should be refactored to query our own database
 func (s *Service) List(ctx context.Context, filter Filter) ([]Invoice, error) {
 	if filter.CustomerID == "" {
-		return nil, errors.New("customer id is required")
+		return nil, errors.New("customer id not found")
```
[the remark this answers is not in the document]
AmanGIT07: If this is the method used in the older RPC then we can skip changing the error message. For our newer methods we can make the change

internal/store/postgres/billing_invoice_repository.go
```diff
@@ -273,6 +274,85 @@ func (r BillingInvoiceRepository) List(ctx context.Context, flt invoice.Filter)
...
+		return nil, 0, fmt.Errorf("%w: %w", invoice.ErrBadInput, err)
+	}
+
+	searchColumns := []string{"id", "provider_id", "state", "currency", "hosted_url"}
+	withSearch, err := frontierutils.AddRQLSearchInQuery(withFilters, rqlQuery, searchColumns)
+	if err != nil {
+		return nil, 0, fmt.Errorf("%w: %w", invoice.ErrBadInput, err)
+	}
+
+	withSort, err := frontierutils.AddRQLSortInQuery(withSearch, rqlQuery)
```
AmanGIT07, line 321: Lets add default sort order in case its not specified

internal/store/postgres/billing_invoice_repository.go
```diff
@@ -273,6 +274,85 @@ func (r BillingInvoiceRepository) List(ctx context.Context, flt invoice.Filter)
...
+		"state",
+		"currency",
+		"amount",
+		"hosted_url",
+		"due_at",
+		"effective_at",
+		"created_at",
+		"period_start_at",
+		"period_end_at",
+	}
```
AmanGIT07: Let's put these supported columns at top of the file for better visibility. Same for search etc columns

internal/store/postgres/billing_invoice_repository.go
```diff
@@ -273,6 +274,85 @@ func (r BillingInvoiceRepository) List(ctx context.Context, flt invoice.Filter)
...
+
+	invoices := make([]invoice.Invoice, 0, len(invoiceModels))
+	for _, invoiceModel := range invoiceModels {
+		inv, err := invoiceModel.transform()
+		if err != nil {
+			return nil, 0, err
+		}
+		invoices = append(invoices, inv)
+	}
+	return invoices, totalCount, nil
```
AmanGIT07: why we need to send totalCount separately? It should be embedded in the response itself as per proto definition

AmanGIT07:
@paanSinghCoder Please add some sample request and response

billing/invoice/service.go
paanSinghCoder: customerID and GroupBy are not present on proto level thus makes sense to keep here.

Remove the query check.

internal/store/postgres/billing_invoice_repository.go
paanSinghCoder: Added createAt default sort

internal/store/postgres/billing_invoice_repository.go
paanSinghCoder: Done.

internal/store/postgres/billing_invoice_repository.go
paanSinghCoder: Added struct `SearchOrgInvoicesPagination` and `SearchOrgInvoicesResult`

billing/invoice/invoice.go
```diff
@@ -113,3 +113,29 @@ type Filter struct {
...
+	CreatedAt     time.Time `rql:"name=created_at,type=datetime"`
+	PeriodStartAt time.Time `rql:"name=period_start_at,type=datetime"`
+	PeriodEndAt   time.Time `rql:"name=period_end_at,type=datetime"`
+}
+
+type SearchOrgInvoicesPagination struct {
+	Limit      int
+	Offset     int
+	TotalCount int64
+}
```
AmanGIT07: this can be replaced by `utils.Page`

paanSinghCoder:
Rename `SearchOrgInvoices` to `SearchOrganisationInvoices`

internal/api/v1beta1connect/billing_invoice_test.go
```diff
@@ -525,3 +527,251 @@ func TestConnectHandler_GetUpcomingInvoice(t *testing.T) {
 		})
 	}
 }
+
+func TestConnectHandler_SearchOrgInvoices(t *testing.T) {
```
AmanGIT07: The handler tests miss a few paths:                                                                                                                                                                                       
  - customer.ErrInvalidUUID / customer.ErrInvalidID → should return CodeInvalidArgument
  - Internal service error (non-ErrBadInput error from SearchOrgInvoices) → should return CodeInternal                                                                                                                                   
  - transformInvoiceToPB error path (e.g., invalid metadata)

billing/invoice/invoice.go
paanSinghCoder: Done.

internal/api/v1beta1connect/billing_invoice_test.go
paanSinghCoder: Added the missing tests.

AmanGIT07 (approved):

PR #1537, opened 2026-04-14
feat: add membership package and AddOrganizationMembers RPC

whoAbhishekSah:
## Summary

Introduces `core/membership` package as the foundation for centralizing all member access management, and wires it into the new `AddOrganizationMembers` AdminService RPC. Service and handler PRs were initially split but merged into this single PR.

### `core/membership/` package
- `AddOrganizationMember(ctx, orgID, principalID, principalType, roleID)` — adds a user to an org with an explicit role
- Creates policy + matching explicit relation atomically (owner role → `owner` relation, else → `member` relation)
- Compensating cleanup: if relation creation fails after policy creation, deletes the orphaned policy (logs a warning if cleanup also fails)
- Audit logging via structured records and event auditor

### `AddOrganizationMembers` AdminService RPC
- Batch endpoint accepting list of `{user_id, role_id}` pairs
- Returns per-member `{user_id, role_id, success, error}` for partial failure handling
- Authorization: `IsSuperUser` (AdminService)
- Domain errors pass through as-is; internal errors masked and logged server-side

## Edge cases handled
- Principal type must be `app/user` (rejected upfront — serviceusers are bound to orgs at creation, not added later)
- Org does not exist or is disabled → error (via `orgService.Get`)
- User does not exist or is disabled → error
- Role does not exist → error
- Role is not org-scoped (even if org-specific) → error
- User is already a member → error
- Policy creation fails → error (no orphan)
- Relation creation fails → policy cleaned up; if cleanup fails, logged with IDs

## Design decisions
- `orgRoleToRelation` is a pure function — checks role name, no DB call
- `validateOrgRole` returns the fetched role to avoid redundant lookups
- `validateOrgRole` always checks `OrganizationNamespace` scope first, then decides platform-wide vs org-specific
- `createPolicy` / `createRelation` as separate private helpers (no deletes needed for Add)
- Handler iterates batch entries and maps errors through `toClientError` to mask internals

## Related
- Parent issue: [#1478](https://github.com/raystack/frontier/issues/1478)
- Proto: https://github.com/raystack/proton/pull/469 (merged)

## Test plan
- [x] 14 unit tests covering all edge cases above
- [x] End-to-end tested on local server (see earlier comment)


core/membership/service.go
```diff
@@ -0,0 +1,229 @@
...
+// orgRoleToRelation maps an org role to the corresponding SpiceDB relation name.
+// Owner role gets "owner" relation, everything else gets "member" relation.
+func orgRoleToRelation(r role.Role) string {
+	if r.Name == schema.RoleOrganizationOwner {
+		return schema.OwnerRelationName
+	}
+	return schema.MemberRelationName
+}
+
+// replacePolicy deletes all existing policies for the principal+resource and creates a new one.
```
rohilsurana: we can remove this, it seems out of sync

core/membership/service.go
```diff
@@ -0,0 +1,229 @@
...
+// - a platform-wide role scoped to organizations, or
+// - a custom role created for this specific organization.
+func (s *Service) validateOrgRole(ctx context.Context, roleID, orgID string) (role.Role, error) {
+	fetchedRole, err := s.roleService.Get(ctx, roleID)
+	if err != nil {
+		return role.Role{}, err
+	}
+
+	// custom role belonging to this org
+	if fetchedRole.OrgID == orgID {
```
rohilsurana, line 160: do we need to check scope here as well, what if that role is not meant for org, but project?

core/membership/service.go
whoAbhishekSah: Done — removed the stale comment (e542286).

core/membership/service.go
whoAbhishekSah: Good catch — updated `validateOrgRole` to always check `schema.OrganizationNamespace` is in `Scopes` before deciding if the role is platform-wide or org-specific. Added a test case `should_return_error_if_org-specific_role_has_project_scope` to cover this (e542286).

whoAbhishekSah:
## Test Results

Tested end-to-end on `pr1537-test-org` using the [claude plugin](https://github.com/whoAbhishekSah/claude-marketplace/). All 8 scenarios pass.

| # | Scenario | Result |
|---|---|---|
| 1 | Add multiple members (alice as owner, bob as viewer) | `success: true` for both |
| 2 | User already a member | `error: "principal is already a member of this resource"` |
| 3 | Invalid user ID | `error: "user doesn't exist"` |
| 4 | Invalid role ID | `error: "role doesn't exist"` |
| 5 | Invalid org ID | `error: "org doesn't exist"` |
| 6 | Partial failure (1 valid, 1 duplicate) | charlie added, alice errored — per-entry results correct |
| 7 | Project-scoped role (not org-scoped) | `error: "role is not valid for organization scope"` |
| 8 | Non-super admin authorization | `code: "permission_denied"` |

### Verified
- Batch processing with per-entry success/error
- Validation: org existence, user existence, role existence, role scope
- Duplicate membership rejection
- SuperAdmin (`IsSuperUser`) authorization enforced
- Members (alice, bob, charlie) appear in the org after the RPC call

rohilsurana (approved):

PR #1521, opened 2026-04-09
feat: add membership package and AddOrganizationMembers RPC

whoAbhishekSah:
## Summary
- Introduce `core/membership` package that manages policy + relation together for org membership operations, ensuring they stay in sync
- Add `AddOrganizationMembers` RPC that accepts a list of `{user_id, role_id}` pairs with explicit roles
- Returns per-member `{user_id, success, error}` results for partial failure handling
- Intended to eventually replace `AddOrganizationUsers` which lacks a role parameter and hardcodes the default viewer role

## What's in the membership package
- `AddOrganizationMember` — validates org role, rejects if already a member, creates policy + explicit relation, audits
- Private core: `replacePolicy` (delete old + create new), `replaceRelation` (cleans up both owner/member relations before creating the correct one)
- `validateOrgRole` — ensures role is org-scoped (global or org-specific)

## Wiring
- Handler iterates batch members, calls `membershipService.AddOrganizationMember` per entry
- Authorization: checks `update` permission on the organization
- Proto depends on: https://github.com/raystack/proton/pull/469

## Test plan
- [ ] Add members with explicit roles via AddOrganizationMembers
- [ ] Verify already-member returns error in result without failing batch
- [ ] Verify invalid role returns error in result
- [ ] Verify policy + relation both created for each successful member
- [ ] Verify authorization rejects unauthorized callers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1538, opened 2026-04-14
feat: add AddOrganizationMembers RPC handler

whoAbhishekSah:
## Summary
Wire the membership package (from #1537) into the `AddOrganizationMembers` AdminService RPC.

**Depends on:** #1537 (membership service package)
**Proto depends on:** https://github.com/raystack/proton/pull/469

### Changes
- `cmd/serve.go` — create and inject `membershipService`
- `internal/api/api.go` — add `MembershipService` to `Deps`
- `internal/api/v1beta1connect/` — handler, interface, wiring
- `pkg/server/connect_interceptors/authorization.go` — `IsSuperUser` for AdminService
- `proto/` — regenerated with `AddOrganizationMembers` RPC
- `Makefile` — proton commit updated

### Handler behavior
- Iterates batch members, calls `membershipService.AddOrganizationMember` per entry
- Domain errors (already member, invalid role, user not found, disabled) returned as-is in per-member results
- Internal errors masked with generic message + logged server-side
- Authorization: superadmin only (`IsSuperUser`)

## Test plan
- [x] Tested manually (see PR #1521 test results)
- [ ] Merge #1537 first, then this PR


internal/api/v1beta1connect/organization.go
```diff
@@ -569,6 +570,66 @@ func (h *ConnectHandler) SetOrganizationMemberRole(ctx context.Context, request
...
+		results = append(results, result)
+	}
+
+	return connect.NewResponse(&frontierv1beta1.AddOrganizationMembersResponse{
+		Results: results,
+	}), nil
+}
+
+// isDomainError returns true if the error is a known domain error safe to expose to clients.
+func isDomainError(err error) bool {
```
rohilsurana, line 606: we should probably do this on a global level

rohilsurana (approved):

internal/api/v1beta1connect/organization.go
whoAbhishekSah: I'll make it global in subsequent PRs.

PR #1564, opened 2026-04-22
refactor: move SearchOrganizationTokens to FrontierService

paanSinghCoder:
## Summary
- Move `SearchOrganizationTokens` from `AdminService` to `FrontierService` so org admins and Users both can list their own org's token transactions (previously restricted to platform super-users is `IsSuperUser` check).
- Bump `PROTON_COMMIT` and regenerate `proto/v1beta1`.
- Authorization: `IsSuperUser` → `IsAuthorized(org, UpdatePermission)`. Super-users continue to pass via the interceptor bypass.
- Admin Tokens page + `AddTokensDialog` now use `FrontierServiceQueries.searchOrganizationTokens`. Request/response shape is unchanged.

## Dependency
Depends on raystack/proton#482. After that merges and `@raystack/proton` is published, bump the pin in `web/sdk/package.json` and `web/apps/admin/package.json`.

## Test plan
- [x] `go build ./...` locally
- [ ] CI green after proton#482 + npm publish + sdk version bump
- [ ] Smoke test: hit `FrontierService/SearchOrganizationTokens` as an org admin, a super-user, and an unrelated user — expect success, success, PermissionDenied.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1565, opened 2026-04-22
feat(sdk): add sorting and RQL infinite loading to views-new tokens

paanSinghCoder:
## Summary
- Replace `listBillingTransactions` (non-RQL, client-mode) with `searchOrganizationTokens` (RQL, server-mode, infinite scroll) on the views-new Tokens page.
- DataTable runs in server mode with a Toolbar. Sort enabled on Date, Tokens, Member; filter on Date and Tokens; hide on Events and Member.
- Default sort is `createdAt desc`.
- Events column shows the raw `description` (the old `source` → friendly-label mapping is dropped; `source` isn't in the new RPC response).
- Balance panel, `useTokens`, and `AddTokensDialog` are untouched.

https://github.com/user-attachments/assets/3d56a898-76f0-490d-8461-1be87a7d5417

## Dependency
Depends on #1564 (backend move + admin dashboard swap). Merge #1564 first, then this.

## Test plan
- [x] `pnpm run build` (web turbo) passes
- [ ] Navigate to Tokens → table loads first page, scrolls to load more
- [ ] Sort via Display Settings on Date / Tokens / Member
- [ ] Filter via Toolbar on Date (range) and Tokens (number)
- [ ] Reset to default restores sort + column visibility
- [ ] Add tokens via dialog → table refreshes


web/sdk/admin/views/organizations/details/layout/add-tokens-dialog.tsx
rohanchkrabrty, line 61: This should be `AdminServiceQueries` since its on admin

web/sdk/admin/views/organizations/details/tokens/index.tsx
rohanchkrabrty, line 95: This should be `AdminServiceQueries`

web/sdk/react/views-new/tokens/components/columns.tsx
```diff
@@ -97,35 +92,46 @@ export function getColumns({
...
+      accessorKey: 'userTitle',
+      enableSorting: true,
+      enableHiding: true,
       cell: ({ row, getValue }) => {
-        const userTitle =
-          row?.original?.user?.title || row?.original?.user?.email || '-';
-        const avatarSrc = row?.original?.user?.avatar;
-        const color = getAvatarColor(getValue() as string);
+        const userId = row?.original?.userId || '';
+        const title = (getValue() as string) || '-';
```
rohanchkrabrty, line 125: IMO user email should be used as fallback. It's consistent with other pages

web/sdk/react/views-new/tokens/components/columns.tsx
```diff
@@ -97,35 +92,46 @@ export function getColumns({
...
+        }
         return (
-          <Text
-            size="regular"
-            variant="secondary"
-            style={{ overflow: 'hidden', textOverflow: 'ellipsis', whiteSpace: 'nowrap', display: 'block' }}
-          >
-            {eventName || '-'}
-          </Text>
+          <Tooltip message={eventName} delayDuration={500}>
```
rohanchkrabrty: This is not the correct usage of new Tooltip. Please fix

web/sdk/react/views-new/tokens/tokens-view.tsx
```diff
@@ -36,6 +42,35 @@ import styles from './tokens-view.module.css';
...
+    icon={<ExclamationTriangleIcon />}
+    heading="No Transactions"
+  />
+);
+
+const ErrorState = () => (
+  <EmptyState
+    icon={<ExclamationTriangleIcon />}
+    heading="Error loading transactions"
+    subHeading="Something went wrong. Please try refreshing the page."
```
rohanchkrabrty, line 71: there should be an action button to retry in error state

web/sdk/react/views-new/tokens/tokens-view.tsx
```diff
@@ -173,27 +234,23 @@ export function TokensView() {
...
       >
-        <Flex direction="column" gap={5}>
+        <Flex direction="column" gap={5} style={{ width: '100%' }}>
           <Text size="small" weight="medium">
             Token transactions
           </Text>
-          <DataTable.Filters />
-          <DataTable.VirtualizedContent
+          <DataTable.Toolbar />
+          <DataTable.Content
```
rohanchkrabrty: even with infinite query the table needs to be virtualized since there can be multiple entries

rohanchkrabrty (changes requested):

web/sdk/react/views-new/tokens/tokens-view.tsx
paanSinghCoder: I dont think there is a standard design we follow for the Retry action on error screens. We can check with design in future.

web/sdk/admin/views/organizations/details/tokens/index.tsx
paanSinghCoder: We recently moved this RPC to FrontierService from AdminService so that both admin and end user can consume it. Intentional.

web/sdk/admin/views/organizations/details/layout/add-tokens-dialog.tsx
paanSinghCoder: We recently moved this RPC to FrontierService from AdminService so that both admin and end user can consume it. Intentional.

rohanchkrabrty: Makes sense.

web/sdk/admin/views/organizations/details/tokens/index.tsx
rohanchkrabrty: Makes sense.

web/sdk/react/views-new/tokens/tokens-view.tsx
rohanchkrabrty: Works

web/sdk/react/views-new/tokens/components/columns.tsx
paanSinghCoder: I think we should keep it as it is as Admin tokens page also doesn't have email as fallback. Keeping SDK tokens page in parity with the Admin tokens page makes sense. 

Also, I think it's a product decision. We will add it if the requirement arises from there.

Moreover, `SearchOrganizationTokens` RPC doesn't have `user_email` in the proto so FE doesn't have it to show.

web/sdk/react/views-new/tokens/components/columns.tsx
paanSinghCoder: Done

web/sdk/react/views-new/tokens/tokens-view.tsx
paanSinghCoder: The reason it was removed is the Virtual table owns its scroll area and doesn't overflow so the page doesn't scroll, the top section of the page is kept sticky. Check the table behaviour with Virtualisation enabled. 


https://github.com/user-attachments/assets/814a6ede-dffe-48e9-b2f5-01f629c606f0

paanSinghCoder: Updated this to use VirtualisedContent. As a result only the table will scroll which might be not so good UX. This will be picked up along with other improvements in CLD-2902.

web/sdk/react/views-new/tokens/components/columns.tsx
rohanchkrabrty: We should use user_email instead of user_id

paanSinghCoder: `SearchOrganizationTokens` doesn't have email field in proto. Also email field is not there in the Admin page (there is a mismatch in admin and sdk) thus removed it for now from sdk. 

As mentioned, this discrepancy will have to be taken up with product and fixed. Already noted down.

rohanchkrabrty (approved):

PR #1572, opened 2026-04-24
refactor(sdk): migrate service account project access to membership API

whoAbhishekSah:
## Summary
- Replace raw `CreatePolicyForProject` / `DeletePolicy` / `ListPolicies` calls with `SetProjectMemberRole` and `RemoveProjectMember` in the `views-new` service accounts dialogs
- Role changes are now atomic (single `SetProjectMemberRole` call instead of non-atomic delete+create cycle)
- Aligns service account project access management with how user/group project access is handled everywhere else in the SDK

## Test plan
- [x] Create a new service account with projects selected — project access granted
- [x] Assign service account to a project — access granted
- [x] Change role on an assigned project — role updated atomically
- [x] Remove service account from a project — access removed
- [x] Delete service account token — works as expected

### DIFF
--- a/web/sdk/react/views-new/service-accounts/components/add-service-account-dialog.tsx
+++ b/web/sdk/react/views-new/service-accounts/components/add-service-account-dialog.tsx
@@ -16,14 +16,13 @@ import {
 import {
   FrontierServiceQueries,
   CreateServiceUserRequestSchema,
-  CreatePolicyForProjectRequestSchema,
+  SetProjectMemberRoleRequestSchema,
   CreateServiceUserTokenRequestSchema,
   ListOrganizationServiceUsersRequestSchema,
   ListOrganizationProjectsRequestSchema,
   ListServiceUserTokensRequestSchema,
   ListServiceUserTokensResponseSchema,
-  ServiceUserRequestBodySchema,
-  CreatePolicyForProjectBodySchema
+  ServiceUserRequestBodySchema
 } from '@raystack/proton/frontier';
 import {
   Button,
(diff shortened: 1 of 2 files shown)


web/sdk/react/views-new/service-accounts/components/add-service-account-dialog.tsx
```diff
@@ -108,12 +108,28 @@ export function AddServiceAccountDialog({
     return orderBy(list, ['title'], ['asc']);
   }, [projectsData]);
 
+  const { data: rolesData } = useQuery(
```
rsbh: need to handle loading and error state

whoAbhishekSah: Done. Added loading state for the roles query — skeletons show while roles load, and the submit button stays disabled until roles are available.

web/sdk/react/views-new/service-accounts/components/add-service-account-dialog.tsx
```diff
@@ -108,12 +108,28 @@ export function AddServiceAccountDialog({
...
+    create(ListRolesRequestSchema, {
+      state: 'enabled',
+      scopes: [PERMISSIONS.ProjectNamespace]
+    }),
+    { enabled: Boolean(orgId) }
+  );
+
+  const ownerRoleId = useMemo(
+    () =>
+      rolesData?.roles?.find(r => r.name === PERMISSIONS.RoleProjectOwner)
```
rohilsurana, line 122: checking permission with role? lets add a todo that we need to change this to permission check

rsbh (approved):

web/sdk/react/views-new/service-accounts/components/add-service-account-dialog.tsx
rohilsurana: resolving this comment, role is fine here, but we probably need to make this a config.

PR #1575, opened 2026-04-26
fix: billing and plans pages

rohanchkrabrty:
## Summary

- Audit `views-new/billing` permissions to match the old flow: gate the page behind a "Restricted Access" `EmptyState` when the user lacks the canSeeBilling permission, and route billing-detail edits to the Stripe customer portal instead of an in-app dialog.
- Polish loading/empty states across the billing page — remove skeletons from PaymentIssue/UpcomingPlanChangeBanner, add two-line text-shaped skeletons in PaymentMethodCard/BillingDetailsCard, fix the UpcomingBillingCycle loading flicker, and add `loading`/`loaderText` to the action buttons during checkout.
- Update the billing transactions table to match Figma: title row + sort control, simplified columns, capped action column width, and accent-styled View invoice button.
- Fix `views-new/plans` monetary formatting by removing `valueInMinorUnits={false}` so the Apsara `Amount` component correctly treats backend cents as minor units, and restyle the interval tabs to medium size via a CSS module override.
- Default the plan card's interval tab to the user's current subscription synchronously (replacing the buggy post-render `useEffect`), and render skeleton rows in `FeatureTable` while `listFeatures` is loading instead of leaving whitespace.


web/sdk/react/views-new/billing/components/invoices.tsx
```diff
@@ -58,77 +60,82 @@ const getColumns = ({
...
+    {
+      header: '',
+      accessorKey: 'invoiceLink',
+      enableSorting: false,
+      classNames: {
+        cell: styles.linkColumn
+      },
+      styles: {
+        cell: { width: '120px', maxWidth: '120px' },
+        header: { width: '120px', maxWidth: '120px' }
```
paanSinghCoder, line 119: Lets use `--rs-space-17` here.

paanSinghCoder (approved):

web/sdk/react/views-new/billing/components/invoices.tsx
rohanchkrabrty: Let's skip rs variables in table cell width. Doesn't make sense

Issue #1585, opened 2026-04-30
Leaky permissions: principal lifecycle events do not consistently revoke access


AmanGIT07:
## Summary                                                                                                                                                                                                                       
                                                                                                                                                                                                                                   
  Frontier supports several principal types (`User`, `ServiceUser`, `PAT`, `Session`) and several lifecycle events that should result in revoked access (`Delete`, `Disable`, expiry). The matrix below shows where revocation     
  works correctly today and where it does not. Multiple cells fail — credentials continue to authenticate after the operation that was supposed to stop them.
                                                                                                                                                                                                                                   
  This issue is a **tracker**. The actual work is split into per-fix sub-issues to keep PRs reviewable. Most of the bug surface collapses into a small number of changes (an auth-layer gate plus a deleter cleanup); the rest are 
  hygiene/defense-in-depth follow-ups.
                                                                                                                                                                                                                                   
  ## Current behavior matrix                                                                                                                                                                                                     

  For each row: *does the principal still authenticate after this event?*                                                                                                                                                          
   
  | Event | User session | PAT | ServiceUser |                                                                                                                                                                                     
  |---|---|---|---|                                                                                                                                                                                                              
  | User deleted | No (FK CASCADE on `sessions.user_id`) | No (FK CASCADE on `user_pats.user_id`) | n/a |                                                                                                                          
  | User disabled | No (`UserRepository.GetByID` filters `notDisabledUserExp`) | No (same user filter) | n/a |                                                                                                                     
  | Org deleted | n/a (sessions not org-scoped) | No (FK CASCADE on `user_pats.org_id`) | **Yes — bug** |                                                                                                                          
  | Org disabled | n/a | **Yes — bug** | **Yes — bug** |                                                                                                                                                                           
                                                                                                                                                                                                                                   
  The user-disabled column is correctly gated by a data-layer filter (`notDisabledUserExp` in `internal/store/postgres/user_repository.go:50`). It's a useful design template for what the org-state gate should look like.        
                                                                                                                                                                                                                                   
  ## Sub-issues                                                                                                                                                                                                                    
                                                                                                                                                                                                                                 
  ### Bugs (active leaks)                                                                                                                                                                                                          
  - [ ] #TBD — Auth path does not check parent org state (PAT + ServiceUser)
  - [ ] #TBD — DeleteOrganization leaves service users orphaned                                                                                                                                                                    
                                                                                                                                                                                                                                 
  ### Future enhancements                                                                                                                                                                                                          
[...]

PR #1595, opened 2026-05-11
feat(user): revoke sessions on user disable

AmanGIT07:
## Summary

`user.Service.Disable` now soft-deletes all of the user's active sessions after flipping state. Outstanding session cookies stop authenticating immediately instead of lingering until `expires_at`. Not a security fix — the existing `notDisabledUserExp` filter at the user repo already blocks disabled-user auth at lookup time — but a hygiene and pre-disable sessions can no longer silently resume if the user is re-enabled within the validity window. They will need to re-login.

Refs umbrella issue #1585.

## Changes

- `core/authenticate/session/service.go` — new `Service.DeleteByUserID` that reuses `repo.List` + `repo.Delete`. 
- `core/user/service.go` — new `SessionService` interface and field; `Disable` calls `DeleteByUserID` after a successful `SetState`. Invalid-id and not-found paths skip the revocation as before.
- `cmd/serve.go` — pass the existing `sessionService` into `user.NewService`.
- Mock: `core/user/mocks/session_service.go` (regenerated via mockery; `core/user` is already in `.mockery.yaml`).

## Technical Details

The DRY choice: `DeleteByUserID` lists the user's active sessions via the existing repo method and calls `repo.Delete` per session. Each delete goes through the existing TX + `createSessionRevokeAuditRecord` path, which already emits a `SessionRevokedEvent` when the actor is different from the session owner (the admin-initiated disable case). 

Mirrors what `serviceUserService.Delete` already does for credentials.

## Test Plan

- [x] `go test ./core/user/... ./core/authenticate/session/...` passes
- [x] `go build ./...` clean
- [x] New unit tests: 4 `TestService_DeleteByUserID` subtests (happy, empty, list-error, delete-error); `TestService_Disable` updated with revoke-on-success, skip-on-failure, and revoke-error-propagation cases
- [ ] Manual: disable a user with an active session cookie, verify subsequent requests with that cookie return Unauthenticated; verify `sessions.deleted_at` is set


core/user/service.go
```diff
@@ -139,7 +146,10 @@ func (s Service) Disable(ctx context.Context, id string) error {
 	if !utils.IsValidUUID(id) {
 		return ErrInvalidID
 	}
-	return s.repository.SetState(ctx, id, Disabled)
+	if err := s.repository.SetState(ctx, id, Disabled); err != nil {
+		return err
+	}
+	return s.sessionService.DeleteByUserID(ctx, id)
```
rohilsurana, line 152: Lets do the opposite, first we delete sessions and then set state, the reason being these are outside transaction as of now, hence if we set state and the app crashes we can have sessions left out, it would be better to remove the sessions even if the state is not disabled.

rohilsurana (approved):

core/user/service.go
AmanGIT07: The state flip is what actually blocks auth (via `notDisabledUserExp` at the user-repo lookup); session revocation is hygiene on top. Comparing failure modes:                
 - Current order, second step fails: user is disabled, leftover sessions can't authenticate anyway.
 - Proposed order, second step fails: some or all sessions are gone but state is still enabled — and the disable they tried to perform never happened.                                        

So I'd keep `SetState` first.

PR #1605, opened 2026-05-13
refactor(admin-ui): replace always-open Drawer pattern with controlled open state (also fixes mount animation)

paanSinghCoder:
## Why
v1's `Drawer` uses CSS transitions on `data-starting-style` — they only fire on a `false → true` open state change. Drawers mounted with `open={true}` literal (v0 pattern carried over) skip the entry animation because base-ui's `useTransitionStatus` initializes `mounted=true` and never applies `data-starting-style`.

## What
Lift the `open` state to each drawer's parent: child accepts `open?: boolean`, parent always renders the child and passes `open={someState}`. Open/close now naturally fires base-ui's enter/exit transitions.

## Sites updated
- `views/organizations/list/create.tsx` (New Organization)
- `views/organizations/details/edit/{kyc,organization,billing}.tsx` (Edit panels)
- `views/products/details.tsx` (Product details)
- `views/webhooks/webhooks/{create,update}/index.tsx` (Create/Update webhook)

## Caveat
Form/UI state in these components now persists across close/reopen (component stays mounted). Acceptable for read-only views; if reset-on-open is wanted for the create forms, follow up with a `useEffect(reset, [open])` or `key` reset.

## Base
This PR targets `chore/sdk-admin-apsara-v1` because the fix only applies in the v1 context. Will rebase / auto-retarget to `main` once the migration PR merges.

### DIFF
--- a/web/sdk/admin/views/organizations/details/edit/billing.tsx
+++ b/web/sdk/admin/views/organizations/details/edit/billing.tsx
@@ -26,6 +26,7 @@ import {
 import { create } from "@bufbuild/protobuf";
 
 interface EditBillingPanelProps {
+  open?: boolean;
   onClose: () => void;
 }
 
@@ -70,7 +71,7 @@ const billingDetailsUpdateSchema = z
 
 type BillingDetailsForm = z.infer<typeof billingDetailsUpdateSchema>;
 
-export function EditBillingPanel({ onClose }: EditBillingPanelProps) {
+export function EditBillingPanel({ open = false, onClose }: EditBillingPanelProps) {
   const { 
     billingAccount,
     billingAccountDetails,
@@ -174,7 +175,7 @@ export function EditBillingPanel({ onClose }: EditBillingPanelProps) {
   const isPrepaid = tokenPaymentType === "prepaid";
 
   return (
-    <Drawer open>
+    <Drawer open={open}>
       <Drawer.Content showCloseButton={false} className={styles["drawer-content"]}>
         <SidePanel
           data-test-id="edit-billing-panel"
--- a/web/sdk/admin/views/organizations/details/edit/billing.tsx
+++ b/web/sdk/admin/views/organizations/details/edit/billing.tsx
@@ -175,7 +175,7 @@ export function EditBillingPanel({ open = false, onClose }: EditBillingPanelProp
   const isPrepaid = tokenPaymentType === "prepaid";
 
   return (
-    <Drawer open={open}>
+    <Drawer open={open} onOpenChange={(open) => !open && onClose()}>
       <Drawer.Content showCloseButton={false} className={styles["drawer-content"]}>
         <SidePanel
           data-test-id="edit-billing-panel"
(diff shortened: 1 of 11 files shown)


web/sdk/admin/views/organizations/list/create.tsx
```diff
@@ -127,7 +129,7 @@ export function CreateOrganizationPanel({
   const showOtherTypeField = watch("type", "other") === "other";
 
   return (
-    <Drawer open>
+    <Drawer open={open}>
```
rohanchkrabrty: onClose should also be passed to Drawer root right?

web/sdk/admin/views/products/details.tsx
```diff
@@ -4,20 +4,26 @@ import styles from "./products.module.css";
...
 
 export default function ProductDetails({
   product,
+  open = false,
   onClose,
   onNavigateToPrices,
 }: ProductDetailsProps) {
   return (
-    <Drawer open>
+    <Drawer open={open}>
```
rohanchkrabrty: Same here

web/sdk/admin/views/webhooks/webhooks/create/index.tsx
```diff
@@ -73,7 +74,7 @@ export default function CreateWebhooks({ onClose: onCloseProp }: CreateWebhooksP
   };
 
   return (
-    <Drawer open={true}>
+    <Drawer open={open}>
```
rohanchkrabrty: Same here

web/sdk/admin/views/webhooks/webhooks/update/index.tsx
```diff
@@ -100,7 +101,7 @@ export default function UpdateWebhooks({ webhookId: webhookIdProp, onClose: onCl
   }, [webhook, methods.reset]);
 
   return (
-    <Drawer open={true}>
+    <Drawer open={open}>
```
rohanchkrabrty: Same here

web/sdk/admin/views/organizations/details/edit/billing.tsx
```diff
@@ -174,7 +175,7 @@ export function EditBillingPanel({ onClose }: EditBillingPanelProps) {
   const isPrepaid = tokenPaymentType === "prepaid";
 
   return (
-    <Drawer open>
+    <Drawer open={open}>
```
rohanchkrabrty: Same here

rohanchkrabrty (changes requested):

web/sdk/admin/views/organizations/list/create.tsx
paanSinghCoder: In the previous version the page never relied on overlay-click-to-close, the cross button always closed it. 

Added the onClose to add the the overlay click to close as it is a standard behaviour.

web/sdk/admin/views/products/details.tsx
paanSinghCoder: Done

web/sdk/admin/views/webhooks/webhooks/create/index.tsx
paanSinghCoder: Done

web/sdk/admin/views/webhooks/webhooks/update/index.tsx
paanSinghCoder: Done

web/sdk/admin/views/organizations/details/edit/billing.tsx
paanSinghCoder: Done.

rohanchkrabrty (approved):

PR #1606, opened 2026-05-13
feat(admin-ui): unify page header height and surface icon prop on Views

paanSinghCoder:
Two small follow-ups to the Apsara v1 migration:

## 1. Header height parity
- Added \`min-height: var(--rs-space-12)\` (56px) to the four list-page navbars (orgs/users/audit-logs/invoices) and to \`<PageHeader>\` itself.
- Roles/Products/Plans/Webhooks already rendered ≥56px due to their Search/Button content; this is a floor.
- Preferences/Admins previously rendered ~52px (no children); now match 56px.

## 2. Icon prop on Views
- Added \`icon?: ReactNode\` to \`RolesView\`, \`ProductsView\`, \`PlansView\`, \`WebhooksView\`, \`PreferencesView\`, \`AdminsView\`.
- Each View threads it through to its header / \`<PageHeader>\` and renders next to the title.
- Each host page imports its sidebar SVG and passes it down, so sidebar and page-header share the same icon source.

## Base
Targets \`chore/sdk-admin-apsara-v1\` (the migration PR). Auto-retargets to \`main\` once that merges.


web/sdk/admin/views/plans/header.tsx
```diff
@@ -1,11 +1,13 @@
+import type { ReactNode } from "react";
 import { DataTable } from "@raystack/apsara-v1";
 import { PageHeader } from "../../components/PageHeader";
 import styles from "./plans.module.css";
 
-export const PlanHeader = ({ header }: any) => {
+export const PlanHeader = ({ header, icon }: { header: any; icon?: ReactNode }) => {
```
rohanchkrabrty: This component can be skipped and the code can be inlined

web/sdk/admin/views/products/header.tsx
```diff
@@ -13,14 +14,17 @@ const defaultPageHeader = {
 export const ProductsHeader = ({
```
rohanchkrabrty: This component can also be skipped and inlined instead

web/sdk/admin/views/roles/header.tsx
```diff
@@ -7,10 +8,11 @@ const pageHeader = {
   breadcrumb: [] as { name: string; href?: string }[],
 };
 
-export const RolesHeader = () => {
+export const RolesHeader = ({ icon }: { icon?: ReactNode }) => {
```
rohanchkrabrty: This component can also be skipped and inlined instead

web/sdk/admin/views/webhooks/webhooks/header.tsx
```diff
@@ -12,14 +13,16 @@ const pageHeader = {
 export type WebhooksHeaderProps = {
   header?: typeof pageHeader;
   onOpenCreate?: () => void;
+  icon?: ReactNode;
 };
 
-export const WebhooksHeader = ({ header = pageHeader, onOpenCreate }: WebhooksHeaderProps) => {
+export const WebhooksHeader = ({ header = pageHeader, onOpenCreate, icon }: WebhooksHeaderProps) => {
```
rohanchkrabrty: Same for this

rohanchkrabrty (changes requested):
IMO we can create a `BaseViewProps` and then extend all view/extend types from it

web/sdk/admin/views/plans/header.tsx
paanSinghCoder: Done.

paanSinghCoder:
> IMO we can create a `BaseViewProps` and then extend all view/extend types from it

@rohanchkrabrty There is only one prop (icon) shared across these components. I don't think creating a base type adds much value here.

web/sdk/admin/views/webhooks/webhooks/header.tsx
paanSinghCoder: Done

web/sdk/admin/views/products/header.tsx
paanSinghCoder: Done

web/sdk/admin/views/roles/header.tsx
paanSinghCoder: Done.

web/sdk/admin/views/products/header.tsx
rohanchkrabrty: This is not resolved @paanSinghCoder

rohanchkrabrty (approved):

web/sdk/admin/views/products/header.tsx
paanSinghCoder: Missed this in previous commit. Removed.

PR #1615, opened 2026-05-14
refactor(group): move group-delete cleanup into membership package

whoAbhishekSah:
Mirrors `project`'s deletion pattern (handler → deleter orchestrator → `DeleteModel` + membership cleanup) and centralises every SpiceDB write performed during a group's lifetime in one package.

## What we create vs. what we delete (service-level audit)

A group's lifetime touches SpiceDB through exactly two service primitives. From that abstraction, every write must be matched by an explicit delete on the same primitive.

### `relation.Service` writes (outside `policy.AssignRole`)

| Created by | Tuple | Deleted by |
|---|---|---|
| `linkGroupToOrg` | `group#org@organization` | `unlinkGroupFromOrg` |
| `linkGroupToOrg` | `organization#member@group#member` | `unlinkGroupFromOrg` |
| `SetGroupMemberRole` (add path) | `group#owner@user` / `group#member@user` | `RemoveAllGroupMembers` → `removeRelations` |

### `policy.Service` writes

| Created by | Policy shape | Deleted by |
|---|---|---|
| `SetGroupMemberRole` | `resource=group, principal=user` | `RemoveAllGroupMembers` → `policyService.Delete` |
| External callers | `principal=group, resource=<other>` (group granted role elsewhere) | `removeGroupAsPrincipalPolicies` → `policyService.Delete` |

Everything created via the group/membership code path is matched 1:1 with an explicit delete via the same primitive. No wildcard sweeps, no defensive cleanup.

## Orchestration

`core/deleter.DeleteGroup(id)` is the new orchestrator:

1. `membership.OnGroupDeleted(id)` — tears down all SpiceDB state above. Symmetric counterpart to `OnGroupCreated`.
2. `group.Service.DeleteModel(id)` — drops the entity row + emits audit. Pure entity cleanup, no SpiceDB writes.

`RemoveAllGroupMembers` uses a two-pass approach: delete every policy first, track which principals had a failure, then clean up direct relations only for principals whose policies were all deleted. Prevents stripping a principal's relations while a surviving policy still references them; a retry can finish the job.

## Known leak (separate, broader)

`policy.AssignRole` writes three SpiceDB tuples but `policy.Delete` only cleans two of them — the resource→rolebinding grant tuple leaks on every policy removal across the codebase (org / project / group member removal all hit it). The historical wildcard `Delete(Object: <resource>)` sweep in `group.DeleteModel` / `project.DeleteModel` was a workaround that only covered the same-resource case. Tracked separately in #1617 — not in scope here.

## Other pieces

- `DeleteGroup` handler routes through `deleterService` instead of calling `groupService` directly.
- `group.RemoveUsers` / `removeUsers` stay — they serve the single-user cascade in `deleter.DeleteUser` (user is being destroyed system-wide), same dual-track as org.

## Test plan

- [x] `go build ./...`
- [x] `go test ./core/membership/... ./core/group/... ./core/deleter/... ./internal/api/v1beta1connect/...`
- New unit tests for `RemoveAllGroupMembers` (principal-dedupe, partial-failure leaves relations for the failed principal intact) and `OnGroupDeleted` (full three-way cleanup, group-not-found). Existing deleter cascade test updated.


whoAbhishekSah:
## Manual Testing Results

Tested group deletion scenarios with fresh org `pr1615-034ec2`:

| Test | Description | Result |
|------|-------------|--------|
| **1** | User loses project access when their only group is deleted | ✅ PASS |
| **2** | User retains project access if they have direct access | ✅ PASS |
| **3** | User retains project access via other groups | ✅ PASS |
| **4** | Org membership preserved after group deletion | ✅ PASS |
| **5** | No orphaned policies for deleted groups | ✅ PASS |
| **6** | No stale group refs in project user lists | ✅ PASS |
| **7** | System functional after deletions | ✅ PASS |

### Setup
- **group-alpha**: alice, bob → viewer on project-one & project-two
- **group-beta**: charlie, dave, alice → manager on project-two
- **bob**: also direct viewer on project-one (edge case)

### Verified Behaviors
1. Deleting group-alpha: alice lost project-one access, bob retained it (direct), bob lost project-two access
2. Deleting group-beta: alice/charlie/dave lost project-two access
3. All users retained org membership after both groups deleted
4. No orphaned policies or stale references found

rohilsurana (approved):

PR #1618, opened 2026-05-15
feat(membership): add ListResourcesByPrincipal with inheritance

AmanGIT07:
## Summary

Adds `membership.Service.ListResourcesByPrincipal(principal, resourceType, filter)` — a policy-driven replacement for today's `org.ListByUser`, `project.ListByUser`, `group.ListByUser` methods that read SpiceDB relations. Purely additive: no callers migrated in this PR.

## Why

SpiceDB is built for permission checks, not listing. Postgres already has the policies and roles needed to answer "what can this principal see," so the listing path moves off `relationService.LookupResources` and onto direct policy reads. Making Postgres policies the single source of truth for listing simplifies the model and decouples a heavily-used UI path from SpiceDB.

Companion to #1616 (legacy service-user policy backfill) — both need to land before any caller is migrated.

## Algorithm

`ListResourcesByPrincipal` runs the underlying algorithm twice when a PAT is present (once for the user, once for the PAT-as-principal) and intersects, so the PAT can narrow but never widen the user's visibility.

For each resource type:

- **Organization**: list policies, take resource IDs. Not role-permission-gated — matches today's relation-based `org.membership` check.
- **Group**: same as org. No inheritance.
- **Project**: unions three sources, deduped, then narrowed by `filter.OrgID`:
  1. **Direct project policies** filtered by `schema.ProjectDirectVisibilityPerms` (mirrors today's `LookupResources(project, ..., project.get)`).
  2. **Group expansion** — project policies on the principal's groups, gated the same way. Runs even with `NonInherited=true`.
  3. **Org inheritance** (skipped if `NonInherited`) — every project in any org whose policy role grants `schema.OrganizationProjectInheritPerms`. Batched via `project.Filter.OrgIDs` to avoid N+1.

## Schema-derived permission lists

Two constants live in `internal/bootstrap/schema/schema.go`:

```go
ProjectDirectVisibilityPerms = []string{
    "app_project_administer", "app_project_get", "app_project_update",
}
OrganizationProjectInheritPerms = []string{
    "app_organization_administer", "app_project_get", "app_project_administer",
}
```

They mirror the `granted->` / `pat_granted->` arrows of `app/project.get` and `app/organization.project_get` in `base_schema.zed`. `inheritance_perms_test.go` is a regex-only drift guard: it scans the embedded schema source and asserts the constants match what's in the schema. It also rejects non-Union expressions (intersection/exclusion would silently break the any-of semantics in `filterByRolePermissions`). No SpiceDB compiler import, no AST walking.

## Files changed 

- `internal/bootstrap/schema/schema.go` — `ProjectDirectVisibilityPerms` and `OrganizationProjectInheritPerms` constants
- `internal/bootstrap/schema/inheritance_perms_test.go` — regex drift guard for both constants
- `core/membership/service.go` — `ResourceFilter`, `ListResourcesByPrincipal` + per-type helpers + `filterByRolePermissions`
- `core/membership/service_test.go` — 16 new table-driven cases (rejection of unsupported type, direct-only listing per type, inheritance for owner/manager/viewer/custom roles, NonInherited, group expansion, OrgID narrowing, ServiceUser principal, deduplication, PAT all-projects, no-PAT short-circuit, stale-relation regression, PAT narrowing)
- `core/project/filter.go` + `internal/store/postgres/project_repository.go` — `OrgIDs []string` plural filter for batched org-inheritance expansion

## Test plan

- [x] `make build` passes
- [x] `make lint` (0 issues)
- [x] `make test -race` on touched packages green
- [x] **Algorithm trace against real local DB**: verified `listOrgInheritedProjectIDs` produces correct results for actual users on the dev DB:
  - Org Owner on a real org (PixxelSpace, 175 projects) → 175 inherited project IDs
  - Org Viewer on the same org → 0 projects (correctly excluded by role-permission filter)
  - Org Owner on an empty org → 0 projects (early-return fast path)

## Not in this PR

- **No callers migrated.** `org.ListByUser`, `project.ListByUser`, `group.ListByUser` still in production; new method is unreachable from the API surface today. Migration of handlers (`ListOrganizationsByUser`, `ListUserGroups`, `ListProjectsByUser`, etc.) and the deletion of old `ListByUser` methods come in follow-up PRs.
- **No end-to-end runtime test possible yet.** End-to-end coverage lands with the handler-migration PRs.


core/project/filter.go
```diff
@@ -10,4 +10,9 @@ type Filter struct {
 	// NonInherited filters out projects that are inherited from access given through an organization
 	NonInherited bool
 	Pagination   *pagination.Pagination
+
+	// OrgIDs narrows results to projects whose org_id is in this list. Used by
+	// membership listing to batch-expand all projects across the orgs a
+	// principal can inherit project visibility from.
+	OrgIDs []string
```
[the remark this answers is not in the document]
AmanGIT07: Doc clarified in the latest push (e2ab44d7): the comment now spells out the AND-intersection semantics explicitly — "If both OrgID and OrgIDs are set, projects must satisfy both (intersection) — typically yields no rows unless OrgID is one of OrgIDs."

Kept the AND semantics rather than mutually-excluding the fields because (a) no caller in this PR sets both, and (b) introducing cross-field validation here would be inconsistent with the rest of `project.Filter`, which has no such guards on its other field combinations. The doc makes the future-caller hazard explicit without adding repo-layer plumbing.

internal/store/postgres/project_repository.go
```diff
@@ -161,6 +161,11 @@ func (r ProjectRepository) List(ctx context.Context, flt project.Filter) ([]proj
 			"org_id": flt.OrgID,
 		})
 	}
+	if len(flt.OrgIDs) > 0 {
+		stmt = stmt.Where(goqu.Ex{
+			"org_id": goqu.Op{"in": flt.OrgIDs},
+		})
+	}
```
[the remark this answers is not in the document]
AmanGIT07: Addressed via doc clarification in the latest push (e2ab44d7), see core/project/filter.go:14-18. The comment on `OrgIDs` now spells out the AND-intersection behaviour explicitly.

Opted for documentation over a repo-level validation guard because (a) no caller in this PR sets both fields simultaneously, and (b) adding cross-field validation here would be inconsistent with the rest of `project.Filter`, which has no other guards on field combinations. If a future caller needs different semantics, that's a separate change with explicit intent.

core/membership/service.go
```diff
@@ -1465,3 +1483,294 @@ func (s *Service) auditGroupMemberRemoved(ctx context.Context, grp group.Group,
...
+	case schema.ProjectNamespace:
+		return s.listProjectsForPrincipal(ctx, principalID, principalType, filter)
+	default:
+		return nil, ErrInvalidResourceType
+	}
+}
+
+// listOrgsForPrincipal returns org IDs where the principal has any policy.
+// Not role-permission-gated; mirrors today's relation-based org.membership
+// check (member + owner).
```
whoAbhishekSah: These look confusing/AI bloat. 
We are not checking membership via the relation but only via policy. Worth removing this comment.

core/membership/service.go
```diff
@@ -1562,3 +1563,300 @@ func (s *Service) auditGroupMemberRemoved(ctx context.Context, grp group.Group,
...
+	ids := make([]string, 0, len(policies))
+	for _, pol := range policies {
+		ids = append(ids, pol.ResourceID)
+	}
+	return utils.Deduplicate(ids), nil
+}
+
+// listGroupsForPrincipal returns group IDs where the principal has any policy.
+// Not role-permission-gated; mirrors today's group.membership (member + owner).
+// No inheritance branch — group.membership has no org-> chain today.
```
whoAbhishekSah: We can remove these comments. Otherwise, let's write it in simpler English: "Not role-permission-gated" doesn't imply much.

whoAbhishekSah: Better is to write the function's logic as comment so that anyone can understand it at a glance.

core/membership/service.go
```diff
@@ -1562,3 +1563,300 @@ func (s *Service) auditGroupMemberRemoved(ctx context.Context, grp group.Group,
...
+		}
+	}
+	return ids, nil
+}
+
+// listDirectProjectIDs returns project IDs from the principal's direct
+// project policies whose role grants any schema.ProjectDirectVisibilityPerms.
+// Mirrors today's LookupResources(project, ..., project.get) gating.
+func (s *Service) listDirectProjectIDs(ctx context.Context, principalID, principalType string) ([]string, error) {
+	policies, err := s.policyService.List(ctx, policy.Filter{
```
whoAbhishekSah, line 1722: Is there a way for us to pass the role filter here instead of `s.filterByRolePermissions(ctx, policies, schema.ProjectDirectVisibilityPerms)`? If we can get the role IDs that have `ProjectDirectVisibilityPerms`, then we can pass them to the Policy filter itself. 

We can also add permission-based filtering in Policy Service to make this easier. It can be taken up as a separate PR and refactoring if you prefer.

whoAbhishekSah: Since this pattern is also used in Groups and Org, I recommend making this change.

whoAbhishekSah (approved):

core/membership/service.go
AmanGIT07: These comments were trying to flag the deliberate asymmetry: project listing gates by role permissions (mirrors SpiceDB's project.get chain, which has granted->app_project_*  arrows), while org and group listing don't (mirror *.membership which is just member + owner — a relation check, no role involved). The intent was to make it obvious that "having any policy = access" is the rule for orgs/groups but not for projects — that asymmetry is hard to infer from function body alone.

AmanGIT07: But I understand these shortened terms may be confusing for someone who is reviewing since they might not have all context as the developer would have.

core/membership/service.go
AmanGIT07: 1. ▎ Since this pattern is also used in Groups and Org, I recommend making this change.                                                                                                                                                   
Quick clarification — filterByRolePermissions is only called from the three project-listing branches (direct, group-expanded, org-inheritance). Org and group listings don't use it  (any policy = access, mirroring today's *.membership = member + owner in the schema). So the refactor benefit lands on project listing. Did you mean "refactoring for the three project sites"?                                                                                                
                                                                                                                                                                                                                                       
2.   On the refactor itself, confirming if I understood that right — you want to push the role-permission filter into the policy layer so membership doesn't do the two-step (list policies → fetch roles → filter) inline?
Two ways to wire it up; both share the same DB mechanism (Postgres JSONB ?| does array intersection — roles.permissions is JSONB):                                                                             
 
   **Option 1** — Prefetch role IDs, use existing policy.Filter.RoleIDs                                                                                                                                                                       
Add Permissions []string to role.Filter. The role repo translates it to:                                                                                                                                                               
                                                                                                                                                                                                                                       
```  SELECT id FROM roles WHERE permissions ?| ARRAY['app_project_get','app_project_administer',...] ```
                                                                                                                                                                                                                                         
  Caller does two calls:                                                                                                                                                                                                                 
  ```                                                                                                                                                                                                                                       
  roles, _ := s.roleService.List(ctx, role.Filter{Permissions: schema.OrgProjectInheritPerms})                                                                                                                                           
  roleIDs := utils.Map(roles, func(r role.Role) string { return r.ID })                                                                                                                                                                
                                                                                                                                                                                                                                         
  policies, _ := s.policyService.List(ctx, policy.Filter{
      PrincipalID:   userID,                                                                                                                                                                                                             
      PrincipalType: schema.UserPrincipal,                                                                                                                                                                                               
      ResourceType:  schema.OrganizationNamespace,
      RoleIDs:       roleIDs,                                                                                                                                                                                                            
  })                                                                                                                                                                                                                                   
```
  Two indexed round-trips. Reuses existing policy.Filter.RoleIDs. Role and policy repos stay isolated — same pattern we used for project.Filter.OrgIDs earlier in this PR.                                                               
   
  **Option 2** — SQL JOIN inside policy repo                                                                                                                                                                                                 
                                                                                                                                                                                                                                       
  Add RolePermissions []string to policy.Filter. Policy repo joins against roles:                                                                                                                                                        
```   
  SELECT p.* FROM policies p                                                                                                                                                                                                             
  JOIN roles r ON r.id = p.role_id                                                                                                                                                                                                     
  WHERE p.principal_id = ? AND p.resource_type = ?
    AND r.permissions ?| ARRAY['app_project_get','app_project_administer',...]                                                                                                                                                           
   ```
  One round-trip, simpler caller:                                                                                                                                                                                                        
                                                                                                                                                                                                                                       
 ```
 policies, _ := s.policyService.List(ctx, policy.Filter{                                                                                                                                                                                
      PrincipalID:     userID,                                                                                                                                                                                                         
      PrincipalType:   schema.UserPrincipal,
      ResourceType:    schema.OrganizationNamespace,                                                                                                                                                                                     
      RolePermissions: schema.OrgProjectInheritPerms,
  })                                                                                                                                                                                                                                     
  ```                                                                                                                                                                                                                              
  Policy repo now knows about the roles table schema.

File: core/membership/service.go [hunk 6]
```diff
@@ -1562,3 +1563,300 @@ func (s *Service) auditGroupMemberRemoved(ctx context.Context, grp group.Group,
...
+		PrincipalID:   principalID,
+		PrincipalType: principalType,
+		ResourceType:  schema.OrganizationNamespace,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("list org policies for inheritance: %w", err)
+	}
+	inheritingOrgIDs, err := s.filterByRolePermissions(ctx, policies, schema.OrganizationProjectInheritPerms)
+	if err != nil {
```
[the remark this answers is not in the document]
AmanGIT07: Thanks for flagging this. The safety here comes from `core/policy/service.go:64-67` — `policy.Service.Create` rejects any policy with `grant_relation = pat_granted` unless `principal_type = app/pat`. So in practice:

- **User pass** (principal = `app/user`): `policyService.List` only returns rows with `grant_relation = granted` — pat_granted rows can't exist for user principals. The wider list is functionally equivalent to the narrower `granted->` subset here.
- **PAT pass** (principal = `app/pat`): can see both `granted` (per-project PAT scopes) and `pat_granted` (all-projects PAT scopes) rows. The wider list is correct here.

The single-list approach holds as long as the `pat_granted ⇒ PAT-only` invariant in policy.Service.Create holds. The comment on `filterByRolePermissions` flags this dependency — if a future schema change makes the two arrow kinds grant *different* permission sets, the gating logic here needs to branch on `grant_relation`. Today they don't diverge, so the union list is safe.

internal/bootstrap/schema/schema.go
```diff
@@ -97,6 +97,25 @@ var (
...
+
+// OrganizationProjectInheritPerms — role permissions that, on an org-level
+// policy, grant the principal visibility into every project in that org.
+// Mirrors the granted-> and pat_granted-> arrows of
+// app/organization.project_get in base_schema.zed.
+var OrganizationProjectInheritPerms = []string{
+	"app_organization_administer",
+	"app_project_get",
+	"app_project_administer",
+}
```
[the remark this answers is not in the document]
AmanGIT07: Acknowledged — these are technically mutable. The codebase already has the same convention for adjacent authz state (e.g. `schema.PredefinedRoles` is also a mutable exported slice). Keeping the surface consistent rather than one-off hardening here. If we want strict freezing across the package, makes sense as a separate pass applying a uniform pattern (predicate functions, frozen accessor with copy, etc.) across all such vars.

core/membership/service.go
AmanGIT07: updated

core/membership/service.go
AmanGIT07: updated

core/membership/service.go
whoAbhishekSah: I like option 1

PR #1620, opened 2026-05-15
fix(api): replace raw error responses with generic ErrInternalServerError

rohilsurana:
## Summary
- 21 call sites across 6 handler files were returning raw internal errors to clients via `connect.NewError(connect.CodeInternal, err)`, leaking DB table names, OIDC config, SMTP details, SpiceDB addresses, and file paths
- Replaced all with `connect.NewError(connect.CodeInternal, ErrInternalServerError)` which returns a generic `"internal server error"` message
- All call sites already had `errorLogger.LogUnexpectedError` / `errorLogger.LogServiceError` calls logging the real error server-side before returning

## Files changed
- `authenticate.go` — 10 fixes (unauthenticated endpoints, highest risk)
- `authorize.go` — 2 fixes
- `kyc.go` — 3 fixes
- `organization.go` — 2 fixes
- `role.go` — 2 fixes
- `serviceuser.go` — 2 fixes

## Test plan
- [x] `go build` and `go vet` pass
- [x] Zero remaining `connect.NewError(connect.CodeInternal, err)` in handler files
- [ ] Verify error responses return generic message, not raw error details

### DIFF
--- a/internal/api/v1beta1connect/authenticate.go
+++ b/internal/api/v1beta1connect/authenticate.go
@@ -296,6 +296,7 @@ func (h *ConnectHandler) getLoggedInSessionID(ctx context.Context) (uuid.UUID, e
 }
 
 func (h *ConnectHandler) GetLoggedInPrincipal(ctx context.Context, via ...authenticate.ClientAssertion) (authenticate.Principal, error) {
+	errorLogger := NewErrorLogger()
 	principal, err := h.authnService.GetPrincipal(ctx, via...)
 	if err != nil {
 		switch {
@@ -309,7 +310,6 @@ func (h *ConnectHandler) GetLoggedInPrincipal(ctx context.Context, via ...authen
 			errors.Is(err, patErrors.ErrDisabled):
 			return principal, connect.NewError(connect.CodeUnauthenticated, ErrUnauthenticated)
 		default:
-			errorLogger := NewErrorLogger()
 			errorLogger.LogUnexpectedError(ctx, nil, "GetLoggedInPrincipal", err)
 			return principal, connect.NewError(connect.CodeInternal, ErrInternalServerError)
 		}
--- a/internal/api/v1beta1connect/authenticate.go
+++ b/internal/api/v1beta1connect/authenticate.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"log/slog"
 	"strings"
 	"time"
 
@@ -310,7 +309,8 @@ func (h *ConnectHandler) GetLoggedInPrincipal(ctx context.Context, via ...authen
 			errors.Is(err, patErrors.ErrDisabled):
 			return principal, connect.NewError(connect.CodeUnauthenticated, ErrUnauthenticated)
 		default:
-			slog.ErrorContext(ctx, "unexpected error in GetLoggedInPrincipal", "error", err)
+			errorLogger := NewErrorLogger()
+			errorLogger.LogUnexpectedError(ctx, nil, "GetLoggedInPrincipal", err)
 			return principal, connect.NewError(connect.CodeInternal, ErrInternalServerError)
 		}
 	}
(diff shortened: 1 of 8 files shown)


internal/api/v1beta1connect/authenticate.go
```diff
@@ -309,7 +309,7 @@ func (h *ConnectHandler) GetLoggedInPrincipal(ctx context.Context, via ...authen
 			errors.Is(err, patErrors.ErrDisabled):
 			return principal, connect.NewError(connect.CodeUnauthenticated, ErrUnauthenticated)
 		default:
-			return principal, connect.NewError(connect.CodeInternal, err)
+			return principal, connect.NewError(connect.CodeInternal, ErrInternalServerError)
```
AmanGIT07, line 312: Lets print the err value like we have for others.

rohilsurana: done

AmanGIT07:
One leak still slips past at `internal/api/v1beta1connect/authenticate.go:47`:

```go
return nil, status.Error(codes.Internal, err.Error())
```

Same unauthenticated `Authenticate()` surface as the other 10 fixes — but this path uses `google.golang.org/grpc/status` instead of `connect.NewError`, so the grep used to find leaks doesn't catch it. `err.Error()` is still passed through to the client (and `errors.Is(err, frontiersession.ErrNoSession)` has already been ruled out above, so this can be any session-extraction error including decode/crypto failures).

Suggest:
```go
return nil, connect.NewError(connect.CodeInternal, ErrInternalServerError)
```

Worth widening the audit query: ``git grep -nE 'status\.(Error|Errorf)\(codes\.Internal,' internal/api/v1beta1connect/`` — returns only this one hit in the package, so it should be a small follow-up.

internal/api/v1beta1connect/authenticate.go
```diff
@@ -196,7 +196,7 @@ func (h *ConnectHandler) AuthToken(ctx context.Context, request *connect.Request
 			errorLogger.LogUnexpectedError(ctx, request, "AuthToken", err,
 				"org_id", orgId,
 				"service_user_id", principal.ServiceUser.ID)
-			return nil, connect.NewError(connect.CodeInternal, err)
+			return nil, connect.NewError(connect.CodeInternal, ErrInternalServerError)
```
AmanGIT07, line 197: Suggestion (separate from the sanitization fix, but related): switch on the error type here instead of returning `CodeInternal` for everything.

`organization.ErrDisabled` is a domain condition, not an internal error — the same service user with the same permissions would succeed once the org is re-enabled. Right now it gets the same `500 / "internal server error"` as a Postgres connection drop, which loses the only signal the caller has to self-diagnose.

The codebase already has three different mappings for this same sentinel:

| Site | Code |
| --- | --- |
| `domain.go`, `invitations.go`, `group.go` (8 sites) | `CodeNotFound` + `ErrOrgDisabled` |
| `serviceuser.go:158` | `CodeFailedPrecondition` + `ErrOrgDisabled` |
| here | `CodeInternal` (catch-all) |

`CodeFailedPrecondition` (matching `serviceuser.go`) is the most accurate — gRPC API guidelines explicitly prefer it over `PermissionDenied` when "some external state must change first". The enumeration argument that probably motivates `CodeNotFound` elsewhere doesn't apply here either, since the caller has already authenticated as a service user of that org and clearly knows it exists.

Suggested shape:
```go
_, err := h.orgService.Get(ctx, orgId)
if err != nil {
    errorLogger.LogServiceError(ctx, request, "AuthToken.orgService.Get", err,
        "org_id", orgId,
        "service_user_id", principal.ServiceUser.ID)
    if errors.Is(err, organization.ErrDisabled) {
        return nil, connect.NewError(connect.CodeFailedPrecondition, ErrOrgDisabled)
    }
    return nil, connect.NewError(connect.CodeInternal, ErrInternalServerError)
}
```

`ErrOrgDisabled` is already defined in `errors.go:25`. Note the existing test `TestConnectHandler_AuthToken_ServiceUser/should return error when service user org is disabled` would need its assertion updated to match. Happy to split into a follow-up if out of scope.

AmanGIT07 (approved):

PR #1624, opened 2026-05-19
fix(authz): symmetric SpiceDB cleanup on policy delete

### LINKED ISSUE #1617: policy.Delete is asymmetric with AssignRole: resource-grant tuple leaks on every policy delete
## Problem

`policy.AssignRole` writes **three** SpiceDB tuples per policy, but `policy.Delete` / `policy.DeleteWithMinRoleGuard` only clean up **two** of them. Every call to delete a policy leaks the third tuple.

### `policy.Create` writes

1. **Policy row** via `repository.Upsert` — INSERT ON CONFLICT UPDATE on `(role_id, resource_id, resource_type, principal_id, principal_type)`. Returns same ID on conflict and updates `grant_relation`, `metadata`, `updated_at`.
2. **Three SpiceDB tuples** via `AssignRole`:

   ```
   T1: rolebinding:<polID>#role_bearer @ <principal_type>:<principal_id>[#member]
   T2: rolebinding:<polID>#role        @ role:<role_id>
   T3: <resource_type>:<resource_id>#<grant_relation> @ rolebinding:<polID>
   ```

T1, T2 have the rolebinding as **Object**. T3 has it as **Subject** (the resource is the Object).

### `policy.Delete` deletes

1. `relationService.Delete(Relation{Object: rolebinding:<polID>})` — wildcard by Object: catches T1, T2; **misses T3**.
2. `repository.Delete(polID)` — drops the policy row.

`DeleteWithMinRoleGuard` has the same gap.

### A1 — T3 always leaks (primary, real)

T3 has the rolebinding as **Subject**, not Object, so the wildcard-by-Object delete never matches it. Every removal path leaks T3:

- `membership.RemoveOrganizationMember` / `RemoveProjectMember` / `RemoveGroupMember`
- `membership.RemoveAllGroupMembers`, `removeGroupAsPrincipalPolicies`
- `project.UpdateOwner`, `project.RemovePrincipalPolicies`

Today `group.DeleteModel` and `project.DeleteModel` paper over it with a wildcard `Delete(Object: <resource>)` sweep at resource-delete time — but:
1. The sweep only catches T3 where the **deleted resource** is the object. T3 on *other* resources (e.g. group-as-principal on a project) still leaks.
2. Per-member removals (not full resource delete) leak T3 forever.

### A2 — Upsert replay can leave a stale T3 (theoretical)

`Create` calls `repository.Upsert`. On conflict (same unique key) it updates `grant_relation` and returns the SAME `polID`. `AssignRole` then re-writes T1/T2/T3 with the NEW `grant_relation`. SpiceDB writes use `OPERATION_TOUCH` (idempotent for identical tuples), so T1 and T2 are fine. But the OLD T3 — `<resource>#<old_grant>@rolebinding:<polID>` — is never deleted: a new tuple at `<resource>#<new_grant>@rolebinding:<polID>` is written alongside it.

Not currently exploitable in practice — `grant_relation` is one of two fixed values (`granted` / `pat_granted`) tied to `principal_type`, which is part of the unique key, so the Upsert branch can never observe a `grant_relation` change. Worth fixing if `grant_relation` ever becomes meaningfully variable.

## Proposed fix (A1)

Make `Delete` symmetric with `AssignRole` — also remove T3 by deleting where the rolebinding is the Subject:

```go
func (s Service) Delete(ctx context.Context, id string) error {
    // T1, T2: rolebinding as Object
    if err := s.relationService.Delete(ctx, relation.Relation{
        Object: relation.Object{ID: id, Namespace: schema.RoleBindingNamespace},
    }); err != nil { return err }

    // T3: rolebinding as Subject (resource-grant tuple)
    if err := s.relationService.Delete(ctx, relation.Relation{
        Subject: relation.Subject{ID: id, Namespace: schema.RoleBindingNamespace},
    }); err != nil && !errors.Is(err, relation.ErrNotExist) { return err }

    return s.repository.Delete(ctx, id)
}
```

Same change in `DeleteWithMinRoleGuard`.

Once landed, the wildcard sweeps in `group.DeleteModel` / `project.DeleteModel` become dead code and can be removed in a followup.

whoAbhishekSah:
## Problem

`policy.AssignRole` writes three SpiceDB tuples per policy:

```
T1: rolebinding:<polID>#role_bearer          @ <principal>
T2: rolebinding:<polID>#role                 @ role:<roleID>
T3: <resource>#<grant_relation>              @ rolebinding:<polID>
```

`policy.Delete` (and `DeleteWithMinRoleGuard`) only swept by Object on the rolebinding, catching T1+T2. T3 has the rolebinding as **Subject**, so it leaked on every per-member removal, ownership swap, and cross-resource principal cleanup.

Full-resource teardown (`group.DeleteModel`, `project.DeleteModel`) papered over it with a wildcard sweep on the resource — but only for T3 tied to the deleted resource. Per-member removals leaked forever.

## Fix

Both delete paths now route through a `deleteRoleBindingRelations` helper that issues two `relationService.Delete` calls: one by Object (T1, T2), one by Subject (T3). The Subject-side delete tolerates `relation.ErrNotExist` so resource-level sweeps that already removed T3 don't fail the policy delete.

## Test plan

- [x] `core/policy` unit tests cover happy path, `ErrNotExist` tolerance, non-`ErrNotExist` failure semantics for both `Delete` and `DeleteWithMinRoleGuard`.
- [x] Caller-side tests pass: `organization`, `group`, `project`, `membership`, `userpat`, `deleter`, `v1beta1connect` policy handler.

## Follow-ups

- Remove the now-redundant T3 sweeps in `group.DeleteModel` and `project.DeleteModel`.
- #1625 — derive `grant_relation` from `principal_type` inside `policy.Service.Create` so the A2 invariant is structural.

Closes #1617.

### DIFF
--- a/core/policy/service.go
+++ b/core/policy/service.go
@@ -2,6 +2,7 @@ package policy
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/raystack/frontier/pkg/utils"
@@ -78,12 +79,7 @@ func (s Service) Create(ctx context.Context, policy Policy) (Policy, error) {
 }
 
 func (s Service) Delete(ctx context.Context, id string) error {
-	if err := s.relationService.Delete(ctx, relation.Relation{
-		Object: relation.Object{
-			ID:        id,
-			Namespace: schema.RoleBindingNamespace,
-		},
-	}); err != nil {
+	if err := s.deleteRoleBindingRelations(ctx, id); err != nil {
 		return err
 	}
 	return s.repository.Delete(ctx, id)
@@ -93,12 +89,31 @@ func (s Service) DeleteWithMinRoleGuard(ctx context.Context, id string, guardRol
 	if err := s.repository.DeleteWithMinRoleGuard(ctx, id, guardRoleID); err != nil {
 		return err
 	}
-	return s.relationService.Delete(ctx, relation.Relation{
+	return s.deleteRoleBindingRelations(ctx, id)
+}
+
+// deleteRoleBindingRelations removes all SpiceDB tuples written by AssignRole:
+// the rolebinding-as-Object tuples (role_bearer, role) and the resource-grant
+// tuple where the rolebinding is the Subject. The Subject-side delete tolerates
+// ErrNotExist because resource-delete sweeps may have already removed it.
+func (s Service) deleteRoleBindingRelations(ctx context.Context, id string) error {
+	if err := s.relationService.Delete(ctx, relation.Relation{
 		Object: relation.Object{
 			ID:        id,
 			Namespace: schema.RoleBindingNamespace,
 		},
-	})
+	}); err != nil {
+		return err
+	}
+	if err := s.relationService.Delete(ctx, relation.Relation{
+		Subject: relation.Subject{
+			ID:        id,
+			Namespace: schema.RoleBindingNamespace,
+		},
+	}); err != nil && !errors.Is(err, relation.ErrNotExist) {
+		return err
+	}
+	return nil
 }
 
 // AssignRole Note: ideally this should be in a single transaction
--- a/core/policy/service_test.go
+++ b/core/policy/service_test.go
@@ -30,7 +30,7 @@ func TestService_Delete(t *testing.T) {
 		setup   func() *policy.Service
 	}{
 		{
-			name:    "delete policy from relation before repository",
+			name:    "delete both rolebinding-as-object and rolebinding-as-subject tuples before repository",
 			id:      "test-id",
 			wantErr: false,
 			setup: func() *policy.Service {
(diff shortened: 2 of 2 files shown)


whoAbhishekSah:
## Manual verification — direct SpiceDB tuple inspection

Tested against a local Frontier + SpiceDB (Postgres backend). For each case I snapshot the `relation_tuple` table before and after the operation and confirm that **all three** tuples written by `AssignRole` (T1 bearer, T2 role, T3 resource-grant) are gone — i.e. no T3 leak. Permission checks are also exercised end-to-end.

| # | Scenario                                                                  | Tuples after delete | Permission after delete | Result |
|---|---------------------------------------------------------------------------|---------------------|-------------------------|--------|
| 1 | `DeletePolicy` — user principal, org resource                             | 0                   | denied                  | pass   |
| 2 | `DeletePolicy` — user principal, project resource                         | 0                   | denied                  | pass   |
| 3 | Recreate identical `(role, resource, principal)` after delete             | clean 3-tuple set, new polID | restored                | pass   |
| 4 | Idempotent delete — `DeletePolicy` called twice on the same id            | 0 after first call  | n/a                     | pass (second call returns `not_found` from repo layer, relations side is clean) |
| 5 | `RemoveOrganizationMember` — per-member removal path                      | 0                   | denied                  | pass   |
| 6 | Cross-resource: `app/group` principal on `app/project` — `DeletePolicy`   | 0                   | n/a                     | pass   |
| 7 | Cross-resource: `app/group` principal on `app/project` — `DeleteGroup`    | 0                   | n/a                     | pass   |
| 8 | `CheckResourcePermission` round-trip (false → true → false → true)        | n/a                 | correct at every step   | pass   |
| 9 | Upsert idempotency — same `(role, resource, principal)` created twice     | same polID, 3 tuples| n/a                     | pass   |
| 10| Service-user principal on org — `DeletePolicy`                            | 0                   | n/a                     | pass   |

Each check was done by SQL against the SpiceDB `relation_tuple` table, e.g.:

```sql
SELECT namespace, object_id, relation, userset_namespace, userset_object_id
FROM relation_tuple
WHERE deleted_xid = '9223372036854775807'::xid8
  AND ((namespace = 'app/rolebinding' AND object_id = '<polID>')
    OR (userset_namespace = 'app/rolebinding' AND userset_object_id = '<polID>'));
```

Cases 5 and 7 are the ones that previously leaked T3 forever (per-member removal and group-as-principal-on-other-resource) — both clean now.

### Note on the `relation.ErrNotExist` tolerance

The Subject-side delete's `errors.Is(err, relation.ErrNotExist)` branch is currently defensive — `relation.Service.Delete` returns nil for empty matches today, so the path isn't reachable from normal flows. I'd keep it for forward-compat: it costs nothing and protects against future `relation.Delete` implementations that surface `ErrNotExist`.

### Out of scope (filed separately)

While testing I found an unrelated leak: `app/serviceuser:<su>#org@app/organization:<org>` (the SU↔org identity-link tuple) leaks on `DeleteServiceUser` when the SU has no remaining org policies at delete time. This is not a rolebinding tuple, so not in scope for this PR — filed as #1629.

rohilsurana (approved):

PR #1630, opened 2026-05-20
fix(serviceuser): symmetric SpiceDB cleanup on delete

### LINKED ISSUE #1629: bug(serviceuser): SU→org identity-link relation tuple leaks on DeleteServiceUser when no org policies remain
## Summary

`serviceuser.Service.Delete` leaks the `app/serviceuser:<su>#org@app/organization:<org>` "identity link" relation tuple in SpiceDB whenever the service user has **no remaining org policies** at delete time.

Concretely: if all policies with the SU as principal have been removed before `DeleteServiceUser` is called, the SU row is dropped from postgres and the SU is gone from the API surface, but the identity-link tuple remains live in SpiceDB forever.

## Repro

```bash
# 1. Fresh org and SU
ORG=$(create_org)
SU=$(create_serviceuser org=$ORG)

# 2. Create a policy granting SU some org role
POL=$(create_policy role=org_viewer resource=app/organization:$ORG principal=app/serviceuser:$SU)

# 3. Delete the policy BEFORE deleting the SU
delete_policy $POL

# 4. Delete the SU
delete_serviceuser $SU

# 5. Query SpiceDB
SELECT * FROM relation_tuple
WHERE deleted_xid = '9223372036854775807'::xid8
  AND (object_id = '<SU>' OR userset_object_id = '<SU>');
-- Expected: 0 rows
-- Actual:   app/serviceuser:<SU>#org@app/organization:<ORG>
```

If step 3 is skipped (i.e. there's at least one SU-principal org policy at delete time), no leak — the auto-cascade path picks up the identity link.

## Root cause

`serviceuser.Service.Delete` (`core/serviceuser/service.go:142`) calls `membershipService.RemoveOrganizationMember` "best-effort" and logs/swallows errors. `RemoveOrganizationMember` (`core/membership/service.go:253`) returns `ErrNotMember` early when the principal has no org policies (`core/membership/service.go:273-275`), so it **never reaches** `cascadeRemovePrincipal`, which is where the SU↔org identity-link cleanup actually lives (`core/membership/service.go:397-407`).

The follow-up `relationService.Delete(Subject={SU})` in `serviceuser.Service.Delete` only cleans tuples where the SU is the **subject** of the relation. The identity-link tuple has the SU as the **object**, so it's missed.

## Why this matters

- **Permission correctness**: not affected today — the SU row is gone, so any check involving it fails on the principal lookup anyway.
- **Hygiene**: SpiceDB accumulates garbage tuples that reference dead service users. Counter and listing operations on `app/serviceuser` see stale rows. Over time this is harder to reason about and harder to back-test relation invariants against.
- **Symmetry with #1617 / #1624**: same class of bug (asymmetric write/delete in the SpiceDB relation layer). Worth fixing the same way — make the SU delete path symmetric with SU creation.

## Suggested fix

In `serviceuser.Service.Delete`, after the existing Subject-side wildcard, also issue an Object-side wildcard for the SU:

```go
// existing: relations where SU is the Subject
if err := s.relationService.Delete(ctx, relation.Relation{
    Subject: relation.Subject{ID: id, Namespace: schema.ServiceUserPrincipal},
}); err != nil {
    return err
}

// new: relations where SU is the Object (e.g. SU#org@org)
if err := s.relationService.Delete(ctx, relation.Relation{
    Object: relation.Object{ID: id, Namespace: schema.ServiceUserPrincipal},
}); err != nil && !errors.Is(err, relation.ErrNotExist) {
    return err
}
```

Alternative: have `RemoveOrganizationMember` clean the identity-link unconditionally (move the block out from under the `cascadeRemovePrincipal` early-return), but this is broader in scope.

## Test plan

- Unit test in `core/serviceuser` covering the "no remaining policies" path.
- Integration / manual: same repro above; assert post-delete tuple count for the SU is 0.

whoAbhishekSah:
## Problem

`serviceuser.Service.Delete` leaks the `app/serviceuser:<su>#org@app/organization:<org>` identity-link tuple when the SU has no remaining org policies at delete time.

Chain:
1. `Delete` calls `membership.RemoveOrganizationMember` best-effort.
2. `RemoveOrganizationMember` returns `ErrNotMember` early when the SU has zero org policies — never reaches `cascadeRemovePrincipal`, where the identity-link cleanup lives.
3. The follow-up `relationService.Delete(Subject={SU})` only sweeps tuples where the SU is the Subject. The identity link has the SU as the Object — missed.

Result: tuple lives in SpiceDB forever, even though the SU row is gone.

## Fix

Add a second `relationService.Delete` after the existing Subject-side one, keyed by Object on the SU. Tolerates `relation.ErrNotExist` for forward-compat. Same shape as #1624.

## Tests

New `core/serviceuser/service_test.go` covers:
- happy path (both sweeps fire)
- membership failure swallowed; both sweeps still fire (the leak repro)
- `ErrNotExist` from Object-side sweep is tolerated
- non-`ErrNotExist` Object-side failure blocks repo delete
- Subject-side failure short-circuits before Object sweep

Also added a `core/serviceuser` mockery entry and the generated mocks.

Closes #1629.


whoAbhishekSah:
## Manual verification — direct SpiceDB tuple inspection

Tested against a local Frontier (PR1630 build) with SpiceDB on Postgres. For each case I snapshot the `relation_tuple` table before/after the operation and confirm the SU↔org identity-link tuple no longer leaks. Postgres rows for `serviceusers` / `serviceuser_credentials` / `policies` also verified.

| # | Scenario | Result |
|---|---|---|
| 1 | Smoke — exact #1629 repro (create SU → create policy → delete policy → delete SU) | pass — 0 tuples reference the SU after delete |
| 2a | `DeleteServiceUser` — bare SU (never had explicit policies) | pass — 0 leaked tuples |
| 2b | `DeleteServiceUser` — SU has live org + project policies at delete time | pass — 0 SU-side tuples; postgres `policies` rows for SU = 0 |
| 2c | `DeleteServiceUser` — SU has both a key credential and an opaque token | pass — credentials rows → 0, tuples → 0 |
| 3 | `DeleteOrganization` cascade with 3 SUs in the org: bare / live-policy / policy-then-deleted (the leaky shape) | pass — 0 leaked tuples for any of the SUs; 0 tuples referencing org or its project |
| 4 | Cross-resource — SU as principal on a project, then `DeleteServiceUser` | partial — SU↔org identity link clean (PR1630's scope); see note |
| 5 | Idempotency — `DeleteServiceUser` twice, then on a non-existent UUID | pass — second call returns `not_found`, no side effects |
| 6 | Regression — SU auth (Basic `<tokenID>:<secret>`) works while alive, hard-fails after delete | pass |

The SQL probe used at every checkpoint:

```sql
SELECT namespace, object_id, relation, userset_namespace, userset_object_id
FROM relation_tuple
WHERE deleted_xid = '9223372036854775807'::xid8
  AND (object_id = '<SU>' OR userset_object_id = '<SU>');
```

rohilsurana (approved):

PR #1655, opened 2026-05-27
fix(sdk): add filter config to old tokens view

rohilsurana:
## Summary
- Add `enableColumnFilter`, `filterType`, and `filterOptions` to the old tokens view columns (used by `OrganizationProfile`)
- Add `DataTable.Filters` trigger button to the transactions table
- Add missing `system.overdraft` event source to `TxnEventSourceMap`
- Fixes: token transaction filters on hue.pixxel.dev were showing 0 results because the Event column had no filter configuration — the DataTable auto-generated filter options from raw `source` values but couldn't match them during client-side filtering

**Related:** The billing page "Upgrade" button not working on hue.pixxel.dev is a studio-side issue — `BillingView` accepts `onNavigateToPlans` but the studio app wasn't passing it. Fixed in pixxelhq/studio#2665.

## Test plan
- [ ] Open tokens page in OrganizationProfile → verify Filter button appears
- [ ] Click Filter → Event → select "Recharge" → verify table shows only recharge transactions
- [ ] Click Filter → Date → verify date range filter works
- [ ] Clear filters → verify all transactions show again

### DIFF
--- a/web/sdk/react/views/tokens/transactions/columns.tsx
+++ b/web/sdk/react/views/tokens/transactions/columns.tsx
@@ -19,12 +19,17 @@ import dayjs from 'dayjs';
 interface getColumnsOptions {
   dateFormat: string;
 }
-const TxnEventSourceMap = {
+const TxnEventSourceMap: Record<string, string> = {
   'system.starter': 'Starter tokens',
   'system.buy': 'Recharge',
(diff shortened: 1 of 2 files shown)


rohilsurana:
Closing - the old tokens view is no longer used, the new views-new/ tokens view already has proper filter config.

PR #1657, opened 2026-05-27
fix(api): dedupe access_pairs and map invalid-id/unknown-perm to 4xx

### LINKED ISSUE #1647: Project-listing handlers: duplicated access_pairs and 500s on invalid input
## Summary

Three pre-existing issues in the project-listing handlers surfaced while testing the `ListByUser` → `List(Filter{Principal})` migration. They're independent of the migration but live on the same handlers and have a consistent shape: invalid input maps to `Internal` instead of `InvalidArgument`, and the `with_permissions` access-pair output is malformed.

1. **`access_pairs` is duplicated when more than one permission is requested.** The handlers emit one entry per `CheckPair` instead of grouping per resource — `N` projects × `M` permissions → `N*M` entries, each project repeated `M` times.
2. **Unknown permission name returns 5xx/404** instead of an empty `access_pairs` block alongside the projects list. The two RPCs even disagree on the code (`Internal` vs `NotFound`).
3. **Malformed / empty `id` returns `Internal`** instead of `InvalidArgument` on `ListProjectsByUser` and `ListServiceUserProjects`.

### Reproduction
Authenticate as a user with at least one accessible project, then:

```bash
curl -s -X POST "http://localhost:8002/raystack.frontier.v1beta1.FrontierService/ListProjectsByCurrentUser" \
  -H "content-type: application/json" -H "connect-protocol-version: 1" \
  -H "Cookie: sid=<sid>" \
  -d '{"orgId":"<org>","withPermissions":["get","delete"]}'
```

For 3 visible projects the response carries 6 `access_pairs` entries — each `project_id` appears twice, each entry redundantly listing `["get","delete"]`.

### Root cause
`internal/api/v1beta1connect/user.go:949-964` (`ListProjectsByCurrentUser`) and `internal/api/v1beta1connect/serviceuser.go:501-516` (`ListServiceUserProjects`) iterate over every `CheckPair` returned by `fetchAccessPairsOnResource`, then for each one re-filter the full pair slice by `resID` and append an `AccessPair{ProjectId: resID, Permissions: <all perms for that resID>}`. With `M` permissions, every resource is visited `M` times and a fully populated `AccessPair` is emitted on each visit.

### Suggested fix
Group `successCheckPairs` by `Object.ID` once, then emit a single `AccessPair` per group:

```go
byResource := map[string][]string{}
order := []string{}
for _, cp := range successCheckPairs {
    rid := cp.Relation.Object.ID
    if _, ok := byResource[rid]; !ok {
        order = append(order, rid)
    }
    byResource[rid] = append(byResource[rid], cp.Relation.RelationName)
}
for _, rid := range order {
    accessPairsPb = append(accessPairsPb, &frontierv1beta1.ListProjectsByCurrentUserResponse_AccessPair{
        ProjectId:   rid,
        Permissions: byResource[rid],
    })
}
```

Same shape for the `ListServiceUserProjectsResponse_AccessPair` type.

### Reproduction
```bash
curl -s -X POST ".../ListProjectsByCurrentUser" \
  -H "Cookie: sid=<sid>" \
  -d '{"orgId":"<org>","withPermissions":["bogus_perm"]}'
# {"code":"internal","message":"internal server error"}

curl -s -X POST ".../ListServiceUserProjects" \
  -H "Cookie: sid=<sa-sid>" \
  -d '{"id":"<su>","orgId":"<org>","withPermissions":["bogus_perm"]}'
# {"code":"not_found","message":"not found"}
```

### Root cause
`fetchAccessPairsOnResource` at `internal/api/v1beta1connect/permission_check.go:96-112` calls `getPermissionName` per `(resource, permission)` pair. `getPermissionName` at `permission_check.go:30-46` translates `permission.ErrNotExist` into `connect.CodeNotFound`.

- `ListServiceUserProjects` at `serviceuser.go:499` does `return nil, err` — the `NotFound` propagates unchanged.
- `ListProjectsByCurrentUser` at `user.go:947` wraps everything into `connect.CodeInternal` — so the same condition surfaces as `internal`.

Neither response is right. The caller asked: "for each of these projects, which of these permissions does the principal hold?" The answer for an unknown permission is "none", not "the request blew up."

### Suggested fix
Filter unknown permission names out inside `fetchAccessPairsOnResource` (or earlier in the handler) and return an empty `access_pairs` set for them rather than erroring. Concretely: treat `permission.ErrNotExist` as "drop this permission" instead of `return nil, err`. Keep the `Internal` path only for genuine repository failures.

Optional: if the desire is to be strict and reject unknown perms, do so explicitly with `InvalidArgument` + a message naming the offending permission — and align both handlers on the same code.

### Reproduction
```bash
# ListProjectsByUser, garbage uuid
curl -s -X POST ".../ListProjectsByUser" -H "Cookie: sid=<admin-sid>" \
  -d '{"id":"not-a-uuid"}'
# {"code":"internal","message":"internal server error"}

# ListServiceUserProjects, empty id but valid org_id
curl -s -X POST ".../ListServiceUserProjects" -H "Cookie: sid=<admin-sid>" \
  -d '{"id":"","orgId":"<org>"}'
# {"code":"internal","message":"internal server error"}
```

### Root cause
- `ListProjectsByUser` handler (`internal/api/v1beta1connect/user.go:864-892`) maps `user.ErrInvalidUUID` to `InvalidArgument`, but `project.Service.List` no longer surfaces that error. With the new `Filter{Principal: {ID: userID}}` path, the malformed UUID flows down to `policyService.List` and fails at the SQL layer as a generic error → the handler's `default` branch returns `CodeInternal`.
- `ListServiceUserProjects` empty `id` is rejected by the authorization interceptor at `pkg/server/connect_interceptors/authorization.go:274-286`, which calls `GetServiceUser` with the empty id. That returns an error which is propagated as `Internal` rather than `InvalidArgument`.

`id` has no `min_len`/`uuid` validation in the proto (`ListProjectsByUserRequest.id`, `ListServiceUserProjectsRequest.id`). The handler is the only line of defence and it doesn't validate before the call chain.

### Suggested fix
Three options, any one of which is enough:
1. Add `(buf.validate.field).string.uuid = true` (and `REQUIRED`) on `id` in both request protos. This makes the response shape correct without any handler change.
2. Validate `id` at the top of each handler with `utils.IsValidUUID(id)` and return `CodeInvalidArgument` early.
3. In `project.Service.List`, when `Filter.Principal != nil`, reject empty or non-UUID `Principal.ID` with a typed error the handlers can map.

Option 1 is the smallest change and makes the behaviour consistent across both RPCs.

## Scope notes
- All three issues are independent of the `ListByUser` migration. They became more visible because that change consolidated the inheritance + group + direct paths onto the same handler surface.

AmanGIT07:
## Summary

Three independent bugs in the project / service-user / current-user-groups listing handlers, and a bundled fix for the same `access_pairs` duplication in `ListCurrentUserGroups`.

Fixes #1647.

### 1. `access_pairs` duplicated per resource

`ListProjectsByCurrentUser`, `ListServiceUserProjects`, and `ListCurrentUserGroups` emitted one `AccessPair` per `(resource, permission)` `CheckPair` and inside each iteration filtered the full slice for the same resource id. With M permissions per resource that produced M copies of the same row.

Fix: group `successCheckPairs` by resource id once and emit one entry per resource.

### 2. Unknown permission name returned 5xx / 404

`fetchAccessPairsOnResource` called `getPermissionName` per `(resource, permission)` pair, which mapped `permission.ErrNotExist` to `CodeNotFound`. `ListProjectsByCurrentUser` then re-wrapped everything as `Internal`; `ListServiceUserProjects` passed `NotFound` through. Both responses were wrong: the caller asked which of these permissions the principal holds; the answer for an unknown permission is "none".

Fix: new `resolvePermissionName` helper returns `(name, ok, err)`. `fetchAccessPairsOnResource` resolves each requested permission once, drops unknown names and duplicate inputs, and reserves `Internal` for genuine repository failures. `getPermissionName` is kept as a thin wrapper for callers that should reject unknown permissions (`CheckResourcePermission`, `CheckFederatedResourcePermission`, `BatchCheckPermission`).

### 3. Empty / malformed `id` returned 5xx instead of 400

After the `ListByUser` → `List(Filter{Principal})` migration, `project.Service.List` returned an untyped `fmt.Errorf` for invalid principals; non-UUID ids flowed to the SQL layer and surfaced as generic DB errors. Both produced `Internal` in the handlers. `ListServiceUserProjects` had the same shape via the interceptor's `GetServiceUser` call.

Fix:
- `project.Service.List` returns `project.ErrInvalidUUID` for empty/non-UUID `Principal.ID` and `project.ErrInvalidPrincipalType` for empty `Principal.Type`.
- `serviceuser.Service.Get` rejects non-UUID ids with `serviceuser.ErrInvalidID` (consistent with `user.Service.Enable`/`Disable`). The `GetServiceUser` handler maps `ErrInvalidID` to `InvalidArgument`, which propagates through the authorization interceptor.
- `ListProjectsByUser`, `ListServiceUserProjects` map `project.ErrInvalidUUID` / `project.ErrInvalidPrincipalType` to `InvalidArgument`. The dead `user.ErrInvalidUUID` case in `ListProjectsByUser` is removed.

## RPCs and status-code changes

| RPC | Input | Before | After |
|---|---|---|---|
| `ListProjectsByCurrentUser` | `withPermissions=["update","delete"]`, N projects | 2N duplicate `access_pairs` | N rows, one per project |
| `ListProjectsByCurrentUser` | unknown permission name | 500 | 200, drops that permission |
| `ListServiceUserProjects` | unknown permission name | 404 | 200, drops that permission |
| `ListServiceUserProjects` | multi-permission | 2N duplicate `access_pairs` | N rows |
| `ListServiceUserProjects` | non-UUID `id` | 500 | 400 |
| `ListProjectsByUser` | non-UUID `id` | 500 | 400 |
| `GetServiceUser` | empty / non-UUID `id` | 500 | 400 |
| `ListCurrentUserGroups` | multi-permission | 2N duplicate `access_pairs` | N rows |
| `ListCurrentUserGroups` | unknown permission name | 500 | 200, drops that permission |

## Frontend impact

None. `useOrganizationProjects` and `useOrganizationTeams` already collapse the response via `acc[id] = permissions` (last-write-wins). The pre-fix duplicate rows were identical, so old and new responses produce the same final map. `ListServiceUserProjects` frontend callers don't read `access_pairs`.

## Test plan

- [x] `go test ./core/project/... ./core/serviceuser/... ./internal/api/v1beta1connect/...`
- [x] `go vet ./...`
- [ ] Manual verification against staging with the request from the bug report — confirm one `access_pairs` row per project for `withPermissions=["update","delete"]`
- [ ] Manual verification: `ListServiceUserProjects` with non-UUID `id` returns 400
- [ ] Manual verification: `ListProjectsByCurrentUser` with `withPermissions=["bogus_perm"]` returns 200 with empty `access_pairs`

### DIFF
--- a/core/project/service.go
+++ b/core/project/service.go
@@ -133,8 +133,11 @@ func (s Service) Create(ctx context.Context, prj Project) (Project, error) {
 
 func (s Service) List(ctx context.Context, f Filter) ([]Project, error) {
 	if f.Principal != nil {
-		if f.Principal.ID == "" || f.Principal.Type == "" {
-			return nil, fmt.Errorf("project: invalid principal filter")
+		if !utils.IsValidUUID(f.Principal.ID) {
+			return nil, ErrInvalidUUID
+		}
+		if f.Principal.Type == "" {
+			return nil, ErrInvalidPrincipalType
 		}
 		if s.membershipService == nil {
 			return nil, fmt.Errorf("project: membership service not wired")
--- a/core/project/service_test.go
+++ b/core/project/service_test.go
@@ -288,14 +288,15 @@ func TestService_List(t *testing.T) {
 
 func TestService_List_WithPrincipal(t *testing.T) {
 	ctx := context.Background()
-	userPrincipal := authenticate.Principal{ID: "user-id", Type: schema.UserPrincipal}
+	userPrincipal := authenticate.Principal{ID: "68f86fec-eb87-49f0-9be0-8d99b00a4a9c", Type: schema.UserPrincipal}
 
 	tests := []struct {
-		name    string
-		setup   func(*testing.T) *project.Service
-		filter  project.Filter
-		want    []project.Project
-		wantErr bool
+		name      string
+		setup     func(*testing.T) *project.Service
+		filter    project.Filter
+		want      []project.Project
+		wantErr   bool
+		wantErrIs error
 	}{
 		{
 			name:   "errors when membership service is not wired",
@@ -309,24 +310,37 @@ func TestService_List_WithPrincipal(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			name:   "errors when Principal has empty ID",
+			name:   "returns ErrInvalidUUID when Principal has empty ID",
 			filter: project.Filter{Principal: &authenticate.Principal{Type: schema.UserPrincipal}},
 			setup: func(t *testing.T) *project.Service {
 				t.Helper()
 				repo, userService, suserService, relationService, policyService, authnService, groupService, roleService := mockService(t)
 				return project.NewService(repo, relationService, userService, policyService, authnService, suserService, groupService, roleService)
 			},
-			wantErr: true,
+			wantErr:   true,
+			wantErrIs: project.ErrInvalidUUID,
 		},
 		{
-			name:   "errors when Principal has empty Type",
-			filter: project.Filter{Principal: &authenticate.Principal{ID: "user-id"}},
+			name:   "returns ErrInvalidUUID when Principal ID is not a valid UUID",
+			filter: project.Filter{Principal: &authenticate.Principal{ID: "not-a-uuid", Type: schema.UserPrincipal}},
 			setup: func(t *testing.T) *project.Service {
 				t.Helper()
 				repo, userService, suserService, relationService, policyService, authnService, groupService, roleService := mockService(t)
 				return project.NewService(repo, relationService, userService, policyService, authnService, suserService, groupService, roleService)
 			},
-			wantErr: true,
+			wantErr:   true,
+			wantErrIs: project.ErrInvalidUUID,
+		},
+		{
+			name:   "returns ErrInvalidPrincipalType when Principal has empty Type",
+			filter: project.Filter{Principal: &authenticate.Principal{ID: "68f86fec-eb87-49f0-9be0-8d99b00a4a9c"}},
+			setup: func(t *testing.T) *project.Service {
+				t.Helper()
+				repo, userService, suserService, relationService, policyService, authnService, groupService, roleService := mockService(t)
+				return project.NewService(repo, relationService, userService, policyService, authnService, suserService, groupService, roleService)
+			},
+			wantErr:   true,
+			wantErrIs: project.ErrInvalidPrincipalType,
 		},
 		{
 			name:   "returns projects from the membership shim",
@@ -461,6 +475,9 @@ func TestService_List_WithPrincipal(t *testing.T) {
 			if (err != nil) != tt.wantErr {
 				t.Errorf("List() error = %v, wantErr %v", err, tt.wantErr)
 				return
+			}
+			if tt.wantErrIs != nil && !errors.Is(err, tt.wantErrIs) {
+				t.Errorf("List() error = %v, want errors.Is(%v)", err, tt.wantErrIs)
 			}
 			if diff := cmp.Diff(got, tt.want); diff != "" {
 				t.Errorf("List() mismatch (-want +got):\n%s", diff)
--- a/core/serviceuser/service.go
+++ b/core/serviceuser/service.go
@@ -112,6 +112,9 @@ func (s Service) Create(ctx context.Context, serviceUser ServiceUser) (ServiceUs
 }
 
 func (s Service) Get(ctx context.Context, id string) (ServiceUser, error) {
+	if !utils.IsValidUUID(id) {
+		return ServiceUser{}, ErrInvalidID
+	}
 	return s.repo.GetByID(ctx, id)
 }
 
--- a/core/serviceuser/service_test.go
+++ b/core/serviceuser/service_test.go
@@ -113,3 +113,53 @@ func TestService_Delete(t *testing.T) {
 		})
 	}
 }
+
+func TestService_Get(t *testing.T) {
+	ctx := context.Background()
+	const validID = "68f86fec-eb87-49f0-9be0-8d99b00a4a9c"
+
+	tests := []struct {
+		name      string
+		id        string
+		setup     func(*mocks.Repository)
+		wantErrIs error
+	}{
+		{
+			name:      "empty id returns ErrInvalidID without hitting the repo",
+			id:        "",
+			setup:     func(repo *mocks.Repository) {},
+			wantErrIs: serviceuser.ErrInvalidID,
+		},
+		{
+			name:      "non-uuid id returns ErrInvalidID without hitting the repo",
+			id:        "not-a-uuid",
+			setup:     func(repo *mocks.Repository) {},
+			wantErrIs: serviceuser.ErrInvalidID,
+		},
+		{
+			name: "valid uuid delegates to the repo",
+			id:   validID,
+			setup: func(repo *mocks.Repository) {
+				repo.On("GetByID", ctx, validID).Return(serviceuser.ServiceUser{ID: validID}, nil)
+			},
+		},
+	}
+
+	for _, tt := range tests {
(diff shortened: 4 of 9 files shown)


AmanGIT07:
## Sandbox RPC test results

Exercised the changed RPCs against a local Frontier built from this branch. Scaffold: one org with 3 projects, 1 group, 1 service user (`app_project_owner` on 2 of 3 projects); a second user with `app_project_viewer` on 1 project for filter-correctness.

### Listed deltas

| RPC | Case | Expected | Got |
|---|---|---|---|
| `ListProjectsByCurrentUser` | `[get,update]` (multi-perm dedup) | 3 pairs each `[get,update]` | ✅ |
| | `[bogus]` (was 500) | 200, 0 pairs | ✅ |
| | `[get,bogus]` (was 500) | 3 pairs each `[get]` | ✅ |
| | `[get,get]` (input dedup) | 3 pairs each `[get]` | ✅ |
| `ListCurrentUserGroups` | `[get,update]` | 1 pair `[get,update]` | ✅ |
| | `[bogus]` (was 500) | 200, 0 pairs | ✅ |
| | `[get,bogus]` | 1 pair `[get]` | ✅ |
| | `[get,get]` | 1 pair `[get]` | ✅ |
| `ListServiceUserProjects` | `[get,update]` | 2 pairs each `[get,update]` | ✅ |
| | `[bogus]` (was 404) | 200, 0 pairs | ✅ |
| | `[get,bogus]` | 2 pairs each `[get]` | ✅ |
| | `[get,get]` | 2 pairs each `[get]` | ✅ |
| | non-UUID id (was 500) | 400 | ✅ |
| | empty id | 400 | ✅ |
| `ListProjectsByUser` | non-UUID id (was 500) | 400 | ✅ |
| | empty id | 400 | ✅ |
| | random valid UUID, no user | 200, empty projects | ✅ |
| `GetServiceUser` *(direct)* | empty id | 400 | ✅ |
| | non-UUID id | 400 expected | **403** — see note |
| | valid UUID, not found | 404 expected | **403** — see note |
| `GetServiceUser` *(indirect via `DeleteServiceUser` / `ListServiceUserProjects` interceptor)* | non-UUID id | 400 | ✅ |
| | empty id | 400 | ✅ |
| | valid UUID, not found | 404 | ✅ |

**Note on direct `GetServiceUser`:** the interceptor at `pkg/server/connect_interceptors/authorization.go:257` calls `IsAuthorized(serviceuser:<id>, manage)` directly — it doesn't route through `handler.GetServiceUser`. So a non-UUID or missing-UUID id never reaches the new `ErrInvalidID` / `ErrNotExist` mapping; it falls into `IsAuthorized → CheckAuthz → handleAuthErr` and returns 403. The handler-level fix is wire-observable through every other interceptor that does call `handler.GetServiceUser` (`DeleteServiceUser`, `ListServiceUserProjects`), both verified above.

### Additional behavioral checks

| Test | Expected | Got |
|---|---|---|
| Permission-filter correctness — viewer-only user asks `[get,delete]` | 1 pair with `[get]` only (failed `delete` excluded) | ✅ |
| Empty `with_permissions` across all 3 list RPCs (omit + `[]`) | `access_pairs` absent from response | ✅ |
| Ordering stability — 5 sequential calls | identical order each time; `access_pairs` order matches `projects` order | ✅ |
| `CheckResourcePermission` known perm | 200 `{status: true}` | ✅ |
| `CheckResourcePermission` unknown perm | 404 (unchanged) | ✅ |
| `BatchCheckPermission` mixed known + unknown | 404 (unchanged) | ✅ |
| `BatchCheckPermission` all known | 200, one pair per body | ✅ |

rohilsurana (approved):

PR #1673, opened 2026-06-03
docs(web): add getting-started README for frontier/web

paanSinghCoder:
## What

Adds a getting-started README at `web/README.md` for the Frontier web monorepo, and links to it from the root README.

## Details

`web/README.md` covers:
- **What's inside** — `apps/admin`, `apps/client-demo`, the `@raystack/frontier` SDK, and shared `tools/*` configs
- **Prerequisites** — Node `>= 22`, pnpm `>= 10.19.0`
- **Getting started** — `pnpm install`, `pnpm dev`, Turbo filters, and per-package commands
- **Backend config** — the `.env` files each app needs, with localhost values (REST on `:8000`, ConnectRPC on `:8002`)
- **Common scripts** and `Makefile` shortcuts

Root `README.md` now links to it under a new **Web** subsection in **Usage**.

Docs only — no code changes.


web/README.md
```diff
@@ -0,0 +1,101 @@
+# Frontier Web
```
whoAbhishekSah, line 1: The admin app gets embedded when the server starts. 
It should read the latest Admin app with all the dependencies(SDK) by default. Can we add a step in the README to inform readers about this dependency so they can run the right steps, and, if possible, add it to the makefile to automate it with server start?

whoAbhishekSah (approved):

web/README.md
```diff
@@ -0,0 +1,101 @@
...
+`apps/admin/.env` (an `apps/admin/.env.example` is provided as a template):
+
+```sh
+FRONTIER_CONNECTRPC_URL=http://localhost:8002/
+```
+
+`apps/client-demo/.env`:
+
+```sh
+FRONTIER_ENDPOINT=http://localhost:8000/
```
rsbh: I don't think we need this environment. check is we are using it anywhere in the code

web/README.md
paanSinghCoder: Added a section in README.

web/README.md
paanSinghCoder: Removed.

Issue #1661, opened 2026-05-29
Authorization state drift: missing SpiceDB cleanup on role/permission/resource/invitation changes, no one-role DB constraint, schema pollution


whoAbhishekSah:
Several write paths don't fully reconcile the relational store (`policies`, `relations`, `roles`, `permissions`) with the SpiceDB datastore, so authorization data drifts over time: access that lingers after a delete, roles that grant more than they define, and stray permissions in the schema. PAT-specific gaps are tracked separately. Tracking here — I'll work on these.

### 1. Role deletion doesn't cascade to rolebindings / policies
`core/role` `Delete` removes the role's own permission relations (`app/role:<id>#<perm>`) but leaves the rolebindings that reference it (`app/rolebinding:<id>#role@app/role:<id>`) and the `policies` rows using it. A deleted role can keep granting access via surviving rolebindings, invisible/unremovable through the API.
**Fix:** on role delete, remove dependent policies (and their rolebinding tuples) first.

### 2. Role migration is add-only / doesn't prune
`core/role` `Update` correctly prunes removed permission relations, but the boot-time predefined-role migration in `internal/bootstrap` (a) rewrites predefined roles **without** removing dropped permission relations, and (b) never deletes roles removed from the schema. Result: narrowed roles still grant old permissions; removed roles persist.
**Fix:** during role migration, diff existing vs desired and delete removed permission relations and removed roles.

### 3. Invitations don't fully remove their SpiceDB relations
`core/invitation` creates two relations on create (`invitation#user@user` and `invitation#org@organization`) but `Delete` removes **only the org relation** — the user relation is never deleted, and accept/expire don't clean up either.
**Fix:** delete both relations on accept / expire / delete.

### 4. No database backstop for one role per (resource, principal)
Membership APIs enforce one role per principal per resource (replace-not-add), but the `policies` unique constraint includes `role_id`, so the generic policy-create path / direct writes can still create multiple roles for the same principal on the same resource.
**Fix:** tighten the `policies` unique constraint to `(resource_id, resource_type, principal_id, principal_type)` (after de-duplicating) and route writes through an upsert.

### 5. Ad-hoc permissions can pollute the compiled schema
`internal/bootstrap` builds the SpiceDB schema partly from the `permissions` table, so permissions created outside the intended resource definitions (e.g. via the create-permission API) get compiled into the schema and leave `app/role` relations/tuples not in the canonical base + generated schema. Nothing reconciles this.
[...]

PR #1678, opened 2026-06-05
fix(invitation): remove user relation on invitation delete

whoAbhishekSah:
## What

Invitation creation writes two SpiceDB tuples anchored on the invitation object:

- `app/invitation:<id>#user@app/user:<email>`
- `app/invitation:<id>#org@app/organization:<org>`

`Invitation.Delete` filtered the relation delete by the **org** relation name only, so the `#user` tuple was left behind on every **accept**, **expire**, and **delete**. These orphan tuples accumulated in SpiceDB over time.

## Fix

Drop the `RelationName` filter in `Delete` so it removes **every** relation anchored on the invitation object (the same pattern `role.Delete` already uses). `Accept` already routes its cleanup through `Delete`, so accept is covered too.

## Test

Extends the existing `TestInvitationAPI` accept flow with an assertion that, after accepting an invitation, `ListRelations` returns no relation on the invitation object. Verified the assertion **fails without the fix** (leftover `#user` tuple) and **passes with it**.

Addresses gap (3) of #1661.

### DIFF
--- a/core/invitation/service.go
+++ b/core/invitation/service.go
@@ -247,12 +247,15 @@ func (s Service) createRelations(ctx context.Context, invitationID uuid.UUID, or
 }
 
 func (s Service) Delete(ctx context.Context, id uuid.UUID) error {
+	// Remove every relation anchored on the invitation object, not just the org
+	// one. createRelations writes both a user (app/invitation:<id>#user) and an
+	// org (app/invitation:<id>#org) tuple; filtering by org alone leaked the user
+	// tuple on every accept/expire/delete. Omitting RelationName matches both.
 	err := s.relationService.Delete(ctx, relation.Relation{
 		Object: relation.Object{
 			ID:        id.String(),
 			Namespace: schema.InvitationNamespace,
 		},
-		RelationName: schema.OrganizationRelationName,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to delete relation for invitation: %w", err)
--- a/test/e2e/regression/api_test.go
+++ b/test/e2e/regression/api_test.go
@@ -2319,6 +2319,14 @@ func (s *APIRegressionTestSuite) TestInvitationAPI() {
 		}))
 		s.Assert().Error(err)
 
+		// no SpiceDB relation should linger on the invitation object after accept:
+		// both the #user and #org tuples must be gone, not just #org (gap #1661.3)
+		leftoverInviteRelations, err := s.testBench.AdminClient.ListRelations(ctxOrgAdminAuth, connect.NewRequest(&frontierv1beta1.ListRelationsRequest{
+			Object: schema.JoinNamespaceAndResourceID(schema.InvitationNamespace, createdInvite.GetId()),
+		}))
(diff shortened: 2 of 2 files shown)


whoAbhishekSah:
## Manual testing notes

Tested this against a local Frontier build of this branch (Postgres + SpiceDB). Summary below.

### What an invitation creates
For each `CreateOrganizationInvitation` we confirmed it writes:
- 1 row in the `invitations` table (Frontier DB)
- 2 SpiceDB tuples on the invitation object:
  - `app/invitation:<id>#user@app/user:<email-slug>`
  - `app/invitation:<id>#org@app/organization:<org-id>`

### What we tested
Created a fresh org and two invitations (one for each cleanup path), then checked the DB and SpiceDB before and after:

1. **Delete path** — called `DeleteOrganizationInvitation`. After delete: the DB row was gone and **both** the `#user` and `#org` tuples were removed.
2. **Accept path** — the invited user called `AcceptOrganizationInvitation`. After accept: the DB row was gone, **both** tuples were removed, and the user was correctly added as an org member.

In both cases SpiceDB had marked the tuples as deleted (it does a soft delete, setting a `deleted_xid` on the row rather than physically removing it), so they no longer show up as live relations.

### Bonus: the bug was real
Before running anything, the test DB already had 2 leftover `#user` tuples with **no matching DB row and no `#org` tuple** — exactly the leak the old `Delete` produced. After the fix, no new orphans are created.

### Note
This PR stops new leaks but does not clean up tuples that already leaked before the fix. Cleaning up existing orphans is a separate task.

✅ Working as expected.

core/invitation/service.go
```diff
@@ -247,12 +247,15 @@ func (s Service) createRelations(ctx context.Context, invitationID uuid.UUID, or
 }
 
 func (s Service) Delete(ctx context.Context, id uuid.UUID) error {
+	// Remove every relation anchored on the invitation object, not just the org
+	// one. createRelations writes both a user (app/invitation:<id>#user) and an
+	// org (app/invitation:<id>#org) tuple; filtering by org alone leaked the user
+	// tuple on every accept/expire/delete. Omitting RelationName matches both.
 	err := s.relationService.Delete(ctx, relation.Relation{
```
whoAbhishekSah, line 254: what about invite accepted flow ? does it clean all relations ?

whoAbhishekSah: Yes — the accept flow cleans all invitation relations. `Accept` ends with `return s.Delete(ctx, id)` (line 381), so it goes through this same `Delete`, which now removes every relation anchored on the invitation object (both `#user` and `#org`).

The membership tuples that accept creates — `AddOrganizationMember` (user→org) and `SetGroupMemberRole` (user→group) — are intentional and persist; those are the access the user is accepting, not invitation relations.

The regression test in this PR actually exercises the **accept** path specifically: it accepts the invitation, then asserts `ListRelations` on the invitation object returns empty (and it fails without the fix, confirming the leak).

rohilsurana (approved):

PR #1680, opened 2026-06-07
fix(api): add error sanitizer interceptor and fix billing response codes

rohilsurana:
## Summary

Add a Connect RPC interceptor that sanitizes `CodeInternal` and `CodeUnknown` errors before they reach clients — prevents leaking internal error details (e.g. database errors, raw service messages).

## How it works

The `UnaryConnectErrorSanitizerInterceptor` sits between OTEL and the logger in the interceptor chain. On the response path: the logger logs the original error with full detail, then the sanitizer replaces `CodeInternal`/`CodeUnknown`/non-connect errors with a generic `CodeInternal: "internal server error"` before the response reaches the client.

This makes it safe for handlers to return `connect.NewError(connect.CodeInternal, err)` with the real error — the interceptor ensures the client only ever sees the generic message.

## Test plan

- [x] `go test ./pkg/server/connect_interceptors/ -run TestUnaryConnectErrorSanitizerInterceptor -v` — 8 unit tests for the interceptor
- [x] `go build ./...` — full build passes

### DIFF
--- a/pkg/server/connect_interceptors/error_sanitizer.go
+++ b/pkg/server/connect_interceptors/error_sanitizer.go
@@ -1,0 +1,31 @@
+package connectinterceptors
+
+import (
+	"context"
+	"errors"
+
+	"connectrpc.com/connect"
+)
+
+func UnaryConnectErrorSanitizerInterceptor() connect.UnaryInterceptorFunc {
+	return func(next connect.UnaryFunc) connect.UnaryFunc {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+			resp, err := next(ctx, req)
+			if err == nil {
+				return resp, nil
+			}
+
+			var connectErr *connect.Error
+			if !errors.As(err, &connectErr) {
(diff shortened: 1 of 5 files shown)


pkg/server/connect_interceptors/error_sanitizer.go
```diff
@@ -0,0 +1,31 @@
...
+	return func(next connect.UnaryFunc) connect.UnaryFunc {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+			resp, err := next(ctx, req)
+			if err == nil {
+				return resp, nil
+			}
+
+			var connectErr *connect.Error
+			if !errors.As(err, &connectErr) {
+				return nil, connect.NewError(connect.CodeInternal, errors.New("internal server error"))
```
whoAbhishekSah: nit: this can be a const so that gets reused.

whoAbhishekSah (approved):

pkg/server/connect_interceptors/error_sanitizer.go
rohilsurana: Yes let me see if we can use from errors.go

PR #1683, opened 2026-06-08
fix(role): reject deletion of a role still in use by policies

whoAbhishekSah:
## Problem

Deleting a role touches two stores: SpiceDB (the role's permission tuples) and Frontier's Postgres `roles` table (the role row). The old order was: delete the SpiceDB tuples first, then delete the row in Postgres.

But the Postgres `policies` table has a foreign key to `roles`, so if a policy still uses the role, the Postgres row delete fails — *after* the SpiceDB tuples were already removed. Result: a broken role that still exists in Postgres but grants nothing in SpiceDB, plus a confusing `500`.

## Change

Deleting a role now **fails cleanly if the role is still used by any policy**:

- Delete the Postgres `roles` row first; only remove the SpiceDB permission tuples if that succeeds.
- A role still referenced by a policy is rejected with `FailedPrecondition` and the message "role is in use by one or more policies" — neither store is touched.
- To delete such a role, remove its policies first. (Deleting a policy already cleans up its own SpiceDB tuples, including the link back to the role.)

We don't auto-delete the policies, because that would silently revoke access for everyone using the role and could be a lot of deletes in one request.

## Tests

- Unit: role in use → returns the error, SpiceDB untouched; role not in use → Postgres row deleted, then its SpiceDB tuples cleaned.
- e2e (`TestOrganizationRoleDeleteInUse`): deleting an in-use role is rejected and nothing is torn down; after the policy is removed, the role deletes cleanly with no leftover SpiceDB tuples.

Addresses gap (1) of #1661.

### DIFF
--- a/core/role/errors.go
+++ b/core/role/errors.go
@@ -7,4 +7,5 @@ var (
 	ErrInvalidID     = errors.New("role id is invalid")
 	ErrConflict      = errors.New("role name already exist")
 	ErrInvalidDetail = errors.New("invalid role detail")
+	ErrRoleInUse     = errors.New("role is in use by one or more policies")
 )
--- a/core/role/service.go
+++ b/core/role/service.go
@@ -280,11 +280,24 @@ func (s Service) Update(ctx context.Context, toUpdate Role) (Role, error) {
 }
 
 func (s Service) Delete(ctx context.Context, id string) error {
-	if err := s.relationService.Delete(ctx, relation.Relation{Object: relation.Object{
-		ID:        id,
-		Namespace: schema.RoleNamespace,
-	}}); err != nil {
+	// resolve name->ID so the row delete and tuple cleanup target the same role
+	roleToDelete, err := s.Get(ctx, id)
+	if err != nil {
 		return err
 	}
-	return s.repository.Delete(ctx, id)
+
+	// Delete the row first. The policies.role_id foreign key rejects this while
+	// any policy still references the role, surfaced as ErrRoleInUse — so a role that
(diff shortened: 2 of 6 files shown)


whoAbhishekSah:
### Manual verification ✅

Tested the `DeleteOrganizationRole` flow against a local build of this branch (live ConnectRPC server + SpiceDB), following the same scenario as the `TestOrganizationRoleDeleteInUse` e2e test.

**Setup:** created an org, a custom role (`in_use_role`, perm `app.organization.grouplist`), a user, and a policy binding the role to the user on the org. At this point the role had 3 SpiceDB permission tuples and the policy had 2 rolebinding tuples.

| Step | Action | Expected | Result |
|------|--------|----------|--------|
| 1 | Delete role **while a policy still uses it** | rejected, nothing torn down | ✅ `FailedPrecondition` — "role is in use by one or more policies" (HTTP 400) |
| 2 | Re-fetch role + recount tuples after the failed delete | role intact | ✅ role still exists (200), perm tuples still 3 — no SpiceDB tuples touched |
| 3 | Delete the policy | succeeds, rolebinding tuples cleared | ✅ 200, rolebinding tuples → 0 |
| 4 | Delete the (now unused) role | succeeds | ✅ 200 |
| 5 | Re-fetch role + recount tuples after delete | gone, no leftovers | ✅ role 404, perm tuples → 0 |

Confirms the fix end-to-end: the FK guard fires before any SpiceDB tuples are removed (no more half-deleted role / opaque 500), the error surfaces cleanly as `FailedPrecondition`, and once the policy is gone the role deletes with no leftover tuples.

rohilsurana (approved):

PR #1567, opened 2026-04-23
refactor: consolidate principal-by-resource listing into membership

rohilsurana:
## Summary
- Adds `membership.ListPrincipalsByResource(ctx, resourceID, resourceType, filter)` as the single entrypoint for listing principals (users, service users, groups) that hold policies on a resource, with optional principal-type / role-ID filtering and role enrichment.
- Rewires every handler that previously hand-rolled member listings — `ListOrganizationAdmins`, `ListOrganizationUsers`, `ListProjectAdmins`, `ListProjectUsers`, `ListProjectServiceUsers`, `ListProjectGroups`, `ListGroupUsers`, `ListOrganizationGroups` (with members), `GetGroup` (with members), `RemoveGroupUser`, `ListUsers`/`ListAllUsers` (org/group filters) — to call the new method and hydrate entities via `*.GetByIDs`.
- Deletes the now-redundant `user.ListByOrg`, `user.ListByGroup`, `user.getUserIDsFromPolicies`, `project.ListUsers`, `project.ListServiceUsers`, `project.ListGroups`, and the handler-level direct calls to `policyService.ListRoles` for role enrichment. Drops the `PolicyService`/`RoleService` deps from `user.Service` and the `OrgID`/`GroupID` fields from `user.Filter`.
- `event.Service` (billing onboarding) switches from `userService.ListByOrg` to the membership method to look up the first org admin's email.

## Test plan
- [x] `go build ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] Unit tests: `go test ./...` green across core/membership, core/user, core/project, core/event, internal/api/v1beta1connect
- [x] New tests in `core/membership/service_test.go` cover resource-type validation, principal-type filter, role-ID filter, role enrichment, dedup, empty results, and error wrapping
- [x] `go test ./test/e2e/regression` passes end-to-end (organization, user, serviceuser suites)
- [x] Manually exercised 13 handlers (24 scenarios) against a local Frontier build — results in PR comment

PR #1694, opened 2026-06-10
refactor(serviceuser): list org service users via membership policies

rohilsurana:
## Summary

`serviceuser.Service.ListByOrg` was the last remaining caller that used `relationService.LookupSubjects` with the `membership` permission for listing purposes. SpiceDB lookups through that permission resolve via direct `member`/`owner` relations, which can go stale relative to policies (same class of issue #1567 fixed for `user.ListByOrg`).

This switches it to the policy-backed membership path:

- Added `membership.Service.ListPrincipalIDsByResource` — a primitive-typed convenience wrapper over `ListPrincipalsByResource`. Needed because `core/membership` imports `core/serviceuser`, so serviceuser's narrow `MembershipService` interface cannot reference membership types without an import cycle.
- `serviceuser.ListByOrg` now resolves service-user IDs via org policies (`ListPrincipalIDsByResource(orgID, OrganizationNamespace, ServiceUserPrincipal)`) and fetches rows with `repo.GetByIDs`.
- Removed `LookupSubjects` from serviceuser's `RelationService` interface — it has no remaining callers in the package.

With this change, `schema.MembershipPermission` has zero non-schema usages in Go code, clearing the path for the schema-side cleanup of `permission membership = member + owner` (follow-up).

Legacy service users that predate policy-backed membership (org relation but no org policy) are covered by the `MigrateServiceUserOrgPolicies` backfill that runs on serve.

## Test plan

- [x] New unit tests for `serviceuser.ListByOrg` (members found / empty / error propagation)
- [x] New unit tests for `membership.ListPrincipalIDsByResource`
- [x] `go test -race ./core/... ./internal/api/...` passes
- [x] `make lint` clean
- [ ] e2e regression (CI)


core/membership/service.go
```diff
@@ -1214,6 +1214,19 @@ func (s *Service) ListPrincipalsByResource(ctx context.Context, resourceID, reso
 	return members, nil
 }
 
+// ListPrincipalIDsByResource returns the IDs of principals of the given type
+// that have at least one policy on the resource. It is a primitive-typed
+// convenience wrapper over ListPrincipalsByResource for consumer packages
+// that cannot import membership types without creating an import cycle
+// (e.g. core/serviceuser, which this package itself imports).
+func (s *Service) ListPrincipalIDsByResource(ctx context.Context, resourceID, resourceType, principalType string) ([]string, error) {
```
rohilsurana, line 1258: For now adding this wrapper method, as other wise we are seeing cyclic dependecy when trying to import MemberFilter type.

rohilsurana:
Addressed review nitpicks in 2bda5890:
- `ListPrincipalIDsByResource` is now a lean single-query path (shared `resourcePolicyFilter` helper, no role enrichment) — the role-service error case is moot since roles are no longer involved
- Added `repo.GetByIDs` failure-path case to `TestService_ListByOrg`

whoAbhishekSah (approved):

PR #1696, opened 2026-06-11
refactor(deleter): delegate org removal cascade to membership

rohilsurana:
## Summary

`deleter.RemoveUsersFromOrg` reimplemented the org-member removal cascade inline (policy walks over projects/groups/resources), duplicating logic that `membership.RemoveOrganizationMember` already handles canonically. This PR consolidates the removal path and deletes the now-orphaned inline implementations.

### Changes

- **`membership.ForceRemoveOrganizationMember`** (new): same cascade as `RemoveOrganizationMember` but without the member-and-owner guards — no `ErrNotMember` when the principal has no org policies, and it removes the org's last owner without complaint. Deletion cascades need this: deleting a user who solely owns an org must succeed (the org is left ownerless, same as today's behavior). Anything user-initiated keeps going through the guarded variant. Implemented by parameterizing the existing method; the cascade internals already supported unguarded owner deletes.
- **`deleter.RemoveUsersFromOrg`**: now cleans up custom-resource policies (outside the membership cascade's scope) and delegates everything else to `ForceRemoveOrganizationMember` per user. Also fixes a latent bug where the policy-list error check tested the wrong variable (`err` instead of `policyErr`).
- **Deleted `organization.Service.RemoveUsers` and `group.Service.RemoveUsers`** (+ its private helper): the deleter was their only caller — API handlers already use the membership service directly. Dropped the org service's now-unused `AuditRecordRepository` dependency and `extractPrincipalInfo` helper, and removed dead method declarations from the deleter/API interfaces (`CascadeDeleter.RemoveUsersFromOrg` had no handler callers).

### Behavior notes

- Org-member-removed audit records (auditrecord + legacy stream) are still emitted with the same shape — now by the membership cascade instead of `org.RemoveUsers`.
- Group-scoped policy cleanup during user deletion previously emitted a per-group `GroupMemberRemovedEvent` legacy audit via `group.RemoveUsers`; the membership cascade removes the same policies/relations but emits the org-level removal audit instead. User deletion itself remains separately audited.
- Net: -264 LOC.

## Test plan

- [x] New unit tests: `membership.ForceRemoveOrganizationMember` (last-owner bypass, no-ErrNotMember cascade, error propagation)
- [x] New unit tests: `deleter.RemoveUsersFromOrg` (membership delegation, custom-resource cleanup incl. out-of-org and missing resources, error propagation)
- [x] `go test -race ./core/... ./internal/api/... ./cmd/...` passes
- [x] `make lint` clean
- [ ] e2e regression (CI)


core/deleter/service.go
```diff
@@ -321,69 +323,46 @@ func (d Service) RemoveUsersFromOrg(ctx context.Context, orgID string, userIDs [
...
-				if !errors.Is(resErr, resource.ErrNotExist) {
-					if resErr != nil {
-						err = errors.Join(err, resErr)
-					} else if userResource.ProjectID != "" && utils.Contains(orgProjectIDs, userResource.ProjectID) {
-						// if the resource belong to org project, delete access
-						if policyErr := d.policyService.Delete(ctx, pol.ID); policyErr != nil {
-							err = errors.Join(err, policyErr)
-						}
+				if errors.Is(resErr, resource.ErrNotExist) {
+					continue
```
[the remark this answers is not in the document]
rohilsurana: This is pre-existing behavior preserved exactly — the old code had the identical skip (`if !errors.Is(resErr, resource.ErrNotExist)` wrapped the whole block, so a missing resource was a no-op there too). This PR only restructured the control flow around it.

The underlying concern is real but, as you note yourself, not fixable here: once the resource row is gone there is no ownership information left to scope the policy to this org, so deleting it blindly from an *org-scoped* removal would be guessing. Orphaned resource policies are a systemic cleanup concern (they're unreachable by every removal path, not just this one), and the right home for fixing it is the resource service owning its access lifecycle — there's a tracked follow-up to move this custom-resource policy walk behind `resourceService` with proper orphan handling, replacing the per-policy `Get` round-trips with a single ownership-resolving query.

The nitpick regression test for the policy-list error branch was added in c0ebbe4c.

core/membership/service.go
```diff
@@ -378,8 +378,23 @@ func (s *Service) removePoliciesByFilter(ctx context.Context, filter policy.Filt
...
+
+// ForceRemoveOrganizationMember is RemoveOrganizationMember without the
+// member-and-owner guards: it does not fail when the principal has no
+// org-level policies and it removes the org's last owner without complaint.
+// It exists for deletion cascades (see core/deleter), where the principal is
+// leaving the system entirely and the org may legitimately be left ownerless.
+// Anything user-initiated should go through RemoveOrganizationMember instead.
+func (s *Service) ForceRemoveOrganizationMember(ctx context.Context, orgID, principalID, principalType string) error {
+	return s.removeOrganizationMember(ctx, orgID, principalID, principalType, false)
+}
```
AmanGIT07, line 395: `ForceRemoveOrganizationMember` will fail for disabled org since `s.orgService.Get(ctx, orgID)` in `removeOrganizationMember` will return `ErrDisabled`. We can add a new method like GetRaw or maybe ignore this error for forced removal

rohilsurana: Good catch — this was a real regression: `DeleteUser` deliberately visits disabled orgs (so user deletion doesn't leave orphan policies), and the old path worked because `org.RemoveUsers` used the state-blind `repository.GetByID`.

Fixed in 1cfd9775 by tolerating `ErrDisabled` in the force path rather than adding `GetRaw` — `Get` discards the org payload on `ErrDisabled`, so the force path reconstructs the minimal org from the input ID and proceeds with the cascade. The guarded `RemoveOrganizationMember` still rejects disabled orgs as before. One minor trade-off: the audit record's resource name is empty for disabled-org removals (the title isn't available without an extra fetch).

Added a test case asserting the cascade completes on a disabled org.

AmanGIT07 (approved):

PR #1707, opened 2026-06-18
feat: add customisable button slots to service account views

rohanchkrabrty:
## Summary

- Introduce a generic `Slot<TContext>` render-prop type in `shared/types.ts` so consumers can customise internal view elements while keeping the customisation on their side.
- Add a `slots.addAccountButton` slot to `ServiceAccountsView` (header toolbar + empty state), exposing the `onClick` handler and permission-derived `disabled` state.
- Add a `slots.generateKeyButton` slot to `ServiceAccountDetailsView`, threaded into the token form and exposing the form's `loading`/`disabled` state.
- Default buttons render as the conditional fallback, so behaviour is unchanged when no slot is provided.

### DIFF
--- a/web/sdk/client/views/service-accounts/components/add-token-form.tsx
+++ b/web/sdk/client/views/service-accounts/components/add-token-form.tsx
@@ -27,7 +27,11 @@ type FormData = yup.InferType<typeof tokenSchema>;
 export interface AddTokenFormProps {
   serviceUserId: string;
   onAddToken: (token: ServiceUserToken) => void;
-  generateKeyButton?: Slot<{ loading: boolean; disabled: boolean }>;
+  generateKeyButton?: Slot<{
+    onClick: () => void;
+    loading: boolean;
+    disabled: boolean;
+  }>;
 }
 
 export function AddTokenForm({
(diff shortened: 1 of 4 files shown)


Shreyag02 (approved):

web/sdk/client/views/service-accounts/components/add-token-form.tsx
```diff
@@ -26,9 +27,18 @@ type FormData = yup.InferType<typeof tokenSchema>;
 export interface AddTokenFormProps {
   serviceUserId: string;
   onAddToken: (token: ServiceUserToken) => void;
+  generateKeyButton?: Slot<{
```
rohilsurana, line 30: this has to be of type = 'button'. else it can default to type='submit' and hence form can submit twice, once the custom onClick and another on onSubmit.

rohanchkrabrty: It's upto the consumer. Consumer can chose to render `<Button type="submit"` or `<Button onClick`. We are not defining the component here, we are just exposing the base props

rohilsurana: yeah but if type is not provided default added will be `submit` . Hence it can cause the issues that we are talking about. How about we handle the submit event and prevent default?

rohilsurana: or if this is an accepted pattern then it should be fine.

rohilsurana (approved):

web/sdk/client/views/service-accounts/components/add-token-form.tsx
rohanchkrabrty: if button has default `type=submit"` and also onClick, it won't create a problem as the form library dedupes the request. 
This is fine with respect to slots pattern

PR #1708, opened 2026-06-19
feat(bootstrap): grant project-create to org member role

whoAbhishekSah:
## What

1. Grants `app_organization_projectcreate` to the predefined organization viewer ("Member") role, so an org Member can create projects rather than only reading the org.
2. Removes a duplicate `PredefinedRoles` entry titled "Group Viewer" that reused the same `Name` (`RoleOrganizationViewer` = `app_organization_viewer`).

## Why the duplicate removal

`PredefinedRoles` defined `RoleOrganizationViewer` twice — "Member" and "Group Viewer" — with the same `Name`, so it never created a distinct role. `MigrateRoles` reconciled the same role twice (last write winning), which meant any edit to the Member role had to be mirrored in both entries or the second would reset it. The "Group Viewer" copy traces back to #399 and looks like a copy/paste that reused the org-viewer name instead of a dedicated group-viewer constant. Removing it is functionally inert (same `Name`, no separate role) and makes the role single-sourced.

## Impact

Additive for permissions — Member gains `projectcreate`, nothing is removed. The dedupe changes no runtime behavior (the duplicate was a redundant reconcile of the same role). Applied via the normal boot-time role reconcile.

### DIFF
--- a/internal/bootstrap/schema/schema.go
+++ b/internal/bootstrap/schema/schema.go
@@ -344,14 +344,7 @@ var PredefinedRoles = []RoleDefinition{
 		Name:  RoleOrganizationViewer,
 		Permissions: []string{
 			"app_organization_get",
-		},
-		Scopes: []string{OrganizationNamespace},
-	},
-	{
-		Title: "Group Viewer",
-		Name:  RoleOrganizationViewer,
-		Permissions: []string{
-			"app_organization_get",
+			"app_organization_projectcreate",
 		},
 		Scopes: []string{OrganizationNamespace},
 	},


whoAbhishekSah:
Closing — adding app_organization_projectcreate to the base Member role over-grants: service users (and all members) hold Member by default, so this would let every member/service-user create projects, which the serviceusers regression tests correctly guard against. The intended model is to grant project creation via the Manager role or an explicit policy; the read-only Member role is correct. The duplicate-role cleanup can be done separately if wanted.

PR #1709, opened 2026-06-21
feat: GitOps platform-user management + bootstrap superuser SA (protected, config-only)

### LINKED ISSUE #1710: RemovePlatformUser is not relation-selective (asymmetric with AddPlatformUser)
## Summary

`AddPlatformUser` and `RemovePlatformUser` are asymmetric in their relation granularity:

- **`AddPlatformUser`** is relation-specific — the request carries a `relation` field (`admin` or `member`), so you can grant exactly one platform relation.
- **`RemovePlatformUser`** is relation-agnostic — `RemovePlatformUserRequest` has no `relation` field, so the handler strips **both** `admin` and `member` (full platform removal).

As a result there is no API way to remove only one platform relation — e.g. to demote a platform admin down to `member`, or to revoke `member` while leaving `admin`. Removal is all-or-nothing.

## Current behavior (for reference)

- `AddPlatformUser(id, relation)` → `Sudo(id, relation)` (user or service account).
- `RemovePlatformUser(id)` → loops `["admin", "member"]` calling `UnSudo(id, relation)` (user or service account).

The underlying service primitives are already symmetric and relation-specific: `Sudo(ctx, id, relationName)` ↔ `UnSudo(ctx, id, relationName)`. Only the `RemovePlatformUser` handler is coarse-grained, because the proto request lacks a `relation` field.

## Proposed fix

1. Add an optional `relation` field to `RemovePlatformUserRequest` in [`raystack/proton`](https://github.com/raystack/proton) (`raystack/frontier/v1beta1`).
2. In the `RemovePlatformUser` handler: when `relation` is set, validate via `IsPlatformRelation` and call `UnSudo(id, relation)` for that relation only; when unset, keep the current behavior (remove both) for backward compatibility.

No service-layer change is needed — `UnSudo(ctx, id, relationName)` already supports relation-specific removal.

## Notes

- Requires a proto change in `raystack/proton`, which is why it was left out of the original PR.
- Context: this gap surfaced while implementing opt-in authoritative superuser reconciliation.

rohilsurana:
## What

Moves platform-superuser management to a **GitOps** model. Config no longer seeds human
superusers; instead it seeds a single **bootstrap superuser service account**, and all
ongoing platform-user management (humans and service accounts, admin and member) happens
out-of-band via a declarative **`frontier reconcile`** CLI driven from the IAM repo. The
bootstrap SA is treated as a protected, config-only break-glass identity.

> Note: this PR started as "opt-in authoritative reconciliation" and then pivoted to GitOps
> (dropping authoritative reconciliation *and* the whole config human-superuser flow), and
> later grew a bootstrap-SA hardening pass. Commits are kept as-is by request, so the history
> is winding — the **final diff** is the source of truth; this description summarises the end
> state.

## Changes

**Remove the config human-superuser flow**
- Dropped `app.admin.users` + `MakeSuperUsers` + the boot-time `authoritative` reconciliation
  that an earlier revision of this PR had added. No human superusers are seeded from config.

**Config-bootstrapped superuser service account** (`app.admin.bootstrap.{client_id, client_secret}`)
- At boot, idempotently ensures a `ClientSecret` service-account credential exists in the
  platform org and is promoted to superuser; rotates the secret if it changed. This is the only
  config-seeded superuser and the break-glass identity. `client_id` must be a UUID (it is the
  credential id); validated with a fail-fast error.
- The SA is created with a **fixed, well-known service-user id**
  (`00000000-0000-0000-0000-000000000001`, a deliberately non-nil sentinel — `uuid.Nil` is the
  audit "system" actor and must not be a real principal).
- Automation (the reconcile workflow) authenticates as this SA via HTTP Basic.
- If `creds.Create` fails after the service-user row is created, the orphan row is rolled back so
  repeated boot failures don't accumulate credential-less service users.

**Harden the bootstrap SA — immutable via the API (security)**
- The SA is managed only via `app.admin.bootstrap` at boot. The API refuses, by its well-known
  id, to: **remove** its platform access (`RemovePlatformUser`), **delete** it
  (`DeleteServiceUser`), or **mint** credentials/tokens/keys for it
  (`CreateServiceUserCredential` / `Token` / `JWK`). Not even platform superusers may — this
  prevents a rotation-proof superuser backdoor on a publicly-known id.
- The authoritative reconciler **excludes** the bootstrap SA by id, so an apply never plans to
  remove it. (Validated end-to-end on a live instance, including the escalation path where the
  per-object `manage` authz is satisfied — the guard still blocks it.)
- Boot-time secret rotation is unaffected (it writes via the repositories, not these handlers).

**Fix platform-user grant/revoke correctness**
- `Sudo` is now idempotent on the **exact relation tuple**, not the derived permission. `check`
  is granted by both `admin` and `member`, so the old permission check skipped creating the
  `member` tuple for an existing admin — silently breaking admin→member downgrades. `Sudo` and
  `UnSudo` are now symmetric (both relation-level).
- `UnSudo(ctx, id, relationName)` is parameterized and actually strips the requested relation
  (previously it could only remove `member`, never `admin`).
- `RemovePlatformUser` honors the new optional `relation` field (raystack/proton#489): when set,
  removes just that relation (e.g. demote admin → member); empty keeps the remove-both default.
- `ListPlatformUsers` reports **all** relations a principal holds (`metadata.relations`), not just
  the last-seen one, so the reconciler converges to the desired state exactly (admin-only,
  member-only, or both — extras are removed). `metadata.relation` is retained for backward
  compatibility.

**Audit platform grants/revokes**
- New `platform.{admin,member}_{added,removed}` events via the `auditrecord` system, on both the
  RPC path (real principal) and boot/config path (system actor, `uuid.Nil`). Change-only.

**Declarative reconcile framework + CLI**
- `internal/reconcile`: a `Reconciler` interface + registry + kind-based (`kind:`) multi-doc YAML,
  with **PlatformUser** as the first kind (parse → ListPlatformUsers → diff `(principal, relation)`
  → apply). Future kinds (roles, plans, traits…) plug in without changing the command/format.
- `frontier reconcile -f <file> [--dry-run] -H <auth>` — authoritative converge of the desired
  state; auth via the existing `--header` mechanism.

**Supporting fix**
- `serviceuser_credentials` org_name made nullable-tolerant (`sql.NullString`) — a platform-level
  service account has no real organization row.

## Testing

- Unit suite green (incl. new `internal/reconcile`, bootstrap SA, audit, relation-selective
  removal, the Sudo/list correctness fixes, and the API immutability guards). Tests run at
  `bcrypt.MinCost` so they don't blow the `-race -count 2` timeout.
- **Full e2e regression suite passes** against real Postgres + SpiceDB — the testbench seeds its
  superuser the production way (authenticates as the bootstrap SA and grants the test admin via
  `AddPlatformUser`). The e2e caught two real integration bugs the unit mocks couldn't (UUID
  `client_id`, nullable `org_name`), both fixed here.
- **Live security validation** against a real instance: confirmed the bootstrap SA cannot be
  removed, deleted, or have credentials/tokens/keys minted — as a superuser, a regular user, or
  unauthenticated — and that the API guard still blocks minting even when the per-object `manage`
  authorization is deliberately satisfied.
- `go build ./...`, `go vet`, `gofmt`, `golangci-lint` all clean.

## Related

- raystack/proton#489 — adds the optional `RemovePlatformUserRequest.relation` field (merged; this PR
  bumps the proton pin and regenerates).
- pixxelhq/IAM#139 — the GitOps consumer: desired-state files + reconcile workflow.
- Closes #1710 (relation-selective platform-user removal).

## Notes

- Default-safe: with no `app.admin.bootstrap` configured, boot is a no-op (no superusers seeded);
  existing platform relations are untouched.
- Breaking for operators relying on `app.admin.users`: that config is removed — seed the bootstrap
  SA instead and manage humans via GitOps.


rohilsurana:
Superseded by the 2-PR split of this work:
- #1719 — relation-level platform-user grant/revoke + relation-selective removal + audit (base `main`)
- #1720 — GitOps platform-user management: bootstrap superuser SA + reconcile CLI + hardening (stacked on #1719)

Together they equal this PR's final diff. Closing in favour of the smaller, independently-reviewable PRs.

PR #1713, opened 2026-06-30
docs(authz): custom resource loading and role inheritance

whoAbhishekSah:
## What

Adds a new authz doc: `docs/content/docs/authz/custom-resources.mdx`.

It explains how custom resource types work end to end.

## Notes

- Docs only. No code changes.


docs/content/docs/authz/custom-resources.mdx
```diff
@@ -0,0 +1,369 @@
...
+```
+
+None of these appears anywhere in the custom-action rules above. So the Admin role does
+**not** get custom resource actions through org inheritance. To act on a custom resource, an
+Admin would need a project role that lists the action, a project admin role, or a direct grant
+on the resource.
+
+---
+
+## Project-level actions: use `user/project` as a proxy for `app/project`
```
AmanGIT07, line 156: Is there any convention we follow for the proxy namespace? I suggest we can mention it to be `*/project` for `app/project` and likewise

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -0,0 +1,369 @@
...
+You do not have to set up any roles for these. They work as soon as the resource is registered:
+
+- **Owner** (`app_organization_owner`) — can do every action on every custom resource in the
+  org.
+- **Project Owner** (`app_project_owner`) — can do every action on resources in their project.
+- **Platform admin** — can do everything.
+- The **user who created a resource** — can act on that one resource.
+
+This works because the [generated rules](#what-rules-get-generated) already include the owner and
+admin permissions. So an owner or an admin is covered without the action being listed in any role.
```
AmanGIT07: There are two admins, platform admin and org admin, we should explicitly mention which one are we talking because it can get confusing at times like here

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -0,0 +1,369 @@
...
+      - user/project:createcomputemachine
+      - user/project:listcomputemachines
+```
+
+A new role starts empty, so you only list what you want it to have. After boot, you assign this
+role to a user or group on a project, the same way you assign any other role.
+
+### In short
+
+- Owners and admins get every custom action for free.
```
AmanGIT07: https://github.com/raystack/frontier/pull/1713/changes#r3527111157

Same here

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -0,0 +1,369 @@
...
+    permissions:
+      - app/project:get             # keep what the role already had
+      - app/project:resourcelist
+      - compute/machine:get
+      - user/project:listcomputemachines
+```
+
+One thing to watch: when you list a role that already exists, Frontier **replaces** its whole
+permission set with the one you write. It does not add to the old set. So you must include the
+permissions the role already had, or it will lose them. In the example, `app/project:get` is
```
AmanGIT07: > must include the permissions the role already had

You mean all the existing one from the `base_schema` right?

AmanGIT07 (approved):

PR #1715, opened 2026-06-30
feat(admin): open stripe customer portal from billing section

rohilsurana:
## What

Adds an edit (pencil) action to the right of the **Billing** section header in the organization details side panel of the admin app (matching the Figma). Clicking it opens the Stripe customer portal for that org so a super user can update the customer's billing address / details.

This works **even when the org's KYC is completed** — unlike the client app, the admin app does not gate the action on KYC.

## How

- Reuses the existing `FrontierService/CreateCheckout` RPC with `setupBody.customerPortal = true` (the same call the client app's billing view uses). Super-user authorization already resolves via the `platform->superuser` schema cascade, so **no backend change is needed**.
- **Opens in a new tab, popup-blocker safe**: the tab is opened synchronously within the click gesture and navigated once the portal URL is ready (opener severed for safety; the tab is closed if the request fails or returns no URL). Stripe's return URL is the current admin page.
- **Disabled state + tooltip**: the action is disabled while the request is in flight or when the org has no billing account. It uses `aria-disabled` (not the native attribute) so it stays hoverable and shows a tooltip — "No billing account for this organization" when disabled, "Update billing details" otherwise. Errors surface via `toastManager`.
- **Icon**: apsara `1.0.0-rc.12` ships no edit/pencil icon, so this uses `Pencil1Icon` from `@radix-ui/react-icons` (already used across the admin app). The header lays the title + action out as a flex row so the icon is right-aligned.

## Related

- Backend audit-record PR (independent): #1714 — records who (super user or any other principal) opened the portal.
- Client-side new-tab PR (independent): #1717 — applies the same new-tab behavior to the client billing card.

## Test plan

- `eslint` passes on the changed file.
- Manual (verified against a local sandbox):
  - Org **with** a billing account → pencil enabled, right-aligned, tooltip "Update billing details"; clicking calls `CreateCheckout` and opens the portal in a new tab.
  - Org **without** a billing account → pencil disabled, with tooltip "No billing account for this organization".
  - Works regardless of KYC status.

### DIFF
--- a/web/sdk/admin/views/organizations/details/side-panel/billing-details-section.tsx
+++ b/web/sdk/admin/views/organizations/details/side-panel/billing-details-section.tsx
@@ -1,12 +1,12 @@
-import { Amount, CopyButton, Flex, Link, List, Text } from "@raystack/apsara";
+import { Amount, CopyButton, Flex, IconButton, Link, List, Text, Tooltip, toastManager } from "@raystack/apsara";
 import styles from "./side-panel.module.css";
 import { convertBillingAddressToString } from "../../../../utils/helper";
 import Skeleton from "react-loading-skeleton";
 import { useContext, useEffect } from "react";
-import { CalendarIcon } from "@radix-ui/react-icons";
+import { CalendarIcon, Pencil1Icon } from "@radix-ui/react-icons";
 import { OrganizationContext } from "../contexts/organization-context";
-import { useQuery } from "@connectrpc/connect-query";
-import { FrontierServiceQueries, GetUpcomingInvoiceRequestSchema } from "@raystack/proton/frontier";
+import { useMutation, useQuery } from "@connectrpc/connect-query";
+import { CreateCheckoutRequestSchema, FrontierServiceQueries, GetUpcomingInvoiceRequestSchema } from "@raystack/proton/frontier";
 import { create } from "@bufbuild/protobuf";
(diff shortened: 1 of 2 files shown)


paanSinghCoder (approved):

web/sdk/admin/views/organizations/details/side-panel/billing-details-section.tsx
```diff
@@ -39,9 +39,81 @@ export const BillingDetailsSection = () => {
...
+      portalTab?.close();
+    }
+  };
+
   return (
     <List>
-      <List.Header>Billing</List.Header>
+      <List.Header className={styles["billing-header"]}>
+        Billing
+        <Tooltip>
```
rohanchkrabrty, line 91: The tooltip icon needs to be visible only when the user hovers on the billing section. CSS can be used to achieve this

rohanchkrabrty (changes requested):
Also remove the comments which are self explanatory like the ones in CSS file etc

web/sdk/admin/views/organizations/details/side-panel/billing-details-section.tsx
rohilsurana: I am assuming you meant, edit icon. Have made that change now. Only visible when hovering over billing section.

Also made one more change on disabled state the icon should also not show background change as well as the cursor should not change.

rohanchkrabrty (approved):

PR #1720, opened 2026-06-30
feat: GitOps platform-user management (bootstrap superuser SA + reconcile CLI)

rohilsurana:
## What

This PR moves platform-superuser management to a file-based (GitOps) flow. Config seeds one protected bootstrap **service account** instead of a list of human superusers. From then on, you manage access by editing a file and running the new `frontier reconcile` command. This is the second of a 2-PR split of #1709, and it builds on #1719.

> ⚠️ **Breaking change** — this removes `app.admin.users`. Only merge it once the new flow is ready: the bootstrap service account is configured and pixxelhq/IAM#139 is in place. Otherwise human superadmins lose access until the first reconcile runs.

## Changes

- **Bootstrap superuser service account** (`app.admin.bootstrap.{client_id, client_secret}`) — created and promoted to superuser at boot, with a fixed id. You rotate the secret by changing the config. `client_id` must be a UUID.
- **Locked down** — this service account cannot be changed through the API. It can't be removed, deleted, or have new credentials, tokens, or keys created, not even by a superuser. The reconcile flow skips it too. Rotating the secret at boot still works.
- **`frontier reconcile` command** — reads a file that lists who should have access and makes the platform match it. Anyone listed is added; anyone not listed is removed. It lives in `internal/reconcile` and can support more kinds later; PlatformUser is the first.
- **Removes** `app.admin.users` and `MakeSuperUsers`.
- **Supporting change** — `org_name` can now be null, so the platform service account (which has no org) can be stored.

## Testing

Unit tests pass (`-race -count 2`). Vet, lint, and format are clean. Full e2e tests pass (the test bench seeds data using the bootstrap service account). I also ran a live security check: no one can remove or delete the service account, or create credentials for it, even when the per-object permission check is made to pass.

## Related

- Builds on #1719.
- pixxelhq/IAM#139 — the repo that uses this flow (the desired-state files and the reconcile workflow).

### DIFF
--- a/cmd/reconcile.go
+++ b/cmd/reconcile.go
@@ -65,11 +65,11 @@ func printReconcileReport(cmd *cli.Command, rep reconcile.Report) {
 		cmd.Printf("%s: no changes\n", rep.Kind)
 		return
 	}
-	verb := "applied"
+	verb, count := "applied", rep.Applied
 	if rep.DryRun {
-		verb = "planned"
+		verb, count = "planned", len(rep.Planned)
 	}
-	cmd.Printf("%s (%s %d):\n", rep.Kind, verb, len(rep.Planned))
+	cmd.Printf("%s (%s %d):\n", rep.Kind, verb, count)
 	for _, p := range rep.Planned {
 		cmd.Printf("  - %s\n", p)
 	}
--- a/cmd/reconcile.go
+++ b/cmd/reconcile.go
@@ -17,13 +17,13 @@ func ReconcileCommand(cliConfig *Config) *cli.Command {
 	)
 	cmd := &cli.Command{
 		Use:   "reconcile",
-		Short: "Reconcile declarative platform configuration to a desired-state file",
+		Short: "Reconcile platform configuration to a desired-state file",
 		Long: heredoc.Doc(`
-			Converge platform resources to a declarative YAML spec via the admin API.
+			Make platform resources match a desired-state YAML file, through the admin API.
 
-			Currently supports the PlatformUser kind (platform admins/members). The file
-			is the source of truth: entries present are ensured, entries absent are removed.
-			Authenticate as a superuser (e.g. the bootstrap service account) via --header.
+			Supports the PlatformUser kind (platform admins and members) for now. The file
+			decides who has access: anyone listed is added, anyone not listed is removed.
+			Log in as a superuser (for example the bootstrap service account) with --header.
 		`),
 		Example: heredoc.Doc(`
 			$ frontier reconcile -f platform-users.yaml --dry-run -H "Authorization:Basic <base64>"
--- a/cmd/reconcile.go
+++ b/cmd/reconcile.go
@@ -1,0 +1,76 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/raystack/frontier/internal/reconcile"
+	cli "github.com/spf13/cobra"
+)
+
+func ReconcileCommand(cliConfig *Config) *cli.Command {
+	var (
+		filePath string
+		dryRun   bool
+		header   string
+	)
+	cmd := &cli.Command{
+		Use:   "reconcile",
+		Short: "Reconcile declarative platform configuration to a desired-state file",
+		Long: heredoc.Doc(`
+			Converge platform resources to a declarative YAML spec via the admin API.
+
+			Currently supports the PlatformUser kind (platform admins/members). The file
+			is the source of truth: entries present are ensured, entries absent are removed.
+			Authenticate as a superuser (e.g. the bootstrap service account) via --header.
+		`),
+		Example: heredoc.Doc(`
+			$ frontier reconcile -f platform-users.yaml --dry-run -H "Authorization:Basic <base64>"
+			$ frontier reconcile -f platform-users.yaml -H "Authorization:Basic <base64>"
+		`),
+		Annotations: map[string]string{
+			"group":  "core",
+			"client": "true",
+		},
+		RunE: func(cmd *cli.Command, args []string) error {
+			data, err := os.ReadFile(filePath)
+			if err != nil {
+				return fmt.Errorf("read desired-state file: %w", err)
+			}
+			adminClient, err := createAdminClient(cliConfig.Host)
+			if err != nil {
+				return err
+			}
+			registry := map[string]reconcile.Reconciler{
+				reconcile.KindPlatformUser: reconcile.NewPlatformUserReconciler(adminClient, header),
+			}
+			reports, runErr := reconcile.Run(cmd.Context(), registry, data, dryRun)
+			for _, rep := range reports {
+				printReconcileReport(cmd, rep)
+			}
+			return runErr
+		},
+	}
+	cmd.Flags().StringVarP(&filePath, "file", "f", "", "Path to the desired-state YAML file")
+	cmd.MarkFlagRequired("file")
+	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "Print the plan without applying changes")
+	cmd.Flags().StringVarP(&header, "header", "H", "", "Header <key>:<value> for auth, e.g. 'Authorization:Basic <base64>'")
+	bindFlagsFromClientConfig(cmd)
+	return cmd
+}
+
+func printReconcileReport(cmd *cli.Command, rep reconcile.Report) {
+	if len(rep.Planned) == 0 {
+		cmd.Printf("%s: no changes\n", rep.Kind)
+		return
+	}
+	verb := "applied"
+	if rep.DryRun {
+		verb = "planned"
+	}
+	cmd.Printf("%s (%s %d):\n", rep.Kind, verb, len(rep.Planned))
+	for _, p := range rep.Planned {
+		cmd.Printf("  - %s\n", p)
+	}
+}
--- a/config/sample.config.yaml
+++ b/config/sample.config.yaml
@@ -179,12 +179,18 @@ app:
 
   # platform level administration
   admin:
-    # Email list of users which needs to be converted as superusers
-    # if the user is already present in the system, it is promoted to su
-    # if not, a new account is created with provided email id and promoted to su.
-    # UUIDs/slugs of existing users can also be provided instead of email ids
-    # but in that case a new user will not be created.
-    users: []
+    # bootstrap seeds a superuser SERVICE ACCOUNT from config (a username/password-style
+    # client_id + client_secret) so automation like the GitOps reconcile flow has a
+    # guaranteed superuser identity without a chicken-and-egg. The account is ensured and
+    # promoted to superuser on every boot (idempotent); the secret is rotated if it changes
+    # here. Authenticate with: Authorization: Basic base64(client_id:client_secret).
+    # Leave client_id/client_secret empty to disable.
+    # client_id must be a UUID (it is the service-account credential id); generate one
+    # (e.g. uuidgen) and keep it stable. client_secret is your chosen password.
+    bootstrap:
+      client_id: ""
+      client_secret: ""
+      # title: "GitOps Bootstrap Superuser"
   # smtp configuration for sending emails
   mailer:
     smtp_host: smtp.example.com
--- a/config/sample.config.yaml
+++ b/config/sample.config.yaml
@@ -184,7 +184,8 @@ app:
     # guaranteed superuser identity without a chicken-and-egg. The account is ensured and
     # promoted to superuser on every boot (idempotent); the secret is rotated if it changes
     # here. Authenticate with: Authorization: Basic base64(client_id:client_secret).
-    # Leave client_id/client_secret empty to disable.
+    # Leave client_id/client_secret empty to disable — but then no superuser is seeded and,
+    # since granting one is itself superuser-gated, none can be created via the API either.
     # client_id must be a UUID (it is the service-account credential id); generate one
     # (e.g. uuidgen) and keep it stable. client_secret is your chosen password.
     bootstrap:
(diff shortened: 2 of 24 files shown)


internal/bootstrap/bootstrapuser.go
```diff
@@ -0,0 +1,199 @@
...
+	// client_id is the service-user credential's id, which is a UUID column; a
+	// non-UUID would otherwise fail with an opaque SQL error deep in the lookup.
+	if _, err := uuid.Parse(clientID); err != nil {
+		return fmt.Errorf("bootstrap superuser: client_id must be a valid uuid: %w", err)
+	}
+
+	cred, err := creds.Get(ctx, clientID)
+	switch {
+	case errors.Is(err, serviceuser.ErrCredNotExist):
+		return createBootstrapSuperUser(ctx, logger, cfg, clientID, users, creds, promoter)
```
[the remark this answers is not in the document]
rohilsurana: Not fixing — this path is actually self-healing in our design.

The bootstrap service-user is created with a fixed, well-known id (`schema.BootstrapServiceUserID`), and the repository's `Create` **upserts** on id conflict (`OnConflict … DoUpdate`). So after any partial failure (a rotate where delete succeeds but recreate fails, or a create whose rollback-delete also fails), the next restart re-enters `createBootstrapSuperUser`: `users.Create(<sentinel>)` upserts the existing row (no error, no duplicate), then recreates the missing credential and re-promotes. State converges automatically — no manual repair.

The `ErrCredNotExist` → create path is intentionally the same whether or not the SA row already exists, precisely because the create is idempotent on the fixed id.

internal/bootstrap/bootstrapuser.go
```diff
@@ -0,0 +1,199 @@
...
+// it changes. Leave client_id/client_secret empty to disable.
+type SuperUserBootstrapConfig struct {
+	ClientID     string `yaml:"client_id" mapstructure:"client_id"`
+	ClientSecret string `yaml:"client_secret" mapstructure:"client_secret"`
+	Title        string `yaml:"title" mapstructure:"title"`
+}
+
+func (c SuperUserBootstrapConfig) enabled() bool {
+	return strings.TrimSpace(c.ClientID) != "" && c.ClientSecret != ""
+}
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837` — `ensureBootstrapSuperUser` now no-ops only when both `client_id` and `client_secret` are empty, and returns an error when exactly one is set. Added a unit test for the half-configured case.

internal/reconcile/platformuser_reconciler.go
```diff
@@ -0,0 +1,159 @@
...
+			Relations: relationsFromMetadata(u.GetMetadata()),
+		})
+	}
+	for _, su := range resp.Msg.GetServiceusers() {
+		// The config-bootstrapped break-glass SA has a fixed, well-known id; it is
+		// server-managed (seeded at boot, removal-guarded), so never reconcile — and
+		// so never try to remove — it.
+		if su.GetId() == schema.BootstrapServiceUserID {
+			continue
+		}
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837` — `validateSpec` now rejects a desired entry whose `ref` is the bootstrap SA (`schema.BootstrapServiceUserID`), so it can never be added via reconcile, not just skipped on the current side. Covered by a diff test.

internal/reconcile/platformuser.go
```diff
@@ -0,0 +1,150 @@
...
+	principalTypeServiceUser = "serviceuser"
+)
+
+// PlatformUserSpec is one desired platform-user entry from the YAML spec.
+// Role maps directly to the platform relation (admin -> superuser, member -> check);
+// the role label and the relation name are the same string ("admin" / "member").
+type PlatformUserSpec struct {
+	Type string `yaml:"type"` // "user" | "serviceuser"
+	Ref  string `yaml:"ref"`  // email/uuid/slug for a user; id for a service user
+	Role string `yaml:"role"` // "admin" | "member"
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837` by tightening the contract rather than plumbing slug: a user `ref` is now documented as email or uuid (slug dropped), which matches `specMatchesPrincipal` (id/email only). Plumbing the login/slug field from `ListPlatformUsers` is a larger change left out of scope.

internal/reconcile/platformuser.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+		// add desired relations not already held
+		for _, rel := range platformRelationOrder {
+			if _, wanted := want[rel]; !wanted {
+				continue
+			}
+			if _, has := p.Relations[rel]; !has {
+				ops = append(ops, Op{Action: opAdd, Type: p.Type, Ref: p.ID, Relation: rel})
+			}
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837` — `diffPlatformUsers` now collects adds and removes separately and returns adds first (`append(adds, removes...)`), so a sequential apply never leaves a principal with no platform access mid-transition. The order-sensitive diff tests were updated.

internal/reconcile/reconcile.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		rec, ok := registry[doc.Kind]
+		if !ok {
+			return reports, fmt.Errorf("no reconciler registered for kind %q", doc.Kind)
+		}
+		specBytes, err := yaml.Marshal(&doc.Spec)
+		if err != nil {
+			return reports, fmt.Errorf("marshal spec for kind %q: %w", doc.Kind, err)
+		}
+		rep, err := rec.Reconcile(ctx, specBytes, dryRun)
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837`, but not via `*yaml.Node`: switching `Spec` to a pointer breaks the spec marshal (the node serializes to `null`, dropping the spec — verified). Instead I keep `Spec yaml.Node` and reject when the marshaled spec is empty/`null`, which catches an absent or null `spec:` (the `spce:` typo) while still allowing an intentional `spec: []`. Added `reconcile_test.go` covering both.

test/e2e/testbench/testbench.go
```diff
@@ -134,6 +139,12 @@ func Init(appConfig *config.Frontier) (*TestBench, error) {
 
 	// let frontier start
 	time.Sleep(time.Second * 2)
+
+	// seed the platform superuser the production way: the config-bootstrapped service
+	// account grants the test admin superuser via the platform API (no boot-time human flow).
+	if err := PromoteBootstrapAdmin(context.Background(), te.AdminClient, OrgAdminEmail); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
rohilsurana: Addressed in `b4551837` — the promotion-failure path now returns `errors.Join(err, te.Close())`.

internal/api/v1beta1connect/serviceuser.go
```diff
@@ -168,11 +169,31 @@ func (h *ConnectHandler) CreateServiceUser(ctx context.Context, request *connect
...
 
+// errBootstrapSAImmutable returns a PermissionDenied error when id is the
+// config-bootstrapped break-glass SA (well-known id). That account is managed
+// only via app.admin.bootstrap at boot — the API must never delete it or mint
+// credentials/keys/tokens for it (which would create a persistent, rotation-proof
+// superuser backdoor). Not even platform superusers are allowed.
+//
+// The response is the generic permission error so it doesn't reveal that this id
+// is the protected SA; the specific reason is logged only.
+func errBootstrapSAImmutable(ctx context.Context, id string) error {
```
whoAbhishekSah, line 180: This guard runs on delete and on the handlers that create new credentials, tokens, and keys. But it doesn't run on `DeleteServiceUserCredential`, `DeleteServiceUserToken`, or `DeleteServiceUserJWK`. So a superuser could delete the SA's current credential or token. They can't create a new one, and the next boot re-creates it, so it recovers on restart — but until then the account can't log in. Might be worth adding the same guard to those three deletes, or noting the short lockout is fine.

cmd/reconcile.go
```diff
@@ -0,0 +1,76 @@
...
+func printReconcileReport(cmd *cli.Command, rep reconcile.Report) {
+	if len(rep.Planned) == 0 {
+		cmd.Printf("%s: no changes\n", rep.Kind)
+		return
+	}
+	verb := "applied"
+	if rep.DryRun {
+		verb = "planned"
+	}
+	cmd.Printf("%s (%s %d):\n", rep.Kind, verb, len(rep.Planned))
```
whoAbhishekSah: When it's not a dry run, this prints `len(rep.Planned)`, so if an apply fails partway the output still says e.g. `applied 5` when only 2 went through (then the error prints). `rep.Applied` already holds the real count, so printing that instead would keep the number right on partial failures.

internal/reconcile/platformuser.go
```diff
@@ -0,0 +1,162 @@
...
+// specMatchesPrincipal reports whether a desired spec refers to a current principal.
+// Users match by id or email; service users by id.
+func specMatchesPrincipal(s PlatformUserSpec, p platformPrincipal) bool {
+	if s.Type != p.Type {
+		return false
+	}
+	if s.Ref == p.ID {
+		return true
+	}
+	return p.Type == principalTypeUser && s.Ref != "" && strings.EqualFold(s.Ref, p.Email)
```
whoAbhishekSah, line 103: Edge case: if a desired entry matches a user by email but `ListPlatformUsers` returns an empty email for that user, the entry looks new (add-by-email, which no-ops since they're already admin) and the current principal matches nothing, so its relation gets removed. Adds run before removes, so the admin ends up stripped. This only happens if the list returns a user with no email, which normally shouldn't happen — flagging in case it's worth a guard.

whoAbhishekSah (commented):
Three small notes, none blocking.

internal/api/v1beta1connect/platform.go
```diff
@@ -55,6 +56,16 @@ func (h *ConnectHandler) RemovePlatformUser(ctx context.Context, req *connect.Re
 			}
 		}
 	} else if req.Msg.GetServiceuserId() != "" {
+		// Protect the config-bootstrapped break-glass SA (well-known id). It is
+		// seeded and managed at boot, not via this API, while reconcile is
+		// authoritative over service accounts — without this guard an apply (or a
+		// stray call) would strip its superuser access until the next restart.
+		// Respond with the generic permission error (don't reveal that this id is the
+		// protected SA); the specific reason is logged only.
```
whoAbhishekSah: Can we simplify this comment with a less loaded sentence?

config/sample.config.yaml
```diff
@@ -179,12 +179,18 @@ app:
...
-    users: []
+    # bootstrap seeds a superuser SERVICE ACCOUNT from config (a username/password-style
+    # client_id + client_secret) so automation like the GitOps reconcile flow has a
+    # guaranteed superuser identity without a chicken-and-egg. The account is ensured and
+    # promoted to superuser on every boot (idempotent); the secret is rotated if it changes
+    # here. Authenticate with: Authorization: Basic base64(client_id:client_secret).
+    # Leave client_id/client_secret empty to disable.
+    # client_id must be a UUID (it is the service-account credential id); generate one
+    # (e.g. uuidgen) and keep it stable. client_secret is your chosen password.
+    bootstrap:
```
whoAbhishekSah, line 192: So if this is not set, superusers cannot be created via the API?

whoAbhishekSah (approved):
Added some comments. Approving.

config/sample.config.yaml
rohilsurana: Yes, this is a minimum requirement for setup from now on.

internal/api/v1beta1connect/serviceuser.go
rohilsurana: Fixed in `84b818db`. Added the guard to all three deletes via a single choke point — `serviceuser.Service.DeleteKey` (which `DeleteSecret`/`DeleteToken`/`DeleteKey` all funnel through): it resolves the credential owner and returns `ErrProtected` if it is the bootstrap SA, which the handlers map to a generic `permission_denied` (with the reason logged, not leaked). Boot-time rotation is unaffected — it deletes via the repository directly, not the service. So the SA can no longer be locked out even transiently.

cmd/reconcile.go
rohilsurana: Fixed in `84b818db` — `printReconcileReport` now prints `rep.Applied` on apply (and `len(rep.Planned)` only for dry-run), so a partial apply reports the real count.

internal/reconcile/platformuser.go
rohilsurana: Leaving as-is (option B). This only triggers if `ListPlatformUsers` returns a user with an empty email, which shouldn’t happen — the list carries the DB email. And it is caught before any harm: reconcile posts a dry-run plan on every PR, so the spurious `add <email>` + `remove <id>` pair would be visible and reviewed before an apply. A guard would add complexity for a can’t-happen-in-practice case.

internal/api/v1beta1connect/platform.go
rohilsurana: Simplified in `84b818db` to: "The bootstrap SA is server-managed (seeded at boot), so it cant be removed via this API. Return the generic permission error and log the reason — the response must not reveal that this id is the protected SA."

config/sample.config.yaml
rohilsurana: Correct. With `app.admin.bootstrap` unset there is no seeded superuser, and since granting a superuser is itself superuser-gated (and `app.admin.users` is removed in this PR), there is no way to create the first one via the API — that is the chicken-and-egg the bootstrap SA solves. Clarified this in the config comment in `84b818db`.

PR #1723, opened 2026-07-01
fix(orgusers): support case-insensitive name filtering

Shreyag02:
## Summary
Fixes "unable to filter by name" on the organization users list. The non-rolefilter builder did not support the `ilike`/`notilike` operators, so a name filter sent as `ilike` was rejected with `wrapBadOperatorError` and the filter silently failed. This adds `ilike`/`notilike` support and, while there, hardens operator handling by routing on the column's declared datatype so string-only operators can no longer be applied to the datetime `org_joined_at` column.

## Changes
- Add `ilike`/`notilike` operator support for string columns.
- Route non-role filters by the column's declared datatype (`rql.GetDataTypeOfField`) instead of one combined operator list.
- Restrict the datetime `org_joined_at` column to `eq/neq/gt/lt/gte/lte`; any other operator (including `like`/`ilike`/`in`/`empty`) now returns `wrapBadOperatorError`.
- Add tests for `ilike`/`notilike`, a valid datetime operator (`gte`), and `ilike` rejected on the datetime column.
  
## Technical Details
`buildNonRoleFilterCondition` looks up the field's datatype up front:
- **datetime** columns accept only `eq/neq/gt/lt/gte/lte`.
- **string** columns accept `eq/neq/like/ilike/in/notin/notlike/notilike/empty/notempty`.

`like`/`notlike` wrap the value in `%...%` themselves; `ilike`/`notilike` expect the caller to include the wildcards (e.g. `%foo%`). The local operator lists mirror `salt/rql`'s upstream `validString/DatetimeOperations`, so the new branches are reachable from the API. Side effect: an unknown field name now fails fast via `GetDataTypeOfField` instead of falling through to the raw column name.

## Test Plan

- [x] Values flow through `?` placeholders, `goqu.Ex{}`, or `goqu.Record{}` — never `fmt.Sprintf` or `+` building a query that gets executed.
- [x] `ToSQL()` callers capture and forward params (`query, params, err := stmt.ToSQL(); db.…Context(ctx, …, query, params...)`). Never `query, _, err := …`.
- [x] No `?` placeholders inside single-quoted SQL literals in `goqu.L` (use `make_interval(hours => ?)`-style functions instead).
- [x] Any `//nolint:forbidigo` or `// #nosec G20x` annotation has a one-line justification on the same line that a reviewer can verify.

### DIFF
--- a/internal/store/postgres/org_users_repository.go
+++ b/internal/store/postgres/org_users_repository.go
@@ -325,15 +325,28 @@ func (r OrgUsersRepository) buildHasAnyRoleSubquery(orgID string) *goqu.SelectDa
 }
 
 func (r OrgUsersRepository) buildNonRoleFilterCondition(filter rql.Filter) (goqu.Expression, error) {
-	//this list will always be a subset of all RQL supported operators
-	supportedStringOperators := []string{"eq", "neq", "like", "ilike", "in", "notin", "notlike", "notilike", "empty", "notempty"}
-	supportedDateTimeOperators := []string{"eq", "neq", "gt", "lt", "gte", "lte"}
-
-	if !slices.Contains(supportedStringOperators, filter.Operator) && !slices.Contains(supportedDateTimeOperators, filter.Operator) {
-		return nil, wrapBadOperatorError(filter.Operator, filter.Name)
+	// Route by the column's declared datatype so that string-only operators
+	// (like/ilike/in/empty) are rejected on the datetime column org_joined_at.
+	datatype, err := rql.GetDataTypeOfField(filter.Name, svc.AggregatedUser{})
+	if err != nil {
+		return nil, err
 	}
 
 	columnName := r.getNonRoleColumnName(filter.Name)
+
+	if datatype == "datetime" {
+		supportedDateTimeOperators := []string{"eq", "neq", "gt", "lt", "gte", "lte"}
+		if !slices.Contains(supportedDateTimeOperators, filter.Operator) {
+			return nil, wrapBadOperatorError(filter.Operator, filter.Name)
+		}
+		return goqu.Ex{columnName: goqu.Op{filter.Operator: filter.Value.(string)}}, nil
+	}
+
+	// string columns
+	supportedStringOperators := []string{"eq", "neq", "like", "ilike", "in", "notin", "notlike", "notilike", "empty", "notempty"}
+	if !slices.Contains(supportedStringOperators, filter.Operator) {
+		return nil, wrapBadOperatorError(filter.Operator, filter.Name)
+	}
 
 	switch filter.Operator {
 	case "empty":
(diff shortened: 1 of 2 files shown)


internal/store/postgres/org_users_repository.go
```diff
@@ -343,11 +356,17 @@ func (r OrgUsersRepository) buildNonRoleFilterCondition(filter rql.Filter) (goqu
...
 		return goqu.Ex{columnName: goqu.Op{filter.Operator: strings.Split(filter.Value.(string), ",")}}, nil
 	case "like":
+		// case-insensitive match; wildcards are added here
 		searchPattern := "%" + filter.Value.(string) + "%"
 		return goqu.Cast(goqu.I(columnName), "TEXT").ILike(searchPattern), nil
 	case "notlike":
 		searchPattern := "%" + filter.Value.(string) + "%"
 		return goqu.Cast(goqu.I(columnName), "TEXT").NotILike(searchPattern), nil
+	case "ilike":
+		// case-insensitive match; the frontend already includes the wildcards (e.g. "%foo%")
```
whoAbhishekSah: Why not follow the same pattern? In one case, the frontend adds `%..%`, and in another, it does not. We should keep it consistent.

internal/store/postgres/org_users_repository.go
```diff
@@ -343,11 +356,17 @@ func (r OrgUsersRepository) buildNonRoleFilterCondition(filter rql.Filter) (goqu
 	case "in", "notin":
 		return goqu.Ex{columnName: goqu.Op{filter.Operator: strings.Split(filter.Value.(string), ",")}}, nil
 	case "like":
+		// case-insensitive match; wildcards are added here
 		searchPattern := "%" + filter.Value.(string) + "%"
 		return goqu.Cast(goqu.I(columnName), "TEXT").ILike(searchPattern), nil
```
whoAbhishekSah: Like also uses iLIKE only.

internal/store/postgres/org_users_repository.go
Shreyag02: Both the `ilike` operator and the `%..%` wildcards are added by Apsara's DataTable filter transform before the request leaves the browser — not by our app code.

In `@raystack/apsara` (`components/data-table/utils/`), `transformToDataTableQuery` builds each filter via two helpers (`filter-operations.tsx`):

- **`getFilterOperator`** maps the UI string operators to `ilike`:
  ```ts
  if (filterType === FilterType.string &&
      (operator === 'contains' || operator === 'starts_with' || operator === 'ends_with')) {
    return 'ilike';
  }
  ```
- **`getFilterValue` → `handleStringBasedTypes`** bakes the wildcards into the value:
  ```ts
  if (operator === 'contains')     processedValue = `%${stringVal}%`;
  else if (operator === 'starts_with') processedValue = `${stringVal}%`;
  else if (operator === 'ends_with')   processedValue = `%${stringVal}`;
  else if (operator === 'ilike') { if (!stringVal.includes('%')) processedValue = `%${stringVal}%`; }
  ```

So for the org-users name filter the client always sends `{ operator: "ilike", value: "%foo%" }` — never `like`. That's why the `ilike` branch here treats wildcards as caller-provided.

You're right that it's inconsistent, though: `like`/`ilike` both compile to SQL `ILIKE` (via goqu `.ILike()`), and since the real caller only ever emits `ilike` with wildcards pre-added, the `like`/`notlike` branches that wrap server-side are effectively dead for this list. I'll align them to one convention — keeping `ilike`/`notilike` (caller owns the wildcards) and dropping the server-side wrapping in `like`/`notlike` — so there's a single, consistent path.

internal/store/postgres/org_users_repository.go
Shreyag02: Correct — goqu's `.ILike()` emits SQL `ILIKE`, so `like` here is already case-insensitive. The only real difference between the `like` and `ilike` branches is who owns the `%..%` wildcards (server-side for `like`, caller-side for `ilike`). See my note on the `ilike` case below — Apsara only ever sends `ilike` with wildcards baked in, so I'll collapse these to one convention.

Shreyag02: Good catch — fixed. I aligned this file to the same convention `org_pats_repository` and `org_billing_repository` already use (the two other repos that own wildcards caller-side + cast to TEXT):

```go
case "like":     return goqu.Cast(goqu.I(columnName), "TEXT").Like(filter.Value.(string)), nil     // case-sensitive
case "notlike":  return goqu.Cast(goqu.I(columnName), "TEXT").NotLike(filter.Value.(string)), nil
case "ilike":    return goqu.Cast(goqu.I(columnName), "TEXT").ILike(filter.Value.(string)), nil    // case-insensitive
case "notilike": return goqu.Cast(goqu.I(columnName), "TEXT").NotILike(filter.Value.(string)), nil
```

So `like` now maps to real SQL `LIKE` (case-sensitive) instead of `ILIKE`, and `ilike` stays `ILIKE`. No more silent case-insensitivity on `like`.

internal/store/postgres/org_users_repository.go
Shreyag02: Done — wildcards are now caller-owned for **all four** operators (`like`/`notlike`/`ilike`/`notilike`), so there is a single consistent convention. The server no longer wraps `%..%` on `like`/`notlike`; the caller (Apsara DataTable, which only ever emits `ilike` with wildcards pre-baked) owns them in every case. This matches `org_pats_repository`/`org_billing_repository`.

Updated the `like`/`notlike` tests accordingly to pass pre-wrapped values.

Shreyag02:
@coderabbitai Addressed the datatype-routing finding. `buildNonRoleFilterCondition` now routes by the field's declared datatype again via `rql.GetDataTypeOfField(filter.Name, svc.AggregatedUser{})`:

- `org_joined_at` (tagged `type=datetime`) only accepts `eq/neq/gt/lt/gte/lte`; string operators (`like`/`ilike`/`in`/`empty`/…) are rejected with `wrapBadOperatorError` instead of generating bogus SQL against `policies.created_at`.
- String columns keep the full string-operator set.

Restored the `org_joined_at` + `ilike` rejection test and added an `org_joined_at` + `empty` rejection test.

whoAbhishekSah (approved):

PR #1735, opened 2026-07-03
fix(admin): add skeleton loader to invite-user Org select

Shreyag02:
## Summary
The Organization select in the Users-list "Invite user" dialog did not render a skeleton loader while its data was loading. This PR adds the skeleton (driven by the real query loading state) and applies a set of consistency cleanups across the admin invite dialogs and the client SDK invite dialog.

Following the review discussion, the previous "disable the submit button" pattern has been replaced with schema-level validation and visible field-level error messages. The submit button no longer silently disables; instead, missing required fields and malformed emails surface inline messages through the Field component.

## Changes
- Add a skeleton loader to the Org select (and the Role/Email fields) in the Users-list invite dialog, gated on the real query loading state rather than a hardcoded flag.
- Replace the disabled-submit pattern with strict schema validation and visible field-level error messages.
- Migrate the invite dialogs' form groups to the apsara Field composite (label + error + aria-invalid), replacing the custom Label + error Text markup.
- Always render the Emails Field (skeleton-swapping only the inner control) so its label no longer pops in after loading, consistent with the Role/Org fields.
- Remove the now-unused invite-users-dialog-label and form-error-message CSS classes from invite-users.module.css.
- Apply the same Field migration and validation approach to the org-Members "Invite user" dialog and the client SDK invite-member-dialog for consistency.

## Technical Details
- Replaced a hardcoded const isLoading = true (which left the dialog stuck on skeletons) with isLoading = isRolesLoading || isOrganizationsLoading.
- Validation is now enforced by the form schema:
  - Admin dialogs (zod): role and organizationId are required; emails is required and each address is validated.
  - Client SDK (yup): type (role) is required; emails is required with a .test that validates every comma-separated address.
- Errors render inline via Field; the submit button is no longer gated on form values. disabled is retained only while the roles/orgs queries are loading (when the fields are skeletons), so it never blocks a completed form.
- The Members dialog intentionally has no skeleton — its roles are provided synchronously by OrganizationContext, so there is no loading state to gate on.
- The shared invite-users-dialog-label and form-error-message CSS classes are retained in the Members dialog's stylesheet because add-tokens-dialog.tsx still uses them.

## Test Plan

N/A

### DIFF
--- a/web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
+++ b/web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
@@ -29,11 +29,21 @@ import { useTerminology } from '../../../../hooks/useTerminology';
 import { handleConnectError } from '~/utils/error';
 
 const inviteSchema = z.object({
-  role: z.string(),
+  role: z.string().min(1, { message: 'Role is required' }),
   emails: z
     .string()
-    .transform(value => value.split(',').map(str => str.trim()))
-    .pipe(z.array(z.string().email()))
+    .min(1, { message: 'Email is required' })
+    .transform(value =>
+      value
+        .split(',')
+        .map(str => str.trim())
+        .filter(str => str.length > 0)
+    )
+    .pipe(
+      z
+        .array(z.string().email({ message: 'Enter valid email address(es)' }))
+        .min(1, { message: 'Email is required' })
+    )
 });
 
 type InviteSchemaType = z.infer<typeof inviteSchema>;
@@ -109,25 +119,6 @@ export const InviteUsersDialog = ({ onOpenChange }: InviteUsersDialogProps) => {
   const isSubmitting = methods?.formState?.isSubmitting;
   const errors = methods?.formState?.errors;
 
-  const values = methods.watch(['emails', 'role']);
-
-  const isDisabled = useMemo(() => {
-    const [emails, role] = values;
-    const emailValue = Array.isArray(emails)
-      ? emails.join(',')
-      : ((emails as string) ?? '');
-    const emailList = emailValue
-      .split(',')
-      .map(e => e.trim())
-      .filter(str => str.length > 0);
-    return (
-      emailList.length <= 0 ||
-      !role ||
-      isSubmitting ||
-      Boolean(errors?.emails)
-    );
-  }, [isSubmitting, values, errors?.emails]);
-
   return (
     <Dialog open onOpenChange={onOpenChange}>
       <Dialog.Content width={600}>
@@ -194,7 +185,6 @@ export const InviteUsersDialog = ({ onOpenChange }: InviteUsersDialogProps) => {
               <Button
                 data-test-id="invite-users-invite-button"
                 type="submit"
-                disabled={isDisabled}
                 loading={isSubmitting}
                 loaderText="Inviting..."
               >
--- a/web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
+++ b/web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
@@ -1,10 +1,9 @@
 import {
   Button,
   Dialog,
+  Field,
   Flex,
-  Label,
   Select,
-  Text,
   TextArea,
   toastManager
 } from '@raystack/apsara';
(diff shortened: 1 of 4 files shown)


web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
```diff
@@ -109,6 +108,26 @@ export const InviteUsersDialog = ({ onOpenChange }: InviteUsersDialogProps) => {
...
+      .map(e => e.trim())
+      .filter(str => str.length > 0);
+    return (
+      emailList.length <= 0 ||
+      !role ||
+      isSubmitting ||
+      Boolean(errors?.emails)
+    );
+  }, [isSubmitting, values, errors?.emails]);
+
```
rohanchkrabrty: Let's avoid this. Making a button disabled without showing why is a bad UX. 

We can update and make the zod schema strict if needed, for cases like required and validation

web/sdk/admin/views/users/list/invite-users.tsx
```diff
@@ -89,13 +90,34 @@ export const InviteUser = () => {
...
+      .filter(str => str.length > 0);
+    return (
+      emailList.length <= 0 ||
+      !role ||
+      !organizationId ||
+      isSubmitting ||
+      Boolean(errors?.emails)
+    );
+  }, [isSubmitting, values, errors?.emails]);
+
```
rohanchkrabrty: Same reason as previous comment

web/sdk/admin/views/users/list/invite-users.tsx
```diff
@@ -149,46 +171,41 @@ export const InviteUser = () => {
...
-                  Invite as
-                </Label>
-                <Controller
-                  name="role"
-                  defaultValue={defaultRoleId}
-                  disabled={isRolesLoading}
-                  control={control}
-                  render={({ field, fieldState: { error } }) => {
-                    const { ref, ...rest } = field;
-                    if (isRolesLoading) return <Skeleton height={33} />;
```
rohanchkrabrty, line 189: There was skeleton loader already in place. What is this PR doing differently?

rohanchkrabrty (changes requested):

web/sdk/admin/views/users/list/invite-users.tsx
Shreyag02: The admin invite dialog previously used `react-loading-skeleton` per-field with a fixed 33px height (mismatched to the ~36px controls) and had **no** skeleton on the Emails field at all. This PR replaces that with apsara's `Skeleton` + `Field` gated on a single `isLoading` flag.

The whole change — the unified skeleton loading and the `isDisabled` submit gating — is matched 1:1 to our already-shipped client SDK dialog at `web/sdk/client/views/members/components/invite-member-dialog.tsx`, so the two invite flows stay consistent rather than diverging.

On the disabled-button UX concern: happy to move to a stricter zod schema with visible validation messages instead if you'd prefer — but note the client `invite-member-dialog` also disables the submit button via `isDisabled`, so we'd want to decide whether to change both for consistency.

web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
Shreyag02: This mirrors the already-shipped client SDK dialog `web/sdk/client/views/members/components/invite-member-dialog.tsx`, which also gates the submit button via an `isDisabled` derived from watched values — so the intent here was to keep the two invite flows consistent.

That said, I take the UX point. If we'd rather show *why* the button is blocked, I can switch to a stricter zod schema (`required`/email validation) with visible `Field` error messages instead of silently disabling. Since the client dialog uses the disabled-button approach too, my preference is to change both together so they don't diverge — happy to go whichever way you prefer.

web/sdk/admin/views/users/list/invite-users.tsx
Shreyag02: Same as the reply on the other thread — this matches the `isDisabled` pattern already used in the client SDK's `invite-member-dialog.tsx`. Happy to move to strict zod validation with visible error messages instead, applied to both dialogs together for consistency, if you'd prefer that over disabling the button.

web/sdk/admin/views/organizations/details/layout/invite-users-dialog.tsx
rohanchkrabrty: Let's update the `sdk/client` to follow this, basically remove the disabled pattern

web/sdk/admin/views/users/list/invite-users.tsx
rohanchkrabrty: Understood, this works then

web/sdk/admin/views/users/list/invite-users.tsx
```diff
@@ -29,12 +28,22 @@ import { handleConnectError } from "~/utils/error";
 import { useTerminology } from "../../../hooks/useTerminology";
 
 const inviteSchema = z.object({
-  role: z.string(),
-  organizationId: z.string(),
+  role: z.string().min(1, { message: "Role is required" }),
+  organizationId: z.string().min(1, { message: "Organization is required" }),
```
rohanchkrabrty: Use the terminology function here in place of `Organization`

rohanchkrabrty (changes requested):

rohanchkrabrty (approved):

PR #1736, opened 2026-07-03
fix(webhooks): surface form validation errors and reset add-webhook  drawer state

Shreyag02:
## Summary
Fixes several issues in the Add Webhook flow.
The originally reported bug — missing input field borders — was already fixed; while verifying, we found the "Add Webhook" form silently failed to submit, retained stale state on reopen, and could dead-end on certain responses.
This PR addresses those.

## Changes
- Render inline validation errors in `CustomField` so failed submissions show why (previously invisible).
- Reset the create drawer form after a successful submission so a freshly opened drawer starts clean, while preserving in-progress input if the drawer is closed accidentally.
- Show an error toast when the create response contains no webhook, instead of silently doing nothing.
- Set proper form `defaultValues` to avoid the controlled/uncontrolled input warning.
- Fix the mismatched description length message ("Must be 10 or more characters" → 3, matching the `min(3)` rule) in both create and update.

## Technical Details
- The root cause of the "click does nothing / needs a second click" behavior was that `CustomField` only rendered Radix's native validity matchers (`valueMissing`, `typeMismatch`), which never fire for these constraint-less inputs. Validation is driven entirely by zod/react-hook-form, whose errors live in `formState.errors` and were never displayed — so an invalid URL or short description blocked `handleSubmit` with no feedback. Now uses `useFormState({ control, name })` to render `errors[name].message` inline.
- The create drawer stays permanently mounted (parent only toggles `open`), so the form was never reset between uses. Reset is now triggered only after a successful submission via a bare `methods.reset()` (which falls back to the configured `defaultValues`), rather than on open — this keeps a clean form for the next creation without discarding data if the user accidentally closes the drawer mid-entry. The `CustomField` fix also benefits the update drawer since it's shared.
- Error styling uses a CSS module (`custom-field.module.css`) rather than inline styles, matching the existing `PageHeader` convention.

## Test Plan

N/A

### DIFF
--- a/web/sdk/admin/components/CustomField.tsx
+++ b/web/sdk/admin/components/CustomField.tsx
@@ -16,7 +16,6 @@ import {
 } from "react-hook-form";
 import { capitalizeFirstLetter } from "../utils/helper";
 import Skeleton from "react-loading-skeleton";
-import styles from "./custom-field.module.css";
 
 type CustomFieldNameProps = {
   name: string;
@@ -183,7 +182,7 @@ export const CustomFieldName = ({
           )}
         </FormControl>
         {errorMessage ? (
-          <Text size="small" className={styles.errorMessage}>
+          <Text size="small" variant="danger">
             {errorMessage}
           </Text>
         ) : null}
--- a/web/sdk/admin/components/CustomField.tsx
+++ b/web/sdk/admin/components/CustomField.tsx
@@ -8,9 +8,15 @@ import {
 import { Flex, Select, Switch, Text, TextArea, Input } from "@raystack/apsara";
 import React, { CSSProperties } from "react";
 
-import { Control, Controller, UseFormRegister } from "react-hook-form";
+import {
+  Control,
+  Controller,
+  UseFormRegister,
+  useFormState,
+} from "react-hook-form";
 import { capitalizeFirstLetter } from "../utils/helper";
 import Skeleton from "react-loading-skeleton";
+import styles from "./custom-field.module.css";
 
 type CustomFieldNameProps = {
   name: string;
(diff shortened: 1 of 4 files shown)


web/sdk/admin/views/webhooks/webhooks/create/index.tsx
```diff
@@ -48,9 +48,27 @@ export default function CreateWebhooks({ open = false, onClose: onCloseProp }: C
...
+  useEffect(() => {
+    if (open) {
+      methods.reset({
+        url: "",
+        description: "",
+        state: false,
+        subscribed_events: [],
+      });
+    }
+  }, [open, methods.reset]);
```
rohanchkrabrty: Let's not reset on drawer close. If someone fills the form and accidently closes the drawer, all data will be lost. Let's reset only after successful submission. L86

Also since the form already has default values set, we don't need to pass them again in the `reset` call

rohanchkrabrty (changes requested):

web/sdk/admin/views/webhooks/webhooks/create/index.tsx
Shreyag02: it was not getting reset even after successful submission
let me cross-check the flow again and add a guardrail.

Shreyag02: updated to reset on successful submission.

web/sdk/admin/components/CustomField.tsx
```diff
@@ -174,6 +182,11 @@ export const CustomFieldName = ({
             />
           )}
         </FormControl>
+        {errorMessage ? (
+          <Text size="small" className={styles.errorMessage}>
```
rohanchkrabrty: The css is not needed. `<Text variant="danger">` does the same

rohanchkrabrty (approved):

PR #1738, opened 2026-07-07
fix(admin): fix Org Security loading state UI

Shreyag02:
## Summary
The Org → Security page had a buggy loading state — the "Allowed email domains" loader showed up as an unstyled empty box, and the whole section would vanish or behave oddly depending on the data. This PR fixes the loader and makes the loading, empty, and error states consistent and clear.

## Changes
- Switched the domains loader from `react-loading-skeleton` to Apsara's `Skeleton` (self-styled, no extra CSS needed).
- Loading, list, empty, and error states now all render inside the same bordered box, so nothing pops in or collapses.
- Added a friendly empty state ("No allowed email domains added yet") instead of rendering nothing.
- Added a distinct error state with a **Retry** button, so a failed fetch is no longer mistaken for an org with no domains.
  
## Technical Details
- The root cause: the default `react-loading-skeleton` `Skeleton` relies on `react-loading-skeleton/dist/skeleton.css`, which is never imported anywhere in `web/sdk/admin` or `web/apps/admin`. The app's `<SkeletonTheme>` only provides color variables, not the base skeleton styles — so the skeleton rendered as an invisible/empty box. Apsara's `Skeleton` ships its own styles, so it works out of the box (same approach as `session-skeleton.tsx` and `apis/details-dialog.tsx`).
- Previously the component had three branches (skeleton / rows / `null`). The `null` empty case made the bordered box disappear, causing the "abrupt close" jump. All four states (loading/list/empty/error) now share the same `.domains-list` container for a stable layout.
- The error state gets `error` + `onRetry` from the `listOrganizationDomains` query in `index.tsx`; Retry calls the query's `refetch()`. The existing error toast is kept.

## Test Plan
  - Loaded the Org → Security page and confirmed the domains loader shows a proper skeleton, then swaps into the list without any layout jump.
  - Verified an org with no domains shows the empty-state text instead of a blank/collapsing box.
  - Simulated an API failure (blocked the `ListOrganizationDomains` request in DevTools) and confirmed the error state + Retry button appear; Retry re-fetches and recovers.
- [x] Build and type checking passes (`tsc --noEmit` clean for the changed files)

N/A

### DIFF
--- a/web/sdk/admin/views/organizations/details/security/domains-list.tsx
+++ b/web/sdk/admin/views/organizations/details/security/domains-list.tsx
@@ -3,12 +3,12 @@ import {
   Button,
   Flex,
   IconButton,
+  Skeleton,
   Text,
   toastManager,
 } from "@raystack/apsara";
 import styles from "./security.module.css";
 import { CheckCircledIcon, TrashIcon } from "@radix-ui/react-icons";
-import Skeleton from "react-loading-skeleton";
 import { useContext, useState } from "react";
 import { useMutation, createConnectQueryKey, useTransport } from "@connectrpc/connect-query";
 import { useQueryClient } from "@tanstack/react-query";
@@ -23,7 +23,7 @@ interface DeleteDomainDialogProps {
 const DeleteDomainDialog = ({
   domain,
 }: DeleteDomainDialogProps) => {
-  const {organization} = useContext(OrganizationContext);
+  const { organization } = useContext(OrganizationContext);
   const [isDialogOpen, setIsDialogOpen] = useState(false);
   const queryClient = useQueryClient();
   const transport = useTransport();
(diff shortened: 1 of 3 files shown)


web/sdk/admin/views/organizations/details/security/domains-list.tsx
```diff
@@ -133,21 +133,49 @@ const DomainItem = ({ domain }: DomainItemProps) => {
...
+  error,
+  onRetry,
 }: DomainsListProps) => {
+  if (!isLoading && error) {
+    return (
+      <Flex
+        justify="between"
+        align="center"
+        className={styles["domains-list"]}
+        style={{ padding: "var(--rs-space-5) var(--rs-space-7)" }}
```
rohanchkrabrty: since there is already a classname, move the inline style to the css. If needed add another class

web/sdk/admin/views/organizations/details/security/domains-list.tsx
```diff
@@ -133,21 +133,49 @@ const DomainItem = ({ domain }: DomainItemProps) => {
...
+      >
+        <Text size="small" variant="secondary">
+          Couldn&apos;t load email domains
+        </Text>
+        <Button
+          variant="outline"
+          color="neutral"
+          size="small"
+          onClick={() => onRetry()}
```
rohanchkrabrty:
```suggestion
          onClick={onRetry}
```

web/sdk/admin/views/organizations/details/security/domains-list.tsx
```diff
@@ -133,21 +133,49 @@ const DomainItem = ({ domain }: DomainItemProps) => {
...
-          key={index}
-          height={20}
-          style={{ margin: "var(--rs-space-5) 0" }}
-        />
-      ))}
+    <Flex
+      justify="between"
+      align="center"
+      className={styles["domains-list"]}
+      style={{ padding: "var(--rs-space-5) var(--rs-space-7)" }}
```
rohanchkrabrty: since there is already a classname, move the inline style to the css. If needed add another class

web/sdk/admin/views/organizations/details/security/domains-list.tsx
```diff
@@ -158,5 +186,15 @@ export const DomainsList = ({
...
+    <Flex className={styles["domains-list"]}>
+      <Text
+        size="small"
+        variant="secondary"
+        style={{ padding: "var(--rs-space-5) 0" }}
+      >
+        No allowed email domains added yet
+      </Text>
+    </Flex>
+  );
```
rohanchkrabrty, line 193: Let's not have this. If there's no data, let's skip the container. We follow this pattern in `sdk/client` too

web/sdk/admin/views/organizations/details/security/index.tsx
```diff
@@ -87,6 +87,8 @@ export const OrganizationSecurity = () => {
         <DomainsList
           isLoading={isLoading}
           domains={domains ?? []}
+          error={error}
+          onRetry={() => refetch()}
```
rohanchkrabrty:
```suggestion
          onRetry={refetch}
```

rohanchkrabrty (changes requested):

web/sdk/admin/views/organizations/details/security/domains-list.tsx
Shreyag02: i got it verified from design team

rohanchkrabrty (approved):

PR #1739, opened 2026-07-08
refactor(group): resolve org from the group instead of the request

AmanGIT07:
## What changed

- Removed `org_id` from the request protos of `GetGroup`, `UpdateGroup`,
  `ListGroupUsers`, `RemoveGroupUser`, `EnableGroup`, `DisableGroup`,
  `DeleteGroup`, and `SetGroupMemberRole`. Each addresses a group by its id.
- These handlers resolve the org from the group record. The request no
  longer carries an org id.
- `GroupRepository.UpdateByID` no longer writes `org_id`. A group's org is
  fixed at creation and cannot change through an update.
- `GroupService.GetByIDs` takes an optional `Filter`. A new `IncludeDisabled`
  option lets `EnableGroup` find a group that is currently disabled.
- `group view` CLI takes `<group-id>` only.
- Regenerated proto from proton.

## Behavior

- A wrong or missing `org_id` on these calls used to return `OrgNotFound`.
  The field is gone. Clients that still send it are ignored (unknown JSON
  fields are dropped), so nothing breaks on the wire.
- Reads and updates of a group in a disabled org return `FailedPrecondition`.
  A read of a disabled group returns `NotFound`.
- `DeleteGroup` is no longer blocked when the group's org is disabled.
- `CreateGroup` and `ListOrganizationGroups` keep `org_id` — the org is a
  real input there.

## Testing

- Unit: connect handlers, core group, cmd.
- Repo (Docker): group repository, including the org-immutable update.
- e2e (Docker): `TestGroupAPI`, including a disable → enable round-trip.

## Depends on

- The matching proton change on branch `refactor/group-rpcs` (removes
  `org_id` from the group request messages). Merge proton first, then bump
  `PROTON_COMMIT` to the merged SHA.


core/group/service.go
```diff
@@ -90,8 +90,12 @@ func (s Service) Get(ctx context.Context, id string) (Group, error) {
 	return s.repository.GetByID(ctx, id)
 }
 
-func (s Service) GetByIDs(ctx context.Context, ids []string) ([]Group, error) {
-	return s.repository.GetByIDs(ctx, ids, Filter{})
+func (s Service) GetByIDs(ctx context.Context, ids []string, flt ...Filter) ([]Group, error) {
+	f := Filter{}
+	if len(flt) > 0 {
+		f = flt[0]
```
whoAbhishekSah, line 96: Is there a reason we are skipping the rest of the filters?

internal/api/v1beta1connect/group.go
```diff
@@ -100,6 +100,38 @@ func (h *ConnectHandler) ListOrganizationGroups(ctx context.Context, request *co
 	return connect.NewResponse(&frontierv1beta1.ListOrganizationGroupsResponse{Groups: groups}), nil
 }
 
+// ensureOrgEnabled fails with FailedPrecondition when the org is disabled.
+func (h *ConnectHandler) ensureOrgEnabled(ctx context.Context, orgID string) error {
```
whoAbhishekSah, line 104: Instead of returning an error when the org is disabled, why don't we return True/False based on the org's enabled status and an error for other cases? it will read much simpler.

internal/api/v1beta1connect/group.go
```diff
@@ -100,6 +100,38 @@ func (h *ConnectHandler) ListOrganizationGroups(ctx context.Context, request *co
...
+	grp, err := h.groupService.Get(ctx, groupID)
+	if err != nil {
+		switch {
+		case errors.Is(err, group.ErrNotExist), errors.Is(err, group.ErrInvalidID), errors.Is(err, group.ErrInvalidUUID):
+			return group.Group{}, connect.NewError(connect.CodeNotFound, ErrGroupNotFound)
+		default:
+			return group.Group{}, connect.NewError(connect.CodeInternal, fmt.Errorf("getGroupInEnabledOrg: group_id=%s: %w", groupID, err))
+		}
+	}
+	if err := h.ensureOrgEnabled(ctx, grp.OrganizationID); err != nil {
```
whoAbhishekSah, line 129: Why does the Group Get RPC checks for org status?

core/group/service.go
AmanGIT07: `flt ...Filter` is an optional single filter — it's a `[]Filter` where callers pass zero or one, not a list of lists. `GetByIDs(ctx, ids)` passes none; `EnableGroup` is the only caller that filters and passes exactly one. `flt[0]` is that one, so there's never a second to skip. Left it variadic so the existing caller doesn't change.

internal/api/v1beta1connect/group.go
AmanGIT07: Returning the error keeps the three call sites to one line (`if err := ensureOrgEnabled(...); err != nil { return nil, err }`). A `(bool, error)` would push the "disabled → FailedPrecondition" mapping into each caller, so I've kept it as-is.

internal/api/v1beta1connect/group.go
AmanGIT07: A disabled org shouldn't expose any of its resources, so reads gate on the org's state. This preserves the existing behavior — the old handler already returned `FailedPrecondition` for a disabled org via the request's `org_id`; we now derive the org from the group instead. A uniform version of this check across all resource RPCs is tracked separately.

test/e2e/regression/api_test.go
```diff
@@ -883,15 +882,13 @@ func (s *APIRegressionTestSuite) TestGroupAPI() {
...
-			Body:  &frontierv1beta1.GroupRequestBody{},
+			Id:   "",
+			Body: &frontierv1beta1.GroupRequestBody{},
 		}))
 		s.Assert().Equal(connect.CodeInvalidArgument, connect.CodeOf(err))
 	})
 	s.Run("7. group admin update a new team without group id should fail", func() {
 		_, err := s.testBench.Client.UpdateGroup(ctxOrgAdminAuth, connect.NewRequest(&frontierv1beta1.UpdateGroupRequest{
-			OrgId: myOrg.GetId(),
 			Body: &frontierv1beta1.GroupRequestBody{
```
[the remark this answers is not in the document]
AmanGIT07: Subtests 6 and 7 pass — the metadata-schema validation rejects their bodies with `InvalidArgument` before the id lookup. A valid body with an empty id returns `NotFound`, which matches the existing behavior (`UpdateByID` already mapped an empty id that way), so leaving it as-is.

whoAbhishekSah (approved):

PR #1741, opened 2026-07-08
fix: remove unused user and org relations from app/pat schema

### LINKED ISSUE #1740: Remove unused user and org relations from app/pat schema definition
## Summary

The `app/pat` definition in the base SpiceDB schema declares two relations, `user` and `org`, that are never used. We should remove them.

```
// internal/bootstrap/schema/base_schema.zed
definition app/pat {
	relation user: app/user
	relation org: app/organization
}
```

## Why they are unused

- **Nothing writes them.** When a PAT is created, the owning user and the org are saved as columns on the `pat` Postgres table (`user_id`, `org_id`). The code never creates an `app/pat#user` or `app/pat#org` tuple in SpiceDB. The only SpiceDB writes for a PAT are its role scopes, where the PAT is the subject (the bearer of a role).
- **Nothing reads them.** No permission in the schema refers to `pat->user` or `pat->org`.
- **Listing does not need them.** Listing PATs by user or by org is a plain SQL query on the `pat` table, not a SpiceDB lookup.

## Impact

None on access. These relations sit in no permission rule, so removing them cannot change any permission check. This is schema cleanup only.

whoAbhishekSah:
## What

The `app/pat` definition in the base SpiceDB schema declared two relations, `user` and `org`, that are never used. This removes them.

## Why they are unused

- **Nothing writes them.** When a PAT is created, its owning user and org are saved as columns on the `pat` Postgres table (`user_id`, `org_id`). The code never writes an `app/pat#user` or `app/pat#org` tuple. The only SpiceDB writes for a PAT are its role scopes, where the PAT is the subject (the bearer of a role).
- **Nothing reads them.** No permission refers to `pat->user` or `pat->org`.
- **Listing does not need them.** Listing PATs by user or org is a plain SQL query on the `pat` table.

## Impact

None on access. These relations sit in no permission rule, so removing them cannot change any permission check. `app/pat` stays valid as a subject type (like `app/user {}`), so it still appears as `app/pat:*` in role relations and as a `bearer` on rolebindings.

Closes #1740

### DIFF
--- a/internal/bootstrap/schema/base_schema.zed
+++ b/internal/bootstrap/schema/base_schema.zed
@@ -6,10 +6,7 @@ definition app/serviceuser {
     permission manage = org->serviceusermanage
 }
 
-definition app/pat {
-	relation user: app/user
-	relation org: app/organization
-}
+definition app/pat {}
 
 definition app/platform {
 	relation admin: app/user | app/serviceuser
--- a/internal/bootstrap/testdata/compiled_schema.zed
+++ b/internal/bootstrap/testdata/compiled_schema.zed
@@ -69,10 +69,7 @@ definition app/organization {
 	permission update = platform->superuser + granted->app_organization_administer + granted->app_organization_update + owner
 }
 
-definition app/pat {
-	relation org: app/organization
-	relation user: app/user
-}
+definition app/pat {}
 
 definition app/platform {
 	relation admin: app/user | app/serviceuser


whoAbhishekSah:
## Manual E2E test — PAT flow

Ran the full PAT lifecycle against a local server that already has this change live (`zed schema read` shows `definition app/pat {}`, matching the committed diff).

**Setup**
- Fresh org `pat1741-*`
- `patuser1` → org owner (created the org)
- `patuser2` → org viewer (added via `CreatePolicy`)
- Each user acted with its own session cookie

**Results**

| Area | Check | Result |
|------|-------|--------|
| Create | `CheckCurrentUserPATTitle` (free, then taken, case-insensitive) | ✅ true → false |
| Create | `CreateCurrentUserPAT` user1 (manager scope) + user2 (viewer scope) | ✅ tokens issued |
| Create | duplicate title | ✅ `409` conflict |
| Listing | `ListRolesForPAT` (org scope) | ✅ owner/manager/viewer returned |
| Listing | `SearchCurrentUserPATs` | ✅ each user sees only their own PAT |
| Get | `GetCurrentUserPAT` own | ✅ `used_at` populated after use |
| Auth | PAT bearer resolves to owning user | ✅ `GetCurrentUser` returns the owner |
| Scope: read | `GetOrganization` via both PATs | ✅ both `200` (both roles have `get`) |
| Scope: write | `UpdateOrganization`: manager vs viewer | ✅ `200` vs `403` |
| Update | `UpdateCurrentUserPAT` downgrade manager → viewer | ✅ same token loses `update` (`403`), keeps `get` (`200`) |
| Update | `RegenerateCurrentUserPAT` | ✅ token rotated, old → `401`, new → `200`, scope preserved |
| Isolation | user2 get/update/delete user1's PAT | ✅ all `404` |
| Delete | `DeleteCurrentUserPAT` | ✅ token → `401`, `Search` → 0 |
| SpiceDB | `bearer app/pat:<id>` rolebinding tuples present, removed on delete | ✅ |
| Auth | invalid token | ✅ `unauthenticated` |

**Conclusion**

No functional impact. The removed relations were not in any permission path. `app/pat` still works as a subject type — rolebindings carry `bearer app/pat:<id>`, scoped permissions enforce correctly (manager could update the org, viewer was denied), and create/update/regenerate/delete all cascade through SpiceDB as before.

AmanGIT07 (approved):

PR #1744, opened 2026-07-09
fix(invitation): auto-delete expired invitations on a schedule

whoAbhishekSah:
## The problem

Expired invitations were not getting cleaned up.

If you try to accept an expired invite, it just fails. If someone sends a new invite to the same person, it writes over the old one. Neither of these actually removes the expired invite.

So an invite that is never accepted and never deleted stays around forever. Each invite is stored in three places: one row in the `invitations` table, and two records in SpiceDB (one links the invite to the user, one links it to the org). All three keep piling up over time.

## The fix

Add a job that runs once a day, at midnight UTC, and deletes expired invites. This matches how the domain and session cleanup jobs already work. If the job ever crashes, the crash is caught so it can't take down the server.

The job does not delete an invite the moment it expires. It waits until the invite has been expired for 7 days. That way a recently expired invite still shows up in the list for a week, and the job cleans it up after that.

To delete, the job uses the existing `invitation.Delete`. That one call removes the table row and both SpiceDB records together.

This is the important part. The code already had a `GarbageCollect` function that deleted only the table row, with a plain SQL delete. That would have left the two SpiceDB records behind with nothing pointing to them. So I removed it and replaced it with a read-only `ListExpired` that just finds the old invites. The job then deletes each one the safe way.

If one invite fails to delete, the job logs it and moves on, so the rest still get cleaned up.

### DIFF
--- a/internal/store/postgres/invitation_repository.go
+++ b/internal/store/postgres/invitation_repository.go
@@ -244,10 +244,7 @@ func (s *InvitationRepository) Delete(ctx context.Context, id uuid.UUID) error {
 }
 
 // ListExpired returns every invitation whose expires_at is at or before
-// expiredBefore. The caller passes a cutoff in the past (now minus the
-// retention window) so recently expired invites are skipped, and deletes the
-// results through the invitation service so both the SpiceDB tuples and the
-// invitations row are removed
+// expiredBefore.
 func (s *InvitationRepository) ListExpired(ctx context.Context, expiredBefore time.Time) ([]invitation.Invitation, error) {
 	var fetchedInvitations []Invitation
 	query, params, err := dialect.From(TABLE_INVITATIONS).
--- a/internal/store/postgres/invitation_repository.go
+++ b/internal/store/postgres/invitation_repository.go
@@ -247,8 +247,7 @@ func (s *InvitationRepository) Delete(ctx context.Context, id uuid.UUID) error {
 // expiredBefore. The caller passes a cutoff in the past (now minus the
 // retention window) so recently expired invites are skipped, and deletes the
 // results through the invitation service so both the SpiceDB tuples and the
-// invitations row are removed together — a raw row delete here would leak the
-// invitation's #user / #org tuples behind.
+// invitations row are removed
 func (s *InvitationRepository) ListExpired(ctx context.Context, expiredBefore time.Time) ([]invitation.Invitation, error) {
 	var fetchedInvitations []Invitation
 	query, params, err := dialect.From(TABLE_INVITATIONS).
--- a/internal/store/postgres/invitation_repository.go
+++ b/internal/store/postgres/invitation_repository.go
@@ -243,27 +243,39 @@ func (s *InvitationRepository) Delete(ctx context.Context, id uuid.UUID) error {
 	})
 }
 
-// TODO(kushsharma): schedue a cron for it
-func (s *InvitationRepository) GarbageCollect(ctx context.Context) error {
-	query, params, err := dialect.Delete(TABLE_INVITATIONS).
+// ListExpired returns every invitation whose expires_at is in the past. The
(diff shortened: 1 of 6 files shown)


whoAbhishekSah:
## How this was tested

Ran locally against a real Frontier (Postgres + SpiceDB). Set the cleanup cron to
`@every 15s` so the sweep runs on demand, then checked the DB row and the SpiceDB
tuples after each step.

| # | What we did | Expected | Result |
|---|-------------|----------|--------|
| 1 | Create an invite | 1 DB row + 2 tuples (`#org`, `#user`) | ✅ |
| 2 | Invite expired 8 days ago (past the 7-day keep window) | swept: row + both tuples gone | ✅ |
| 3 | Invite expired 1 hour ago (still inside the keep window) | left alone, still listed | ✅ |
| 4 | Wait for another sweep with nothing old enough | nothing deleted, no error | ✅ |
| 5 | Unit tests (`go test ./core/invitation/...`) | pass | ✅ |
| 6 | `gofmt`, `go vet`, `go build` | clean | ✅ |

The invite is removed through `Delete`, so the row and both tuples go together —
the old row-only cleanup would have left the tuples behind. None were left after
the sweep.

AmanGIT07:
> ## How this was tested

@whoAbhishekSah Can you validate these scenarios too -
1. user invited once, the invitation hasn't expired and we re-invite the same user. 
2. user invite expired and cleaned up and we re-invite

AmanGIT07 (approved):

internal/store/postgres/invitation_repository.go
```diff
@@ -243,27 +243,41 @@ func (s *InvitationRepository) Delete(ctx context.Context, id uuid.UUID) error {
...
 
-// TODO(kushsharma): schedue a cron for it
-func (s *InvitationRepository) GarbageCollect(ctx context.Context) error {
-	query, params, err := dialect.Delete(TABLE_INVITATIONS).
+// ListExpired returns every invitation whose expires_at is at or before
+// expiredBefore. The caller passes a cutoff in the past (now minus the
+// retention window) so recently expired invites are skipped, and deletes the
+// results through the invitation service so both the SpiceDB tuples and the
+// invitations row are removed together — a raw row delete here would leak the
+// invitation's #user / #org tuples behind.
```
AmanGIT07: `row-only delete would leak the #user / #org tuples behind.` - this line is repeated multiple times, we can clean it up

whoAbhishekSah:
@AmanGIT07 Validated both re-invite scenarios locally against a real Frontier (Postgres + SpiceDB). Checked the `invitations` row and both SpiceDB tuples (`#user`, `#org`) after each step.

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | Invite a user, then re-invite the same user while the invite is still active | Same invite is reused: 1 row, 2 tuples, no duplicate | ✅ Same invite ID reused. Still 1 row, expiry unchanged (not extended), still 2 tuples |
| 2 | Invite a user, clean up the invite, then re-invite | Fresh invite: 1 row, 2 tuples, no leftovers | ✅ Cleanup removed the row and both tuples. Re-invite created a new invite ID with 1 row and 2 tuples |

For scenario 2 the cleanup was done through the same `invitation.Delete` the cron runs per invite (via `DeleteOrganizationInvitation`), so the row and both tuples go together. The earlier comment already covered the cron picking the right invites to sweep.

### One extra case worth flagging

Re-inviting a user whose invite has **expired but is still inside the 7-day keep window** (not swept yet) creates a **second** row plus 2 new tuples, and leaves the old expired row in place. So for a short time there are 2 rows and 4 tuples for the same user in the same org.

This is not new. `Create` only reuses an invite when it has not expired, and the upsert keys on `id` only (there is no unique constraint on org + user), so this has always happened. This PR actually makes it better: the stale duplicate now gets swept 7 days after it expires instead of staying forever.

PR #1745, opened 2026-07-09
fix: wait for HTTP server drain during graceful shutdown

AmanGIT07:
## Summary
Graceful shutdown now finishes draining in-flight requests before the
process tears down its dependencies.

## Changes
- `ServeConnect` tracks its shutdown goroutines with a `sync.WaitGroup`
  and waits for them after `ListenAndServe` returns.
- A failed `server.Shutdown` returns early instead of also logging the
  "shutdown complete" message.
- The metrics server shutdown gets the same grace period as the connect
  server, and its error is logged instead of dropped.

## Technical Details
`http.Server.Shutdown` closes the listener first and drains active
requests afterwards, so `ListenAndServe` returns before the drain
finishes. `ServeConnect` now waits for the shutdown goroutines, so it
only returns once requests are done.

## Test Plan
- [x] Added `TestServeConnectReturnsAfterShutdownOnContextCancel`, runs
  the real server with and without the metrics listener
- [x] `go test -race ./pkg/server/...` passes
- [x] `golangci-lint run pkg/server/` reports 0 issues

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -266,6 +266,8 @@ func ServeConnect(ctx context.Context, logger *slog.Logger, cfg Config, deps api
 
 	// Start server
 	if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		// No shutdownWG.Wait here: ctx is never cancelled on this path, so the
+		// shutdown goroutines only unblock when the process exits right after.
 		return fmt.Errorf("connect server failed: %w", err)
 	}
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4,16 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
-
-	"log/slog"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/raystack/salt/server/spa"
@@ -208,6 +208,9 @@ func ServeConnect(ctx context.Context, logger *slog.Logger, cfg Config, deps api
 		Handler: handler,
 	}
 
+	// counts shutdown goroutines still draining their servers
+	var shutdownWG sync.WaitGroup
+
 	// start dedicated metrics server if configured
 	if cfg.MetricsPort > 0 {
 		metricsMux := http.NewServeMux()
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -253,6 +269,11 @@ func ServeConnect(ctx context.Context, logger *slog.Logger, cfg Config, deps api
 		return fmt.Errorf("connect server failed: %w", err)
 	}
 
+	// Shutdown closes the listener first, which makes ListenAndServe return
+	// while in-flight requests are still draining. Wait for the drain to
+	// finish so callers don't tear down the database and other dependencies
+	// under requests that are still running.
+	shutdownWG.Wait()
```
rohilsurana, line 267: This wait only helps HTTP/1.1 requests. The server wraps the mux in `h2c.NewHandler(mux, &http2.Server{})`. h2c hijacks the connection and hands it to `http2.Server.ServeConn`. The `net/http` docs say `Shutdown` does not wait for or close hijacked connections.

I checked x/net v0.44.0 (the version in go.mod). The hook that makes shutdown send GOAWAY to HTTP/2 connections is only set up inside `http2.ConfigureServer`, and this code never calls it. So gRPC clients get no GOAWAY and no wait (gRPC needs HTTP/2, and over plain TCP that means h2c). Their requests can still get cut when the process shuts down.

Two suggestions:
- Call `http2.ConfigureServer(server, h2s)` with the same `http2.Server` you pass to `h2c.NewHandler`. Then `Shutdown` sends GOAWAY, and h2c clients stop opening new streams. This is the known workaround from golang/go#26682. It still does not wait for running streams, but it is much better than nothing.
- The comment above this line promises more than the code does: the wait covers HTTP/1.1 requests, not hijacked h2c connections. It would be accurate as something like: "Wait for Shutdown to finish draining HTTP/1.1 requests. Hijacked h2c connections are not tracked by Shutdown and are not waited on."

pkg/server/server.go
```diff
@@ -227,22 +230,35 @@ func ServeConnect(ctx context.Context, logger *slog.Logger, cfg Config, deps api
...
-			metricsServer.Shutdown(context.Background())
+
+			ctxShutdown, cancel := context.WithTimeout(context.Background(), connectServerGracePeriod)
+			defer cancel()
+
+			if err := metricsServer.Shutdown(ctxShutdown); err != nil {
+				logger.ErrorContext(ctxShutdown, "metrics server shutdown error", "err", err)
+				return
+			}
+			logger.Info("Graceful shutdown of metrics server complete")
```
rohilsurana: Nit: if the metrics server never started (say the port was busy), this goroutine still logs "Graceful shutdown of metrics server complete" when the context is canceled. A bit misleading. Not blocking.

pkg/server/server_test.go
```diff
@@ -0,0 +1,87 @@
...
+			if resp.StatusCode == http.StatusOK {
+				return
+			}
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	t.Fatalf("server did not respond at %s in time", url)
+}
+
+func TestServeConnectReturnsAfterShutdownOnContextCancel(t *testing.T) {
```
rohilsurana, line 106: This test also passes on main, without the `WaitGroup`. There, `ServeConnect` returned after cancel too, just too early. So the test does not prove the fix. It is still worth keeping: the 15s timeout would catch a future deadlock in the `Wait`.

One way to test the real behavior, which also removes duplication: the shutdown goroutine now appears twice (connect and metrics, almost identical). Pull it into a small helper like `gracefulShutdown(ctx, logger, srv, name, wg)`. Then unit test the helper with a plain `http.Server` and a slow handler: start a request that sleeps, cancel the context, and check that the serve function only returns after the request finished.

Two small notes on the test itself:
- `freePort` closes the probe listener and then reuses the port. Another process can grab it in between. This is the standard trick, just a known source of flakes in CI.
- The test boots the full connect stack with an empty `api.Deps{}`. That works today. It will break if handler construction ever touches deps, for reasons unrelated to shutdown.

rohilsurana (commented):
This fixes a real problem. `http.Server.Shutdown` closes the listener first and finishes requests after. So `ListenAndServe` returns while requests are still running. `StartServer` in `cmd/serve.go` then runs its deferred `Close()` calls right away (DB client, session service, billing services). Before this change, those could close while requests were still being served.

The `WaitGroup` approach is correct. The wait cannot hang, because `server.Shutdown` returns within the 10s grace period. If the server fails to start, the function returns before the `Wait()`, so a bad port cannot deadlock. The metrics server shutdown also improves: it now has a timeout, logs errors, and gets waited on.

Two things worth fixing before merge (details inline):

1. The wait does not cover h2c connections. That is most gRPC traffic on this server.
2. The new test passes with or without the fix, so it does not prove the new behavior.

Not part of this PR, but related: `ServeUI` has no shutdown handling at all. Its `http.ListenAndServe` ignores the context.

rohilsurana (approved):

PR #1751, opened 2026-07-13
docs: RFC 0001 for declarative management of platform resources

rohilsurana:
Documents a declarative way to manage Frontier's platform configuration: desired-state YAML files, the `frontier reconcile` and `frontier export` commands, and the five rules every resource kind follows.

The RFC now opens with an "At a glance" section (a flow diagram, the object-vs-value model with a decision diagram, the kinds table, and the five rules in one line each), so the whole model fits on one screen before the detail. This TLDR is just the lead keywords; the RFC has the full text.

## TLDR

- **A file per kind, two commands.** `frontier reconcile -f file [--dry-run]` plans, then applies; `frontier export <kind>` prints the live state in the same format.
- **The file.** `apiVersion: v1`, a `kind`, and a `spec`, checked whole before anything applies.
- **Object or value.** Each entry is one or the other. That decides what a missing entry means and whether delete applies.
- **Five rules every kind follows:** Scope and identity; the file is the desired state; check the whole file before changing anything; converge, do not transact; export is the reverse of reconcile.
- **Five kinds:** PlatformUser, Permission, Role, Preference, Webhook.
- **Boot changes.** Config superuser promotion is gone; boot no longer updates existing roles; `resources_config_path` is deprecated.
- **Auth.** Any superuser credential; the bootstrap service account exists so automation always has one.

## What it covers

- The problem with the three old flows: config-driven superusers, boot-time resource YAML, and hand-run API changes.
- The file format, the two commands, and the reconcile framework.
- The five rules every kind must follow, so future kinds start from the same contract.
- The five kinds against those rules: PlatformUser, Permission, Role, Preference, and Webhook.
- The two boot behavior changes, and the plan to deprecate and remove `resources_config_path`.
- Key decisions with reasons, trade-offs, and future work.

## Status

The RFC is marked Accepted, and the flow is implemented in the `reconcile` package and the `reconcile` and `export` commands. PlatformUser, Permission, Role, and Preference are merged to `main`; the Webhook kind is in review (#1772). This document describes the shipped design.

### DIFF
--- a/docs/rfcs/0001-declarative-reconcile.md
+++ b/docs/rfcs/0001-declarative-reconcile.md
@@ -5,7 +5,7 @@
 | **Status** | Accepted. Implemented in the `reconcile` package and the `reconcile` and `export` commands. |
 | **Author** | Rohil Surana |
 | **Created** | 2026-07-09 |
-| **Updated** | 2026-07-20 |
+| **Updated** | 2026-07-23 |
 
 ## Summary
 
@@ -18,6 +18,54 @@ be seeded from a running server.
 This document sets the rules every kind follows. The kinds are PlatformUser, Permission, Role,
 Preference, and Webhook. New kinds plug in under the same rules without changing the commands
 or the file format.
+
+## At a glance
+
+The whole tool is one loop:
+
+```mermaid
+flowchart LR
+    file["Desired-state file<br/>one YAML doc per kind"]
+    plan["Plan<br/>the difference<br/>always printed"]
+    server[("Live Frontier server<br/>admin API")]
+
+    file -->|"reconcile reads the file"| plan
+    server -->|"reconcile reads the live state"| plan
+    plan -->|"apply, skipped on a dry run"| server
+    server -.->|"export prints the live state as a file"| file
+```
+
+Each entry is either an **object** or a **value**. That one choice decides what a missing entry
+means and whether you can delete it:
+
+```mermaid
+flowchart TD
+    r["A resource the kind manages"] --> q{"Can reconcile<br/>create and delete it?"}
+    q -->|"yes"| o["Object<br/>Permission, Webhook, custom Role<br/>on the server but not in the file: the plan fails<br/>to delete one, set delete: true"]
+    q -->|"no"| v["Value<br/>PlatformUser, Preference, predefined Role<br/>always exists, so no delete flag<br/>not in the file: back to its default"]
+```
+
+The kinds split cleanly, except Role, which is both: a custom role is an object, a predefined
+role is a value.
+
+| Kind | Object or value | Identity | A missing entry | How to remove |
+|---|---|---|---|---|
+| PlatformUser | value | principal + relation | access removed | leave the entry out |
+| Permission | object | namespace + name | plan fails | set `delete: true` |
+| Role, custom | object | name | plan fails | set `delete: true` |
+| Role, predefined | value | name | reset to the shipped definition | cannot be removed |
+| Preference | value | trait name | reset to the trait default | leave the entry out, it resets |
+| Webhook | object | URL | plan fails | set `delete: true` |
+
+Every kind, current and future, follows the same five rules:
+
+1. **Scope and identity.** A kind manages only the entries and fields it declares. Identity is stable, so a rename is a delete plus a create.
+2. **The file is the desired state.** Each entry is an object or a value. A field you write is the whole value. A field you leave out takes the default. An empty field means empty.
+3. **Check the whole file before changing anything.** Nothing applies until the file passes. A check that needs the server runs later, in the diff.
+4. **Converge, do not transact.** One API call per change, adds and updates before deletes, no rollback. To recover from any failure, fix the cause and run again.
+5. **Export is the reverse of reconcile.** For every state the server can reach, reconciling its export plans zero changes.
+
+The rest of this document is the detail behind that picture.
 
 ## Problem
 
@@ -73,13 +121,26 @@ spec:
   and applies it unless this is a dry run.
 - `frontier export <kind>` prints the live state of one kind as a file on stdout.
 
+A reconcile run has three steps. A dry run stops after the plan:
+
+```mermaid
+flowchart TD
+    s1["1. Check the whole file<br/>a bad file stops the run<br/>before any change"]
+    s2["2. Diff each kind against the live server<br/>print one plan per kind<br/>a dry run stops after this"]
+    s3["3. Apply, in file order<br/>adds and updates first, then deletes<br/>one API call per change, no rollback"]
+    fail["If a call fails<br/>fix the cause and run again<br/>what already applied plans no change"]
+
+    s1 --> s2 --> s3
+    s3 -.-> fail
+```
+
 Both log in as a superuser through the admin API. The bootstrap service account exists so
 automation always has one. The `-H` flag passes the token for now. A file or environment input
 is planned (see future work).
 
 ## The five rules
 
-Every kind, current and future, follows these five rules.
+These are the one-liners from [At a glance](#at-a-glance), in full.
 
 ### 1. Scope and identity
 
@@ -136,16 +197,8 @@ export of a state the server can reach is a bug in the kind.
 
 ## The kinds
 
-| Kind | Type | Identity | A missing entry | Delete |
-|---|---|---|---|---|
-| PlatformUser | value | (principal, relation) | access removed | remove the entry |
-| Permission | object | namespace + name | plan fails | `delete: true` |
-| Role, custom | object | name | plan fails | `delete: true` |
-| Role, predefined | value | name | reset to the shipped definition | not allowed |
-| Preference | value | trait name | reset to the trait default | reset, no flag |
-| Webhook | object | URL | plan fails | `delete: true` |
-
-Only what is specific to each kind follows.
+The table in [At a glance](#at-a-glance) maps each kind to its type and behavior. Only what is
+specific to each kind follows.
 
 **PlatformUser.** An entry is `{type, ref, relation}`: a user (by email or id) or a service
 user (by id), with relation `admin` or `member`. The file is the full access list, so a missing
--- a/docs/rfcs/0001-declarative-reconcile.md
+++ b/docs/rfcs/0001-declarative-reconcile.md
@@ -1,0 +1,237 @@
+# RFC 0001: Declarative management of platform resources
+
+| | |
+|---|---|
+| **Status** | Accepted. Implemented in the `reconcile` package and the `reconcile` and `export` commands. |
+| **Author** | Rohil Surana |
+| **Created** | 2026-07-09 |
+| **Updated** | 2026-07-20 |
+
+## Summary
+
+Frontier can manage its platform configuration from a file instead of one-off API calls.
+You write down what should exist for one kind of resource. `frontier reconcile` compares the
+file with the live server, prints a plan, and applies the difference through the admin API.
+`frontier export` does the reverse: it prints the live state in the same format, so a file can
+be seeded from a running server.
+
+This document sets the rules every kind follows. The kinds are PlatformUser, Permission, Role,
+Preference, and Webhook. New kinds plug in under the same rules without changing the commands
+or the file format.
+
+## Problem
+
+Platform configuration used to be managed in three ways, and all three had problems.
+
+1. **Superusers came from server config** (`app.admin.users`). Adding one meant a config change
+   and a restart. Removing one did not work: the server only promoted, never demoted.
+2. **Custom permissions and roles came from a boot-time file** (`resources_config_path`). It
+   could add and update, never remove. There was no preview. A bad file stopped the server
+   from booting.
+3. **Everything else was set by hand** through the admin API. Nothing recorded what should
+   exist, so changes drifted with no record.
+
+## Goals
+
+- One reviewable file per kind that says what should exist.
+- A plan before every change.
+- Removal that works, and that always shows in the plan first.
+- Changes that survive restarts and upgrades.
+- New kinds that plug in without new commands or a new file format.
+
+## Non-goals
+
+- Managing user data (organizations, projects, groups, users, policies). People manage those
+  through the product, and a file would fight them.
+- Renaming a role's internal name. The server looks predefined roles up by name in many code
+  paths.
+- Deciding how a deployment stores its files or triggers runs. This document defines the
+  commands and their behavior. Each install wires them into its own process.
+
+## The file
+
+A file holds one or more YAML documents. Each names a version and a kind, and carries a spec:
+
+```yaml
+apiVersion: v1
+kind: PlatformUser
+spec:
+  - type: user
+    ref: alice@example.org
+    relation: admin
+```
+
+- A document with no `apiVersion` is read as `v1`. An unknown version is rejected.
+- A document with content but no `kind`, or with a missing spec, is rejected. That is usually a
+  typo. To mean an empty list, write `spec: []`.
+- The whole file is checked before anything applies. Documents then apply in the order they
+  appear.
+
+## The commands
+
+- `frontier reconcile -f <file> [--dry-run]` checks the whole file, prints one plan per kind,
+  and applies it unless this is a dry run.
+- `frontier export <kind>` prints the live state of one kind as a file on stdout.
+
+Both log in as a superuser through the admin API. The bootstrap service account exists so
+automation always has one. The `-H` flag passes the token for now. A file or environment input
+is planned (see future work).
+
+## The five rules
+
+Every kind, current and future, follows these five rules.
+
+### 1. Scope and identity
+
+A kind manages a set it declares: which entries, and which fields within an entry. Anything
+outside that set is invisible. The diff skips it, export leaves it out, and a file entry that
+names it is rejected. Values other tools set (webhook headers, extra role metadata, signing
+secrets) pass through writes untouched. Every entry has one stable identity, like a name or a
+URL. Reconcile never changes an identity. A rename is a delete plus a create.
+
+### 2. The file is the desired state
+
+Each entry is an object or a value. The resource decides which. A kind does not pick from a
+menu.
+
+- An **object** is something the flow creates and deletes: a custom role, a permission, a
+  webhook. Every object on the server must appear in the file. One that is missing fails the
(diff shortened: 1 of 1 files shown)


docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+**What a role permission change actually does.** A role's permissions live as data
+relationships in SpiceDB, one per permission the role grants. An update swaps that set in
+one step. Every principal holding the role gains or loses the permission at that moment,
+across all server pods, with no restart.
+
+**One warning when narrowing roles.** The SpiceDB schema also grants some permissions
+directly, outside roles. For example, organization `get` is also granted to the org's
+`member` and `owner` relations and to platform superusers. Removing `get` from a role does
+not block someone who also holds one of those direct grants. Role-only permissions (like
+`rolemanage`, `policymanage`, and all custom permissions) always flip cleanly. Before
```
rohilsurana: This sentence is wrong for `rolemanage` and `policymanage`. Both have direct grants in `base_schema.zed`, exactly like the `get` example above:

```
permission rolemanage   = platform->superuser + granted->app_organization_administer + granted->app_organization_rolemanage + owner
permission policymanage = platform->superuser + granted->app_organization_administer + granted->app_organization_policymanage + owner
```

An org owner keeps both no matter what the role says. In fact almost every predefined org permission carries `+ owner`. Only custom permissions flip cleanly, because they are granted purely through `granted->` role paths.

I tested this on a live server built from #1737. Two users held `app_organization_owner`: one with the `owner` relation (the normal case — every membership path pairs the role with the relation via `orgRoleToRelation`), one with only a role policy. After narrowing the role's permissions to `[get, update, policymanage]` via reconcile:

| Check | owner relation + role | role only |
|---|---|---|
| `rolemanage` | still allowed | denied |
| `delete` | still allowed | denied |

So the role mechanism works instantly for role-path access, but the direct `owner` grant bypasses it for every real owner.

Suggested fix for the paragraph: name only custom permissions as the clean case, and note that narrowing a predefined role today does not restrict org owners at all, because ownership is granted through the `owner` relation directly in the schema. Worth adding the forward-looking point too: once ownership moves fully to roles and the relation-based grants (`+ owner`, `+ member`) come out of the base schema, reconcile will be able to properly control access for predefined roles as well. That migration is out of scope here, but naming it explains why the limitation exists and where it goes.

rohilsurana (commented):
Verified the narrowing warning against the schema and against a live server (built from #1737, seeded an org, narrowed `app_organization_owner` via reconcile). One sentence in it is factually wrong; details inline. More comments on the other sections will follow.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+   Silently skipping an operator-owned object is not an allowed answer. Drift must
+   either fail the plan or show up in it.
+5. **Destructive deletion needs an explicit `delete: true`**, unless the kind picked
+   removal-by-absence in rule 4. A kind whose API has no delete must define what the
+   flag means (for example, deactivate) or reject it.
+6. **Validation runs before anything applies.** One bad entry fails the whole document.
+7. **Adds and updates run before removes.** A half-finished apply can leave extra access
+   for a moment, but never less than intended. On a failure mid-apply, the report still
+   shows what was applied.
+8. **Kinds apply in dependency order** when several files are applied together: an
```
rohilsurana: This reads like the tool reorders kinds, but the commands section (line 89) says documents apply in file order. The intent is that ordering is the file author's job. Suggest rewording rule 8 to say that plainly: the tool applies documents in the order they appear, and the author must put dependencies first (permissions before roles). Validation could also reject a file where a Role document comes before a Permission document, so a wrong order fails fast instead of failing mid-apply.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+Both commands talk to the admin API. They log in with any superuser credential; the
+bootstrap service account exists so automation always has one.
+
+### The framework
+
+Each kind implements a small interface in `internal/reconcile`:
+
+- `Reconcile(spec, dryRun)`: read the current state through the API, diff it against the
+  spec, report the plan, and apply it unless this is a dry run.
+- `Export()` (optional): return the current state as a spec value.
```
rohilsurana: "Optional" here contradicts rule 9 ("Every kind implements export"). Export is required for every kind. If the Go interface stays separate for mechanical reasons, the doc should still say: a kind is not complete without export, and registration should fail for a kind that lacks it. Raised the same point on #1731, which currently describes `Exporter` as "the optional reverse of Reconciler".

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+   - **Removal** — only for kinds whose purpose is controlling a small access list,
+     where absence meaning removal is the feature.
+   - **Reset to the server default** — only for settings-like kinds where every entry
+     always has a value and "not managed" has a well-defined default.
+   Silently skipping an operator-owned object is not an allowed answer. Drift must
+   either fail the plan or show up in it.
+5. **Destructive deletion needs an explicit `delete: true`**, unless the kind picked
+   removal-by-absence in rule 4. A kind whose API has no delete must define what the
+   flag means (for example, deactivate) or reject it.
+6. **Validation runs before anything applies.** One bad entry fails the whole document.
```
rohilsurana: Validation is per document, but documents in one file can contradict each other. Example: a Permission document deletes `compute/order:get` while the Role document still lists it. Permissions apply first, the delete cascades the permission out of the role, and then the role update references a permission that no longer exists. Suggest extending this rule with cross-document validation: role permission references must resolve against the server state, plus permissions created in this file, minus permissions deleted in this file.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+  decision 5. Would need a large refactor of the membership, invitation, and notification
+  paths. Titles cover the real need.
+
+## Drawbacks and accepted trade-offs
+
+- Predefined role permission updates from new versions stop flowing to existing
+  installations on their own. Desired-state files are how they roll out now.
+- A failed apply can stop between adds and removes. The report shows exactly how far it
+  got, and adds-first ordering keeps the failure mode on the safe side.
+- The reconciler trusts the file. Review of the file and of the dry-run plan is the
```
rohilsurana: Worth one more drawback here: concurrent runs. Two reconciles racing each other, or one racing a manual admin change, follow last-write-wins — read, diff, and apply are not atomic. That is acceptable because the flow is declarative: re-running the same file converges to the same end state, so a lost race is corrected by the next apply. Saying this in the doc heads off the question.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Design
+
+### The file format
+
+A desired-state file holds one or more YAML documents. Each document names a kind and
+carries a spec:
+
+```yaml
+kind: PlatformUser
```
rohilsurana, line 105: The document format has no version field, so there is no way to tell old files from new ones if a kind's spec shape ever changes. Add `apiVersion: v1` to every document now. It costs one line per document and saves a migration puzzle later.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+## Terminology
+
+| Term | Meaning |
+|---|---|
+| **SpiceDB** | The database Frontier uses to answer permission checks. It stores a schema (the rules) and relationships (the data). All Frontier pods share one SpiceDB. |
+| **Base schema** | The SpiceDB rules built into the Frontier binary. It defines the object types (`app/organization`, `app/role`, ...) and how a permission check can be satisfied. It only changes at runtime to add custom permissions. |
+| **Operator** | A person who runs a Frontier installation: edits desired-state files, reviews plans, and applies them. Different from a product user, who uses Frontier through applications. |
+| **Kind** | One type of managed resource, for example `PlatformUser` or `Role`. Each kind has its own file format and its own rules. |
+| **Reconcile** | Compare a desired-state file with the live server and make the server match. Runs through the admin API. |
+| **Platform user** | A user or service account with platform-wide access: relation `admin` (superuser) or `member`. |
```
rohilsurana: Terminology nit: the doc says "service account" here and in the bootstrap row, but "service user" in the PlatformUser section. The codebase says service user. Suggest standardizing on "service user" throughout, with at most one aside that other systems call this a service account.

rohilsurana (commented):
Follow-up comments on the remaining sections, per review discussion.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Right on both counts — verified in the schema: `rolemanage` and `policymanage` carry `+ owner`, and 27 grants in `base_schema.zed` do. Rewrote the paragraph in 76e2d265: direct `owner`/`member`/superuser grants bypass role narrowing, an org owner keeps everything regardless of the role, only custom permissions flip cleanly for everyone, and the forward-looking note about moving ownership grants out of the base schema is in.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Reworded rule 8 in 76e2d265: the tool applies documents in file order and does not reorder; dependency order is the author's job (permissions before roles), and validation should reject a Role document appearing before a Permission document so a wrong order fails fast.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Resolved the stronger way: `Export` is folded into the `Reconciler` interface itself (raystack/frontier#1731, e1c67b4c), so a kind cannot compile without it — no separate interface and no registration check needed. The framework section now says so (76e2d265).

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Added to rule 6 in 76e2d265: cross-document checks are part of the contract — a role's permission references must resolve against the server state, plus permissions the file creates, minus permissions it deletes. Implementation lands in #1737 before it merges.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Added as a drawback in 76e2d265: concurrent runs are last-write-wins since read, diff, and apply are not atomic; acceptable because re-running the same file converges, so a lost race is corrected by the next apply.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Added in 76e2d265 and implemented in raystack/frontier#1731 (b92eedaa): every document carries `apiVersion: v1`, a document without it is read as v1 so existing files keep working, an unknown version is rejected, and `frontier export` emits the field.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+A document with a missing spec is rejected, because that is usually a typo. An empty list
+must be written as `spec: []`.
+
+### The commands
+
+- `frontier reconcile -f <file> [--dry-run] --host <host> -H "Authorization:Basic <token>"`
+  parses and validates the whole file first, then handles each document in file order. It
+  prints one report per kind: `no changes`, a plan (dry run), or what was applied.
```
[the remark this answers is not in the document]
rohilsurana: Implemented in raystack/frontier#1731 (b92eedaa): `Run` now parses and checks the whole file — version, registered kinds, spec presence — before dispatching anything, so a malformed later document stops the run before any change applies. Entry-level (semantic) validation still runs per kind right before that kind applies, and a mid-apply failure still returns the partial report. The RFC wording now states exactly this split (76e2d265).

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+- `frontier export <kind>` prints the live state of one kind as a desired-state document on
+  stdout. The kind argument is case-insensitive and accepts a plural.
+
+Both commands talk to the admin API. They log in with any superuser credential; the
+bootstrap service account exists so automation always has one.
```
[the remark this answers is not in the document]
rohilsurana: Agreed on the risk; handled in two steps. The doc now marks `-H` as an interim mechanism that automation must mask in logs, and a file- or environment-based input is listed as planned work (76e2d265). The flag itself stays for now so the change is deliberate rather than rushed into this proposal.

docs/rfcs/0001-declarative-reconcile.md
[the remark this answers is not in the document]
rohilsurana: The RFC does not promise reordering — that was the ambiguity. Rule 8 is reworded (76e2d265): documents apply strictly in file order, dependency order is the file author's job, and validation should reject a known-bad order (Role before Permission) to fail fast.

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+
+### The framework
+
+Each kind implements a small interface in `internal/reconcile`:
+
+- `Reconcile(spec, dryRun)`: read the current state through the API, diff it against the
+  spec, report the plan, and apply it unless this is a dry run.
+- `Export()` (optional): return the current state as a spec value.
+
+New kinds register in one map in the CLI. The behavior rules every kind must follow are
```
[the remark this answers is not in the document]
rohilsurana: Fixed at the interface level: `Export` is part of `Reconciler` itself now (raystack/frontier#1731, e1c67b4c), so a kind cannot be registered without it. The doc lines you flagged are updated to match (76e2d265).

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,317 @@
...
+5. **No role renames.** The server looks predefined roles up by name in more than a dozen
+   live code paths (owner constraint checks, group creation, invitation acceptance,
+   service user creation, admin listings, notifications). Renaming the name would break
+   them all. The display `title` is the thing to rename, and the file can manage it.
+6. **Permissions apply before roles** when both change: roles reference permission slugs,
+   so the permissions must exist first. Deleting a permission cascades out of the roles
+   that grant it, so the reverse order is not needed for removal.
+7. **Reads and writes use the public API**, not the database. The reconciler is an API
+   client like any other: every change is validated, audited, and visible to all pods at
+   once. It also means the CLI works against any reachable Frontier, old or new.
```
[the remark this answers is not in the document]
rohilsurana: Qualified in 76e2d265: decision 7 now states per-kind minimums — PlatformUser needs v0.108.0, Permission and Role need the release that ships this proposal, since the boot change is what makes role edits survive restarts.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Standardized in 76e2d265: the doc says service user throughout, with a new terminology row carrying the one aside ("what other systems call a service account"). The bootstrap term is now "bootstrap service user".

docs/rfcs/0001-declarative-reconcile.md
```diff
@@ -0,0 +1,346 @@
...
+their own anymore. Only fresh installs get the new defaults. Release notes must call out
+predefined-role changes, and operators roll them out through their desired-state files. A
+brand-new installation works the same way: boot installs plain defaults, and the first
+reconcile brings it to the desired state.
+
+## Decisions and reasons
+
+1. **Three different rules for a missing entry.** Platform users: removed, because
+   controlling that list is the kind's purpose and the list is small. Permissions and
+   custom roles: the plan fails, because their removal has a bigger blast radius, and
```
rohilsurana: The RFC says this flow replaces the boot-time loaders but never says what happens to `resources_config_path` itself. Meanwhile the Permission section relies on it staying alive (the delete guard checks it). That leaves the same objects with two owners: a permission in the boot file can never be deleted through reconcile until someone also edits the boot config and restarts, and a role can be created by the boot file and then changed by reconcile, so the file stops describing the server.

Decision: deprecate `resources_config_path`, remove it after 2 minor versions. Suggest adding a short subsection under server-side changes that states:

- The loader is deprecated as of this RFC and removed after 2 minor versions.
- Migration is one export: run `frontier export permission` and `frontier export role`, commit the output as the desired-state file, and drop `resources_config_path` from the server config.
- The permission delete guard stays for the deprecation window and goes away with the loader.

rohilsurana: Added as a subsection under server-side changes (commit follows 3ae71e2d): `resources_config_path` is deprecated as of this RFC and removed after two minor versions; migration is one export cycle (`frontier export permission` + `frontier export role`, commit, drop the config key); the permission delete guard lives only for the deprecation window and goes away with the loader. The Permission section's guard bullet now cross-references this so the two-owners situation is visibly temporary.

docs/rfcs/0001-declarative-reconcile.md
rohilsurana: Superseding my earlier reply after discussion: we are NOT building cross-document validation. Kinds stay independent — a reconciler never resolves another kind's references. The server is the enforcement point: a role create or update that references a missing permission fails at the API call with a clear error, the run stops there, and nothing is lost; the next dry run shows the resulting drift. Rule 6 is reworded to state this trade explicitly (commit after 6b121722), and raystack/frontier#1737 has a test pinning the apply-time failure path (f2a5c1cf). Two reasons for the change: the server already guarantees integrity, and our per-kind invocation model means in-file cross-checks would rarely fire anyway.

rohilsurana:
## Review notes (self-review after an independent pass)

Three wording fixes, all in 3ce47415:

- The predefined-role section said an omitted field "takes the definition's value" without
  the caveat that a field which is empty on the server converges only when it is listed.
  The headline now points at that edge, which is described a few lines below.
- The reset target is `schema.PredefinedRoles`, which can name the same role twice. The
  code takes the first definition, matching boot. The RFC now states this.
- Rule 8's parenthetical read awkwardly next to "permissions go before roles". Reworded so
  the good order and the rejected order read the same way round.

No behavior contradictions found between the RFC and the code on the implementation branch.

rohanchkrabrty (approved):
LGTM 🚀

PR #1755, opened 2026-07-13
fix(users): default user detail to Security tab via index redirect

Shreyag02:
## Summary
The user detail page always renders the Security tab, but landing on the bare `/users/:userId` (e.g. via the detail breadcrumb or a direct link) left the URL without the `/security` suffix, so the Security chip didn't appear selected. This PR makes `/users/:userId/security` the canonical default route via a route-level redirect, so the tab is always selected and the URL is consistent.

## Changes
- Add an index redirect from `/users/:userId` to `/users/:userId/security`, mirroring the existing org-detail default-tab pattern (`members`).

## Technical Details
The `:userId` route is now an element-less path group with two children: an `index` route that renders `<Navigate to="security" replace />` and a `security` route that renders `UsersPage`. Because the parent has no `element`, there's no `<Outlet />` requirement — the matched child's element (`UsersPage`) renders directly. The redirect handles every entry into the bare user URL (breadcrumb, direct link), and the Security chip's existing active check (`currentPath.startsWith(link.path)`) then matches — no navbar change or component-level `useEffect` needed.

## Test Plan
    - Verified `/users/:userId` redirects to `/users/:userId/security` with the Security tab selected.
    - Verified the breadcrumb avatar/name lands on the `/security` URL via the redirect.
    - Verified user row clicks (already targeting `/security`) still work as before.
    - Verified unknown subpaths (e.g. `/users/:userId/foo`) no longer render the detail view (fall through to the app catch-all redirect).

N/A

### DIFF
--- a/web/apps/admin/src/pages/users/UsersPage.tsx
+++ b/web/apps/admin/src/pages/users/UsersPage.tsx
@@ -1,5 +1,5 @@
 import { UsersView, useAdminPaths } from "@raystack/frontier/admin";
-import { useCallback, useEffect } from "react";
+import { useCallback } from "react";
 import { useParams, useNavigate, useLocation } from "react-router-dom";
 import { clients } from "~/connect/clients";
 import { exportCsvFromStream } from "~/utils/helper";
@@ -11,15 +11,6 @@ export default function UsersPage() {
   const navigate = useNavigate();
   const location = useLocation();
   const paths = useAdminPaths();
-
-  // Security is the default tab for the user detail page: redirect the bare
-  // `/users/:userId` index to `/users/:userId/security` so the tab stays selected.
-  useEffect(() => {
-    const securityPath = `/${paths.users}/${userId}/security`;
-    if (userId && location.pathname !== securityPath) {
-      navigate(securityPath, { replace: true });
-    }
-  }, [userId, location.pathname, navigate, paths.users]);
 
   const onExportUsers = useCallback(async () => {
     await exportCsvFromStream(adminClient.exportUsers, {}, "users.csv");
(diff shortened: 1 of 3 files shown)


web/sdk/admin/views/users/details/layout/navbar.tsx
```diff
@@ -43,7 +43,7 @@ export const UserDetailsNavbar = ({
           </Breadcrumb.Item>
           <Breadcrumb.Separator />
           <Breadcrumb.Item
-            href={`/${paths.users}/${user?.id}`}
+            href={`/${paths.users}/${user?.id}/security`}
```
rohanchkrabrty: This change is not needed her, only the router change is required

web/apps/admin/src/pages/users/UsersPage.tsx
```diff
@@ -12,6 +12,15 @@ export default function UsersPage() {
...
 
+  // Security is the default tab for the user detail page: redirect the bare
+  // `/users/:userId` index to `/users/:userId/security` so the tab stays selected.
+  useEffect(() => {
+    const securityPath = `/${paths.users}/${userId}/security`;
+    if (userId && location.pathname !== securityPath) {
+      navigate(securityPath, { replace: true });
+    }
+  }, [userId, location.pathname, navigate, paths.users]);
+
```
rohanchkrabrty: Please add redirect in the route definition itself. Let's avoid useeffect

rohanchkrabrty (changes requested):

rohanchkrabrty (approved):

PR #1756, opened 2026-07-14
fix: wait for background goroutines during shutdown

AmanGIT07:
Based on #1747 — merge that first; this base retargets to main after.

## Summary
`StartServer` now waits for the UI server and the audit log listener to
finish shutting down before running deferred cleanup, instead of only
waiting for the connect server.

## Changes
- The log listener, UI server, and connect server run under an
  `errgroup`; `StartServer` returns `g.Wait()`, so deferred cleanup
  (database and service `Close()` calls) runs only after all three
  finish.
- All three receive the group context, so a connect server startup
  failure also winds down the UI server and listener.
- A cancelled context is no longer logged as "log listener failed" on
  normal shutdown.

## Technical Details
`errgroup.WithContext` cancels the group context when any member
returns an error. Only `ServeConnect` returns a non-nil error, so
`g.Wait()` returns exactly what `StartServer` returned before.

## Test Plan
- [x] `go build`, `go vet`, `golangci-lint` clean
- [x] Manual: Ctrl-C logs both server "complete" lines before the first
  "cleaning up" line, with no listener warning
- No new unit test: `StartServer` needs live Postgres and SpiceDB before
  reaching this code; the composed pieces are tested in #1745 and #1747

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -333,6 +333,14 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
 		return server.ServeConnect(gctx, logger, cfg.App, deps, promRegistry)
 	})
 
+	// Once shutdown starts, unregister the signal handler so a second
+	// SIGINT/SIGTERM gets default handling and can kill a stuck drain
+	// instead of being swallowed.
+	go func() {
+		<-gctx.Done()
+		cancelFunc()
+	}()
+
 	// Wait for every server to finish draining before the deferred cleanup
 	// above closes the database and other dependencies under them.
 	return g.Wait()
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -3,14 +3,18 @@ package cmd
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"os"
 	"os/signal"
+	"slices"
 	"strings"
 	"syscall"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/raystack/frontier/core/aggregates/orgbilling"
 	"github.com/raystack/frontier/core/aggregates/orginvoices"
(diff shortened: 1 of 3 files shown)


cmd/serve.go
```diff
@@ -311,16 +313,29 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
...
-	go server.ServeUI(ctx, logger, cfg.UI, cfg.App)
+	g.Go(func() error {
+		return server.ServeConnect(gctx, logger, cfg.App, deps, promRegistry)
+	})
 
-	// start connect server (blocking)
-	return server.ServeConnect(ctx, logger, cfg.App, deps, promRegistry)
+	// Wait for every server to finish draining before the deferred cleanup
+	// above closes the database and other dependencies under them.
+	return g.Wait()
```
rohilsurana, line 346: Now that `StartServer` actually waits out the drain, a second Ctrl-C during that window does nothing. `ctx` comes from `signal.NotifyContext` (line 124), and its stop function only runs via the `defer` when `StartServer` returns — so while the drain runs (up to ~10s), further SIGINT/SIGTERM are swallowed by the registration. Before this PR the process exited almost immediately, so it never mattered.

Unregistering the handler once shutdown starts restores default signal handling, giving operators "first signal drains, second signal quits now":

```suggestion
	// Once shutdown starts, unregister the signal handler so a second
	// SIGINT/SIGTERM gets default handling and can kill a stuck drain
	// instead of being swallowed.
	go func() {
		<-gctx.Done()
		cancelFunc()
	}()

	// Wait for every server to finish draining before the deferred cleanup
	// above closes the database and other dependencies under them.
	return g.Wait()
```

`cancelFunc` doubles as the stop function, and by the time `gctx` is done the context is already cancelled (either by a signal or by a member error), so calling it here only unregisters the handler.

rohilsurana (commented):
Verified the whole chain: the listener returns `ctx.Err()` on cancel, #1747's `ServeUI` returns on every path, and `ServeConnect`'s drain is bounded, so `g.Wait()` cannot hang for more than the grace period. One suggestion inline about signal handling during that window.

cmd/serve.go
AmanGIT07: Good catch — the swallow-window only became real once the drain actually waits. Applied as suggested in 9a901169: the handler unregisters when shutdown starts, so the first signal drains and a second one quits immediately.

rohilsurana (approved):

PR #1757, opened 2026-07-14
fix: drop orgId from group RPCs

rohanchkrabrty:
## Summary

- Bumped `@raystack/proton` to the latest (`0.1.0-f0b06e61...`) in both `web/sdk` and `web/apps/admin` so the SDK regenerates without `org_id` on the affected group RPCs.
- Removed the now-invalid `orgId` argument from `getGroup`, `updateGroup`, `listGroupUsers`, `removeGroupUser`, `deleteGroup`, and `setGroupMemberRole` calls across the team views.
- Left `ListOrganizationGroups` and `ListCurrentUserGroups` untouched — they still take `org_id`.
- Verified the full `pnpm build` (SDK + admin) passes against the regenerated SDK types with no new type errors.

### DIFF
--- a/web/sdk/client/views/teams/team-details-view.tsx
+++ b/web/sdk/client/views/teams/team-details-view.tsx
@@ -86,8 +86,7 @@ export function TeamDetailsView({
   } = useQuery(
     FrontierServiceQueries.getGroup,
     create(GetGroupRequestSchema, {
-      id: teamId || '',
-      orgId: organization?.id || ''
+      id: teamId || ''
     }),
     {
       enabled: !!organization?.id && !!teamId,
@@ -114,8 +113,7 @@ export function TeamDetailsView({
   } = useQuery(
     FrontierServiceQueries.listGroupUsers,
     create(ListGroupUsersRequestSchema, {
-      id: teamId || '',
-      orgId: organization?.id || ''
+      id: teamId || ''
     }),
     { enabled: !!organization?.id && !!teamId }
   );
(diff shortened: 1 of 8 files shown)


web/sdk/client/views/teams/team-details-view.tsx
```diff
@@ -86,8 +86,7 @@ export function TeamDetailsView({
   } = useQuery(
     FrontierServiceQueries.getGroup,
     create(GetGroupRequestSchema, {
-      id: teamId || '',
-      orgId: organization?.id || ''
+      id: teamId || ''
     }),
     {
       enabled: !!organization?.id && !!teamId,
```
Shreyag02, line 92: shouldn't we remove org id from enabled as well

web/sdk/client/views/teams/team-details-view.tsx
```diff
@@ -114,8 +113,7 @@ export function TeamDetailsView({
   } = useQuery(
     FrontierServiceQueries.listGroupUsers,
     create(ListGroupUsersRequestSchema, {
-      id: teamId || '',
-      orgId: organization?.id || ''
+      id: teamId || ''
     }),
     { enabled: !!organization?.id && !!teamId }
```
Shreyag02, line 118: same issue here for enable calc

Shreyag02 (changes requested):

web/sdk/client/views/teams/team-details-view.tsx
rohanchkrabrty: It's not needed here since it determines the query shouldn't run when activeOrg is not there

Shreyag02 (approved):

PR #1763, opened 2026-07-16
feat(admin): show org slug in detail URLs instead of UUID

Shreyag02:
## Summary
Admin organization detail pages now show the org's readable **slug** (its `name`) in the URL instead of the raw UUID
(e.g. `/organizations/acme-corp` instead of `/organizations/6f3a…`).
All API calls still run on the org **id** — the slug is display-only — so the change is purely about the URL, not the data layer.

## Changes
  - Org detail URLs use the slug for a readable, shareable address.
  - The org **id** travels with in-app navigation (list, admins, create, audit) via router state, and is used for every org RPC.
  - Cold loads (refresh, bookmark, old UUID links) resolve the URL id/slug in one call, with a loader while it resolves.
  - Old `/organizations/<uuid>` links keep working and are rewritten in place to the canonical slug URL once resolved (disabled orgs stay on the id URL).
  - Breadcrumb navigates client-side (no full reload); tab links keep the slug from the current URL.

## Technical Details
  - **Why id, not slug, for RPCs:**
  The id is the org's stable primary key and rides along with in-app navigation via router state, so every org RPC (`getOrganization`, KYC, members, roles, billing) uses the id directly and never pays a slug→id lookup. The slug is display-only.
  - **How the id reaches the page:**
  Navigators pass `{ state: { orgId } }` alongside the slug URL. The detail page holds that id across tab switches (which clear router state).
  - **Cold-load fallback:**
  With no router state, the page resolves the URL segment via a single `getOrganization` call. The server's `GetRaw` accepts either form — a UUID (`GetByID`) or a slug (`GetByName`) — and returns disabled orgs too, so ids, enabled slugs, and disabled slugs all resolve in one request; only a genuinely unknown segment 404s and redirects to the list. Resolution runs in parallel with the id-scoped RPCs, so there's no per-load waterfall.
  - **Canonical URL:**
  An old UUID URL is `replace`-navigated to its slug once resolved (preserving the active sub-tab), so each org has a single live URL and the back-button stays sane. Disabled orgs are left on the id URL — the id is always a valid address for them.
  - **Admins/audit views** resolve the org by id (`useOrganizationLookup`) to display its title and build the slug link.

  Note: an org's `name` (slug) is globally unique and is never freed — disabling only flips the org's `state` — so a slug never gets reassigned to a different org and bookmarks stay pointed at the same org.

## Test Plan
    - List → org detail shows a slug URL; members/projects/tokens/etc. tabs load and stay on the slug.
    - Breadcrumb (org title + "Organizations") navigates without a full reload.
    - Audit log → organization link and Create org → redirect both land on the correct detail page.
    - Disabled org opened from the list works (id via state).
    - Refresh / paste a slug URL resolves (loader → page); old UUID link still works and rewrites in place to the slug; an unknown segment redirects to the list.
 - [x] Build and type checking passes (admin app `tsc` clean; SDK rebuilt)

N/A


rohilsurana:
@Shreyag02 the disabled org flow is broken, we need to have access to disabled org as well, in case we need to unblock it again. Let's look into how we can make that happen.
One way is that we use UUID in url for disabled orgs instead of the slug change that we are doing here.

rohilsurana (approved):

PR #1767, opened 2026-07-17
chore(server): remove the resources_config_path boot loader flow

rohilsurana:
## Status

The deprecation window has elapsed. `resources_config_path` was deprecated at v0.110.0 for two minor versions, and frontier is now at v0.114.0, so the removal is due. This stays a draft on the deployment gate below: it should land only once every deployment has cut over to the reconcile flow.

## What

Removes the `app.resources_config_path` boot loader end to end. Custom permissions and role overrides are now managed through the reconcile flow (the `Permission` and `Role` kinds), so the server no longer reads a resource config file at startup.

- `cmd/serve.go`: drop the resource blob load and the `SchemaConfigRepository` wiring.
- `internal/bootstrap`: drop the `FileService` (`schemaConfig`) dependency. `MigrateSchema` now re-applies the base schema merged with the permissions already in the DB (via `AppendSchema`, which preserves them) instead of reading a config file. `MigrateRoles` no longer creates config-defined custom roles, since reconcile owns them. `BuiltinPermissions` (the permission delete-guard) reports only base-schema permissions.
- `pkg/server/config.go`: remove the `resources_config_path` and `resources_config_path_secret` fields.
- `internal/store/blob/schema_repository.go`: removed, no longer used.
- `config/sample.config.yaml`: drop the sample entries.
- `test/e2e`: the regression and smoke suites seed custom compute resources through the admin API instead of a resource config file. The shared setup lives in `testbench/helper.go`.

## Why it is safe

Existing custom permissions survive across boots because they live in the database, and `AppendSchema` merges them back into the schema. Existing role overrides survive because `MigrateRoles` is skip-if-exists. So on a server that has already migrated, this is a no-op. On one that has not, it stops seeding from the config, which is why the cutover gate matters.

## Gating

- This is a breaking change for users still on `resources_config_path`, so removal waits out the two-minor-version deprecation window. That window is now over: deprecated at v0.110.0, past the v0.112.0 removal target, current release v0.114.0.
- A deployment must run the reconcile flow first, so roles and permissions are owned there, and must drop `resources_config_path` from its own config before taking this build.

## Verification

Rebased onto current main (70 commits ahead of the old base) with no conflicts. `go build ./...`, `go vet`, and unit tests for `internal/bootstrap`, `pkg/server`, and `core/resource` pass. gofmt and vet clean.


rohilsurana (commented):
Automated code review (xhigh). Findings below, most important first. Two are behavior questions worth confirming before merge; the rest are cleanups.

cmd/serve.go
```diff
@@ -134,22 +134,6 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
...
-		return err
-	}
-	defer func() {
-		logger.Debug("cleaning up resource blob")
-		if err := resourceBlobFS.Close(); err != nil {
-			logger.Warn("resource blob cleanup failed", "err", err)
-		}
-	}()
-
 	// load billing plans
```
rohilsurana, line 137: **Removing the loader and its deprecation warning together leaves upgraders with no signal.**

An operator who still sets `resources_config_path` on upgrade gets nothing. The key becomes an unknown field that config loading ignores without error, and the old `logger.Warn` deprecation notice is gone. Any new custom permission or role in that file is never registered, so `CheckPermission` and `CreatePolicy` for those resources fail until they reseed through the admin/reconcile API. The docs (`configurations.mdx`, `custom-resources.mdx`) still describe this field. Consider keeping a boot-time warning for one release and updating those docs.

internal/bootstrap/service.go
```diff
@@ -144,25 +137,18 @@ func NewBootstrapService(
...
+// the database.
 func (s Service) BuiltinPermissions(ctx context.Context) (map[string]struct{}, error) {
-	custom, err := s.schemaConfig.GetDefinition(ctx)
-	if err != nil {
-		return nil, err
-	}
-	custom.Permissions = filterDefaultAppNamespacePermissions(custom.Permissions)
-
-	defs, err := ApplyServiceDefinitionOverAZSchema(custom, GetBaseAZSchema())
+	defs, err := ApplyServiceDefinitionOverAZSchema(&schema.ServiceDefinition{}, GetBaseAZSchema())
```
rohilsurana, line 151: **Config-defined permissions lose delete protection.**

`BuiltinPermissions` now covers only the base schema. A permission that used to come from the config file (say `compute/order:create`) is no longer in the builtin set, so `DeletePermission` accepts it where it used to reject with `FailedPrecondition`. That removes the permission and its SpiceDB tuples, and boot no longer recreates it. Every role and policy referencing it silently loses the grant. Please confirm this dropped protection is intended.

internal/bootstrap/service.go
```diff
@@ -269,16 +255,8 @@ func (s Service) MigrateRoles(ctx context.Context) error {
 		}
```
rohilsurana, line 255: **Config-defined roles are no longer self-healed at boot.**

Predefined roles are still recreated by `migrateRole`, but the loop that recreated config-defined roles is gone. So a deleted custom role now stays deleted for good, and anything bound to it loses access with no boot-time recovery. Confirm this is expected now that reconcile is the seeding path.

_(anchored near the diff; refers to internal/bootstrap/service.go:249)_

test/e2e/testbench/helper.go
```diff
@@ -69,6 +69,47 @@ func PromoteBootstrapAdmin(ctx context.Context, ad frontierv1beta1connect.AdminS
 	return fmt.Errorf("promote bootstrap admin (server unavailable after retries): %w", lastErr)
 }
 
+// SeedComputeResources creates the custom "compute" permissions and roles the
+// regression suites rely on, through the admin API. This replaces the removed
+// boot-time resources_config loader: tests now seed custom resources the same
+// way operators do, via reconcile/the admin API.
+func SeedComputeResources(ctx context.Context, ad frontierv1beta1connect.AdminServiceClient) error {
+	basic := base64.StdEncoding.EncodeToString([]byte(BootstrapClientID + ":" + BootstrapClientSecret))
```
rohilsurana: **Duplicated basic-auth context builder.**

These two lines are a verbatim copy of `PromoteBootstrapAdmin` (around lines 48-49): the same base64 of `BootstrapClientID:BootstrapClientSecret` and the same `Authorization: Basic` header. Extracting a small helper like `bootstrapAdminContext(ctx)` keeps the credential constants and header shape in one place so the two auth paths cannot drift.

internal/store/blob/schema_repository_test.go
```diff
@@ -1,100 +0,0 @@
-package blob
```
rohilsurana, line 1: **Orphaned test fixtures.**

This deleted test was the only reader of `file://testdata` via `NewStore`. Nothing in `internal/store/blob` reads it now, so `testdata/compute.yml` and `testdata/database.yml` are left behind as dead fixtures. Worth deleting them in this PR so they do not mislead future readers.

test/e2e/regression/api_test.go
```diff
@@ -106,6 +99,11 @@ func (s *APIRegressionTestSuite) SetupSuite() {
 
 	ctx := context.Background()
 
+	// Seed the custom "compute" permissions and roles the tests below rely on,
+	// via the admin API. This replaces the removed boot-time resources_config
+	// loader — tests now seed custom resources the same way operators do.
```
rohilsurana: Minor and optional: this new comment uses an emdash. Not a blocker, and the repo already uses them elsewhere, so only worth changing if you want to keep new comments plain. Same at `test/e2e/regression/onboarding_test.go:64`.

test/e2e/testbench/helper.go
rohilsurana: Fixed in b74cbdbb: extracted `bootstrapAdminContext(ctx)`, now used by both `PromoteBootstrapAdmin` and `SeedComputeResources` so the header shape lives in one place.

internal/store/blob/schema_repository_test.go
rohilsurana: Fixed in cc6f33c6: deleted the orphaned `testdata/compute.yml` and `testdata/database.yml`.

test/e2e/regression/api_test.go
rohilsurana: Fixed in b74cbdbb: rewrote the comment without an emdash here and in `onboarding_test.go`.

cmd/serve.go
rohilsurana: Docs updated in 85ed5870 (removed `resources_config_path` from the config docs and added a banner to custom-resources.mdx; a full rewrite is coming in a separate PR). The boot-time deprecation warning is intentionally removed: reconcile and the admin API own custom permissions and roles now, and this PR is gated on a deployment running reconcile first. Leaving this open for a call on whether a one-release warning is still wanted.

internal/bootstrap/service.go
rohilsurana: Intended. Under the reconcile model custom permissions are owned by reconcile and the admin API, so config no longer contributes to the builtin set and such a permission becomes deletable. This PR is gated on reconcile owning that state first.

internal/bootstrap/service.go
rohilsurana: Intended, same reasoning as the permission delete-guard: reconcile is the seeding path now, so config-defined roles are no longer recreated at boot. Predefined roles are still self-healed by `migrateRole`.

cmd/serve.go
rohilsurana: No deprecation warning needed. Reconcile and the admin API own custom permissions and roles now, and the deployment is gated on reconcile running first. Resolving.

whoAbhishekSah (approved):

PR #1768, opened 2026-07-17
fix(store): return transaction errors from rql search repositories

AmanGIT07:
## Summary
Search methods in four rql repositories discarded the error returned by `WithTxn`. A failed transaction begin or commit produced an empty result with no error.

## Changes
- Capture and return the `WithTxn` error in `OrgBillingRepository.Search`, `OrgProjectsRepository.Search`, `OrgUsersRepository.Search`, and `ProjectUsersRepository.Search`

## Technical Details
`WithTxn` returns errors from transaction begin, commit, and the callback. Only callback errors reached the caller, through a captured outer variable. In `OrgBillingRepository.Search` the group-by branch shadowed that variable, so its failures were dropped as well.

## Test Plan
- [x] `go test ./internal/store/postgres/` passes (covers the four Search methods)

- [x] Values flow through `?` placeholders, `goqu.Ex{}`, or `goqu.Record{}` — never `fmt.Sprintf` or `+` building a query that gets executed.
- [x] `ToSQL()` callers capture and forward params (`query, params, err := stmt.ToSQL(); db.…Context(ctx, …, query, params...)`). Never `query, _, err := …`.
- [x] No `?` placeholders inside single-quoted SQL literals in `goqu.L` (use `make_interval(hours => ?)`-style functions instead).
- [x] Any `//nolint:forbidigo` or `// #nosec G20x` annotation has a one-line justification on the same line that a reviewer can verify.


internal/store/postgres/org_billing_repository_test.go
```diff
@@ -216,3 +222,25 @@ func TestPrepareGroupByQuery(t *testing.T) {
...
+// transaction begin returns an error.
+func txnFailClient() *db.Client {
+	return &db.Client{DB: sqlx.NewDb(sql.OpenDB(failConnector{}), "postgres")}
+}
+
+type failConnector struct{}
+
+func (failConnector) Connect(context.Context) (driver.Conn, error) {
+	return nil, errors.New("connection failed")
+}
```
rohilsurana: should we extract this into a test helper? it is being used in other test files as well

AmanGIT07: done

rohilsurana (approved):

PR #1769, opened 2026-07-17
fix: log audit log failures instead of dropping them

AmanGIT07:
## Summary
Legacy audit log writes (`audit.Logger.Log` / `LogWithAttrs`) discarded their error in the connect handlers and in three core services. Failures are now logged. The calling operation still succeeds, matching the existing handling in `CreateOrganization`.

## Changes
- Connect handlers (12 files): check the audit log error and record it through the existing `ErrorLogger` with operation tag `<RPC>.AuditLog`
- `core/membership`: log audit log failures through the service's slog logger
- `core/deleter`, `core/organization`: log audit log failures through `slog.WarnContext`
- Package helpers with no request in scope (`logAuditForCheck`, `auditPolicyCreationEvent`) log through `slog.ErrorContext`

## Technical Details
No caller-visible behavior change: an audit log failure never fails the request.

## Test Plan
- [x] `go test ./internal/api/v1beta1connect/ ./core/membership/ ./core/deleter/ ./core/organization/` passes

### DIFF
--- a/core/deleter/service.go
+++ b/core/deleter/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 
 	"github.com/raystack/frontier/core/audit"
 
@@ -276,7 +277,9 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
 		return err
 	}
 
-	audit.NewLogger(ctx, id).Log(audit.OrgDeletedEvent, audit.OrgTarget(id))
+	if err := audit.NewLogger(ctx, id).Log(audit.OrgDeletedEvent, audit.OrgTarget(id)); err != nil {
+		slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.OrgDeletedEvent)
+	}
 	return nil
 }
 
--- a/core/membership/service.go
+++ b/core/membership/service.go
@@ -442,10 +442,12 @@ func (s *Service) removeOrganizationMember(ctx context.Context, orgID, principal
 	}
 
 	s.auditOrgMemberRemoved(ctx, org, principalID, targetAuditType)
-	audit.GetAuditor(ctx, org.ID).Log(audit.OrgMemberDeletedEvent, audit.Target{
+	if err := audit.GetAuditor(ctx, org.ID).Log(audit.OrgMemberDeletedEvent, audit.Target{
 		ID:   principalID,
 		Type: principalType,
-	})
+	}); err != nil {
+		s.log.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.OrgMemberDeletedEvent)
+	}
 
 	return nil
 }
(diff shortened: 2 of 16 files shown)


core/deleter/service.go
```diff
@@ -276,7 +277,9 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
 		return err
 	}
 
-	audit.NewLogger(ctx, id).Log(audit.OrgDeletedEvent, audit.OrgTarget(id))
+	if err := audit.NewLogger(ctx, id).Log(audit.OrgDeletedEvent, audit.OrgTarget(id)); err != nil {
+		slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.OrgDeletedEvent)
+	}
```
rohilsurana, line 282: the pattern seems to be getting repeated everywhere, so can we just inject a logger in auditor and log failure there with a wrapper function?

rohilsurana: i see this is the legacy audit service, we can keep this like this for now, and get rid of it completely in future

AmanGIT07: **Option A — add best-effort wrappers, keep Log/LogWithAttrs unchanged**                              
  
```                                                                                                                           
  // Service gets a logger through the existing Option pattern                                                               
  func WithLogger(l *slog.Logger) Option {                                                                                   
        return func(s *Service) { s.logger = l }                                                                             
  }                                                                                                                          
                                                                                                                             
  func (s *Logger) TryLog(action EventName, target Target, logCtx ...any) {                                                  
        s.TryLogWithAttrs(action, target, nil, logCtx...)                                                                    
  }                                                                                                                          
                                                                                                                             
  func (s *Logger) TryLogWithAttrs(action EventName, target Target, attrs map[string]string, logCtx ...any) {                
        if err := s.LogWithAttrs(action, target, attrs); err != nil {                                                        
                args := []any{                                                                                               
                        "error", err,                                                                                        
                        "event", action.String(),                                                                            
                        "org_id", s.orgID,                                                                                   
                        "target_id", target.ID,                                                                              
                        "target_type", target.Type,                                                                          
                }                                                                                                            
                if len(attrs) > 0 {                                                                                          
                        args = append(args, "attrs", attrs)                                                                  
                }                                                                                                            
                args = append(args, logCtx...)                                                                               
                s.service.logger.WarnContext(s.ctx, "failed to write audit log", args...)                                    
        }                                                                                                                    
  }                                                                                                                          
 ```                                                                                                                            
Call sites collapse to one-liners, e.g. ` auditor.TryLog(audit.ProjectCreatedEvent, audit.ProjectTarget(id))`. The variadic tail carries site-specific context the audit struct doesn't hold — useful where two RPCs emit the same event:   `auditor.TryLog(..., "operation", "AdminCreateOrganization.AuditLog")`. 

**Downside**: two parallel methods (Log + TryLog) on a  service we plan to delete.                                                                                                 
                                                                                                                             
**Option B — make Log/LogWithAttrs void and log failures inside**                                                              
  ```                                                                                                                           
  func (s *Logger) Log(action EventName, target Target) {                                                                    
        s.LogWithAttrs(action, target, nil)                                                                                  
  }                                                                                                                          
                                                                                                                             
 func (s *Logger) LogWithAttrs(action EventName, target Target, attrs map[string]string) {                                  
        l := &Log{ /* unchanged construction */ }                                                                            
        // ...                                                                                                               
        if err := s.service.Create(s.ctx, l); err != nil {                                                                   
                s.service.logger.WarnContext(s.ctx, "failed to write audit log",                                             
                        "error", err, "event", action.String(), "org_id", s.orgID,                                           
                        "target_id", target.ID, "target_type", target.Type)                                                  
        }                                                                                                                    
  }                                                                                                                          
  ```                                                                                                                           
 Single method instead of two, every bare call site in the codebase becomes correct as-is, all the per-site if-blocks from this PR get deleted, and errcheck has nothing to flag by construction. 

**Downside**: no caller can ever learn the write failed (nothing needs that today), and site-specific log context is gone unless we also add a variadic tail here.

AmanGIT07: For now we can keep as is, we can keep this approach for future reference, @rohilsurana

rohilsurana (approved):

PR #1777, opened 2026-07-20
fix(spicedb): migrate batch check to the stable bulk permissions API

AmanGIT07:
## Summary
Move the SpiceDB integration off the experimental API surface. The batch check migrates from the deprecated `BulkCheckPermission` to the stable `CheckBulkPermissions`, and the client no longer requests experimental APIs since nothing uses them afterwards.

## Changes
- `internal/store/spicedb/relation_repository.go` (`BatchCheck`): rename the request, item, and pair types and the client method to their stable names. Same fields, same accessors, no logic change.
- `internal/store/spicedb/spicedb.go`: construct the client with `authzed.NewClient` and hold `*authzed.Client` instead of the experimental variant. Same dial options.

## Technical Details
The stable API is available in the already-pinned `authzed-go` version; no dependency change. The stable response guarantees pairs in request order, which the result indexing relies on.

## Test Plan
- [x] PAT e2e regression suite passes against real SpiceDB, including the batch-check mixed-results case, with the client constructed through `NewClient`

### DIFF
--- a/internal/store/spicedb/relation_repository.go
+++ b/internal/store/spicedb/relation_repository.go
@@ -296,9 +296,9 @@ func (r *RelationRepository) ListRelations(ctx context.Context, rel relation.Rel
 
 func (r *RelationRepository) BatchCheck(ctx context.Context, relations []relation.Relation) ([]relation.CheckPair, error) {
 	result := make([]relation.CheckPair, len(relations))
-	items := make([]*authzedpb.BulkCheckPermissionRequestItem, 0, len(relations))
+	items := make([]*authzedpb.CheckBulkPermissionsRequestItem, 0, len(relations))
 	for _, rel := range relations {
-		items = append(items, &authzedpb.BulkCheckPermissionRequestItem{
+		items = append(items, &authzedpb.CheckBulkPermissionsRequestItem{
 			Resource: &authzedpb.ObjectReference{
 				ObjectId:   rel.Object.ID,
 				ObjectType: rel.Object.Namespace,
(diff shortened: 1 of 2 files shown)


internal/store/spicedb/relation_repository.go
```diff
@@ -296,9 +296,9 @@ func (r *RelationRepository) ListRelations(ctx context.Context, rel relation.Rel
 
 func (r *RelationRepository) BatchCheck(ctx context.Context, relations []relation.Relation) ([]relation.CheckPair, error) {
 	result := make([]relation.CheckPair, len(relations))
-	items := make([]*authzedpb.BulkCheckPermissionRequestItem, 0, len(relations))
+	items := make([]*authzedpb.CheckBulkPermissionsRequestItem, 0, len(relations))
```
rohilsurana, line 299: Have we tested this with currently default set version of spicedb in the docker as well as k8s charts?

rohilsurana (approved):

internal/store/spicedb/relation_repository.go
AmanGIT07: Yes — the e2e suite exercises the migrated batch check against the SpiceDB image pinned in this repo, including per-item allow/deny results.

`CheckBulkPermissions` graduated to the stable v1 API in SpiceDB v1.30.0 — [release notes](https://github.com/authzed/spicedb/releases/tag/v1.30.0) ("CheckBulkPermission has now graduated!"), implemented in [authzed/spicedb#1792](https://github.com/authzed/spicedb/pull/1792) — so any server at or above v1.30.0 serves it. Both the defaults in this repo and our deployment satisfy that. I also confirmed it live via gRPC reflection against the deployed server version's image: `authzed.api.v1.PermissionsService.CheckBulkPermissions` is listed.

Server-side, both API names delegate to the same internal bulkcheck implementation, so engine behavior is unchanged — this PR just calls the stable name directly.

No version bump needed anywhere.

PR #1797, opened 2026-07-23
fix(reconcile): validate the permission namespace charset at plan time

rohilsurana:
## What

Closes the Permission kind's Rule 2 and Rule 4 gaps on the rules-v2 audit. Reconcile-side only, one commit.

## The gap

A permission's slug is `service_resource_verb`, built by joining the three parts with `_` (`schema.FQPermissionNameFromNamespace`). The verb was charset-checked but the **namespace** was not, and `_` is legal inside a namespace part. So the map from `{namespace, name}` to slug was not one-to-one: `resource_order/item` + `get` and `resource/order_item` + `get` both flatten to `resource_order_item_get`.

The diff keys on that slug. So a file entry whose namespace differs from a stored one but collides on the slug was treated as already present: the plan showed zero ops while the desired state actually differed (Rule 2), and because nothing ever detected the mismatch it never self-corrected on re-run (Rule 4).

## The fix

- New `schema.IsValidPermissionNamespace`: the namespace must be `service/resource` with each part lowercase alphanumeric, no underscore.
- Called from `validatePermissionSpec` (server-free), so it runs in both `Validate` (whole file, before any apply) and the diff. An ambiguous namespace now fails the plan up front with a clear message.

Forbidding underscore-in-a-part makes the slug one-to-one, which closes both cells at once. The rule sits inside SpiceDB's own object-type grammar (lowercase, and although SpiceDB allows `_`, Frontier's slug scheme cannot represent it unambiguously), so it never rejects a namespace the server could actually store, apart from the ambiguous underscore ones the fix targets.

## Scope decisions

- **Reconcile-side only.** This fully closes the reconcile Rule 2/4 gap. The server's `CreatePermission` still has no namespace charset check, so a direct API call could create an ambiguous namespace outside the reconcile flow; adding the same check there is a reasonable follow-up but is out of scope here.
- **Pre-existing malformed rows.** If a database already holds a custom permission whose namespace has an underscore in a part (SpiceDB allows it), `Export` will emit it and the new `Validate` will reject re-reconciling it. That is intended: the loud failure names the offending namespace so an operator renames it (rename = delete + create, per Rule 1). No silent absorption.
- **Charset is minimal** (lowercase alphanumeric, no underscore), consistent with the existing name check, rather than fully mirroring SpiceDB's no-leading-digit rule.

## Testing

- Test first. The headline: a file namespace that flattens to a stored permission's slug now fails instead of planning zero ops. Plus: underscore-in-a-part and uppercase namespaces are rejected, valid custom namespaces (`resource/aoi`, `user/project`, `org/user`, `compute/disk`) still pass, the check fails in `Validate` before any server call, and a direct unit test for `IsValidPermissionNamespace`.
- `go build`, `go vet`, `go test ./internal/reconcile/... ./internal/bootstrap/schema/...`, `gofmt`, and `golangci-lint` all pass.

## Status

Closes the last two audit cells (Permission Rules 2 and 4). Targets `main`.

### DIFF
--- a/internal/bootstrap/schema/schema.go
+++ b/internal/bootstrap/schema/schema.go
@@ -311,6 +311,30 @@ func IsValidPermissionName(name string) bool {
 	return true
 }
 
+// IsValidPermissionNamespace checks that a custom permission namespace is in
+// service/resource form with each part lowercase alphanumeric. An underscore is
+// forbidden inside a part: FQPermissionNameFromNamespace joins service, resource,
+// and the verb with "_", so an underscore in a part would let two different
+// namespaces flatten to the same slug. Uppercase is forbidden because SpiceDB
+// object type names are lowercase, so it could never be stored anyway.
+func IsValidPermissionNamespace(namespace string) bool {
+	parts := strings.Split(namespace, "/")
+	if len(parts) != 2 {
+		return false
+	}
+	for _, part := range parts {
+		if part == "" {
+			return false
+		}
+		for _, r := range part {
+			if !((r >= 'a' && r <= 'z') || (r >= '0' && r <= '9')) {
+				return false
+			}
+		}
+	}
+	return true
+}
+
 func IsPlatformPermission(name string) bool {
 	name = strings.ToLower(name)
 	return name == PlatformSudoPermission || name == PlatformCheckPermission
--- a/internal/bootstrap/schema/schema.go
+++ b/internal/bootstrap/schema/schema.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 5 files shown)


internal/bootstrap/schema/schema.go
```diff
@@ -311,6 +311,30 @@ func IsValidPermissionName(name string) bool {
...
+	parts := strings.Split(namespace, "/")
+	if len(parts) != 2 {
+		return false
+	}
+	for _, part := range parts {
+		if part == "" {
+			return false
+		}
+		for _, r := range part {
+			if !((r >= 'a' && r <= 'z') || (r >= '0' && r <= '9')) {
```
rohilsurana: Wouldn't regex be simpler to read?

internal/reconcile/permission.go
```diff
@@ -48,6 +48,13 @@ func validatePermissionSpec(s PermissionSpec) error {
 	if parts := strings.Split(s.Namespace, "/"); len(parts) != 2 || parts[0] == "" || parts[1] == "" {
 		return fmt.Errorf("namespace %q must be in service/resource form", s.Namespace)
 	}
+	// The slug joins service, resource, and verb with "_", so an underscore inside
+	// a namespace part would make two different namespaces flatten to the same slug
+	// and be silently treated as one. Require each part to be lowercase alphanumeric
+	// so the slug is one-to-one and a plan cannot mistake one namespace for another.
+	if !schema.IsValidPermissionNamespace(s.Namespace) {
+		return fmt.Errorf("invalid namespace %q (each of service/resource must be lowercase alphanumeric, no underscore)", s.Namespace)
```
rohilsurana:
```suggestion
		return fmt.Errorf("invalid namespace %q (each of service/resource must be lowercase alphanumeric,)", s.Namespace)
```

internal/bootstrap/schema/schema.go
rohilsurana: Done in 319ad0f2. IsValidPermissionNamespace is now a single regex, `^[a-z0-9]+/[a-z0-9]+$` (exactly two lowercase-alphanumeric parts, one slash), instead of the split plus rune loop.

internal/reconcile/permission.go
rohilsurana: Applied. Dropped "no underscore" since lowercase alphanumeric already excludes it, and cleaned the trailing comma so it reads "must be lowercase alphanumeric".

PR #1799, opened 2026-07-23
fix(admin): Admin org design fixes

Shreyag02:
## Summary
Design polish for the IAM Admin Console **organization flow**, plus the design-review follow-ups. Covers the org listing, the org detail tabs, and the project members dialog.

Frontend only — no backend/Go changes. Mostly `web/sdk/admin/views/organizations/*`, with the navbar and empty-state work also reaching `audit-logs`, `invoices` and `users`.

## Changes

**Navbars**
- Hide the primary action while search is active — "New organization" on the org listing, "Invite user" on the users listing. Folded into the existing `showSearch` branch rather than adding a second conditional.
- Navbar height pinned to 48px (`--rs-space-11`) on **all six** admin navbars. Four were on 56px.
- The duplicated `.navbar` block is now one shared class in `admin/components/navbar.module.css` (6 copies → 1, `.nav-chip` 2 → 1).

**Zero states**
- Copy now matches Figma across the org listing, apis, invoices, projects, tokens and pat. Headings are singular per the design (`Invoice`, `Project`, `API`, `PAT`); Tokens stays plural, as designed.
- The org listing had no zero state at all and gains one, with both design actions: **Add new organization** and **Documentation**.
- Extracted the shared `DocumentationButton`, used on all six zero states.
- Icons: added `BanknotesIcon` (invoices) and `ProjectsIcon` (projects).
- Folded the duplicated `.zero-state` padding into `.zero-state-container`.

**Empty states**
- Copy matched to Figma. The subheading previously merged the keyword and filter variants into one hybrid string; members now uses the filter copy per its frame, the rest use the keyword copy.

**Detail listings**
- Row action kebab is revealed on row hover, on focus-within, and while its menu is open.
- Project "Add member" is disabled when the org has a single member; the row stays highlighted while its menu is open.
- "View Invoice" link no longer underlined.

## Bug fixed along the way

**The empty state was unreachable on every org detail tab.** `DataTable` seeds its internal query once at mount and never re-reads the prop, so it could not see the search — on the detail tabs the search lives on the org context, outside the table. `hasActiveQuery` therefore stayed false and a no-match search rendered the **zero** state instead of "No X found".

Fixed by deciding zero-vs-empty in the view rather than via the `zeroState` prop, across members, projects, invoices, tokens, apis and pat. The org listing was already correct — its `DataTable.Search` sits inside the table.

## Technical Details
- **Hover reveal** is pure CSS: the action column is `opacity: 0`, going to `1` on `tr:hover`, `:focus-within`, and `:has([data-popup-open])`. The open row is highlighted via `tr:has([data-test-id="…"][data-popup-open])`.
- **`canAddMember`** is derived from `orgMembersMap` (> 1 member) and threaded through the project actions menu and the members dialog.
- **`DocumentationButton`** renders as an anchor via Apsara's `render` prop, so it is a real link, and takes an optional `href` for future deep links.
- **Icons** are vendored as local components rather than adding a dependency, matching the existing `CpuChipIcon` precedent. `BanknotesIcon` is Heroicons' 24px outline rescaled to the 16 grid, with the stroke scaled to match.
- **`.module.css` imports stay relative** — `esbuild-css-modules-plugin` joins the specifier onto the importer's directory and ignores tsconfig `paths`, so the `~` alias cannot be used for stylesheets. TS/TSX imports in the touched files use `~`.
- Restored `Dialog.CloseButton` on the project members and PAT details dialogs, and dropped the `initialFocus` override that existed only because the close button had been removed. `pat-details-dialog.tsx` is now unchanged from main.

## Test Plan
- [x] Build and type checking passes (`tsup` build, `tsc --noEmit`, eslint — 0 errors)

### Manual checks
- Navbars: primary action hidden during search on the org and users listings; all six navbars 48px with no clipping.
- Zero states: empty tabs show the design copy; org listing zero state shows both buttons and Documentation opens the docs in a new tab.
- **Empty states: a no-match search shows "No X found", not the zero state** — on all six detail tabs.
- Kebab appears only on hover / while open; row stays highlighted with the menu open.
- Single-member org: project "Add member" is disabled.
- Both dialogs can be closed via the close button.
- Invoices: "View Invoice" has no underline.

Not applicable — frontend only, touches no `*_repository.go` / `goqu.*` code.

---

### Design items
| # | Item | Status | What changed |
|---|------|--------|--------------|
| 1 | Navbar: hide primary action during search + 48px height | ✅ Fixed | `list/navbar.tsx`, `users/list/navbar.tsx`, `components/navbar.module.css` |
| 2 | Side-panel heading aligned with cells | ✅ Already correct | No change |
| 3 | Listing avatar 4px radius (not circular) | ✅ Already correct | No change |
| 4 | Listing background = Background/Base/Primary | ✅ Already correct | No change |
| 5 | Drawer heading text = Foreground/Base/Primary | ✅ No change needed | Title already Primary |
| 6 | Role field: strip "Org" prefix | ✅ No change needed | Already handled in `schema.go` by #1705; a prefixed title means stale DB data, fixed via `frontier reconcile` |
| 7 | Zero state for empty tabs + hide headers | ✅ Fixed | `apis`, `invoices`, `members`, `projects`, `tokens`, `pat`, `list` |
| 8 | Listing kebab visible only on hover | ✅ Fixed | `members`, `projects`, `projects/members` columns + module.css |
| 9 | Disable project "Add member" for single-member orgs + keep row selected | ✅ Fixed | `projects/columns.tsx`, `projects/index.tsx`, `projects/members/*` |
| 10 | "View Invoice" link without underline | ✅ Fixed | `invoices/columns.tsx`, `invoices.module.css` |


web/sdk/admin/utils/helper.ts
```diff
@@ -41,6 +41,13 @@ export function randomSuffix(length = 4): string {
   return out;
 }
 
+// Role titles come prefixed with the scope (e.g. "Organization Owner").
+// For display we only want the bare title: "Owner", "Member", "Admin".
+export function formatRoleTitle(title?: string): string {
+  if (!title) return '';
+  return title.replace(/^organization\s+/i, '');
+}
+
```
rohanchkrabrty: We should be handling this by updating configs, and not in code

web/sdk/admin/views/organizations/details/apis/index.tsx
```diff
@@ -36,6 +36,20 @@ const NoCredentials = () => {
...
 
+const ZeroState = () => {
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
+        icon={<InfoCircledIcon />}
+        heading="API Credentials"
+        subHeading="Service accounts provide programmatic API access on behalf of this organization."
```
rohanchkrabrty: Copy doesn't match

web/sdk/admin/views/organizations/details/apis/apis.module.css
```diff
@@ -7,6 +7,14 @@
...
 }
 
+.zero-state-container {
+  padding: var(--rs-space-10);
+}
+
+.zero-state {
+  padding-top: var(--rs-space-17);
+}
+
```
rohanchkrabrty: we can put this into the container class itself instead of a separate class

web/sdk/admin/views/organizations/details/invoices/index.tsx
```diff
@@ -70,6 +70,20 @@ const NoInvoices = () => {
...
+const ZeroState = () => {
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
+        icon={<FileTextIcon />}
+        heading="Invoices"
+        subHeading="Invoices generated for this organization's billing activity will appear here."
+      />
```
rohanchkrabrty: copy doesnt match

web/sdk/admin/views/organizations/details/invoices/index.tsx
```diff
@@ -70,6 +70,20 @@ const NoInvoices = () => {
   );
 };
 
+const ZeroState = () => {
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
```
rohanchkrabrty: class is not needed as mentioned above

web/sdk/admin/views/organizations/details/members/index.tsx
```diff
@@ -54,6 +54,20 @@ const NoMembers = () => {
...
 
+const ZeroState = () => {
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
+        icon={<UsersIcon />}
+        heading="Members"
+        subHeading="Members are users who belong to this organization and can access its resources."
```
rohanchkrabrty: use terminology function for organization

web/sdk/admin/views/organizations/details/projects/members/index.tsx
```diff
@@ -199,14 +203,16 @@ export const ProjectMembersDialog = ({
         />
       ) : null}
       <Dialog open onOpenChange={onClose}>
-        <Dialog.Content className={styles["dialog-content"]}>
+        <Dialog.Content
+          className={styles["dialog-content"]}
+          initialFocus={() => initialFocusRef.current}
```
rohanchkrabrty: Why is the `initialFocus` overriden here? If it's to move focus from the Dialog.Close then it's intended UX and should be maintained.

Let me know if there was different reason for this

web/sdk/admin/views/organizations/details/projects/index.tsx
```diff
@@ -51,6 +51,21 @@ const NoProjects = () => {
...
+  const t = useTerminology();
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
+        icon={<FileIcon />}
+        heading={t.project({ plural: true, case: "capital" })}
+        subHeading={`${t.project({ plural: true, case: "capital" })} group resources and members within this ${t.organization({ case: "lower" })}.`}
+      />
```
rohanchkrabrty: copy doesn't match

web/sdk/admin/views/organizations/details/tokens/index.tsx
```diff
@@ -45,6 +45,20 @@ const NoTokens = () => {
...
+  return (
+    <div className={styles["zero-state-container"]}>
+      <EmptyState
+        variant="empty2"
+        className={styles["zero-state"]}
+        icon={<CoinIcon />}
+        heading="Tokens"
+        subHeading="Token transactions for this organization's consumable balance will appear here."
+      />
+    </div>
```
rohanchkrabrty, line 57: copy doesn't match

web/sdk/admin/views/organizations/list/list.module.css
```diff
@@ -5,7 +5,7 @@
   display: flex;
   align-items: center;
   justify-content: space-between;
-  min-height: var(--rs-space-12);
+  height: var(--rs-space-11);
```
rohanchkrabrty: This needs to be done for other pages too. We can create reusable component or a global class to achieve this.

we can also create a css var `--admin-navbar-height: 48px` and then reuse it in other places

web/sdk/admin/views/organizations/list/navbar.tsx
rohanchkrabrty: we can put the conditional logic to show the new button here, instead of creating multiple if else

rohanchkrabrty (changes requested):

web/sdk/admin/utils/helper.ts
Shreyag02: It turns out the helper wasn't doing anything anyway. 
On further deep dive #1705  already dropped the prefix in `schema.go`, so the regex had nothing to strip. 
If the prefix does turn up somewhere, it'll be leftover data rather than code.

web/sdk/admin/views/organizations/details/projects/members/index.tsx
Shreyag02: yes it was to move focus from the Dialog.Close. Reverted as it's intended UX and should be maintained.

rohanchkrabrty (approved):

PR #1800, opened 2026-07-24
refactor(deleter): move custom-resource policy cleanup to resource service

whoAbhishekSah:
## What

When a user is deleted, the deleter walked the user's policies itself and deleted the ones on custom resources. That was the last spot where it removed access without going through a domain service.

This moves that cleanup into the resource service, as a new method `RemovePrincipalAccess`. The deleter now calls it once per user and no longer touches policies directly for this path.

## Behavior

Unchanged. Deleting a user still removes their access to resources in the org's projects. Org, project, and group membership is still handled by the membership cascade.


whoAbhishekSah:
Tested the `DeleteUser` RPC end to end on a local server built from this branch.

**Setup**
- Local Postgres + SpiceDB. Custom permissions reconciled first so custom resource namespaces exist (`resource/aoi`, `resource/workflow`).
- 2 orgs, 3 projects, 5 users. Every user joined through `CreateOrganizationInvitation` + `AcceptOrganizationInvitation`.
- 4 custom resources created. Creating a resource writes no policy row — only SpiceDB tuples (`owner`, `project`). So all resource policies were created explicitly.
- 2 custom org roles carrying `resource_aoi_*` permissions. Policies assigned at org, project, and resource level.
- After each call, verified the `policies` and `users` tables and SpiceDB `relation_tuple` rows directly.

**Results**

| Case | Access before delete | State after `DeleteUser` | Result |
|---|---|---|---|
| user in 1 org | org policy, project policy, 2 resource policies (custom role on `resource/aoi` resources in two projects) | 0 policy rows, 0 SpiceDB tuples, user row gone | pass |
| user in 2 orgs | a resource policy in each org's project, memberships in both orgs | policies in both orgs gone, memberships gone, user row gone | pass |
| user removed from org first (`RemoveOrganizationMember`), then deleted | 1 resource policy left over from the org removal (that RPC never reaches the deleter) | user row and SpiceDB tuples gone, but the policy row stays orphaned | pre-existing gap, unchanged by this PR |

**Note on the gap**: `RemovePrincipalAccess` runs per org the user is still a member of. A user who already left the org keeps orphan resource policy rows in Postgres after deletion. This is tracked in #1801 and will be handled separately.

rohilsurana (approved):

PR #1807, opened 2026-07-27
fix(membership): clean custom-resource policies when removing an org member

### LINKED ISSUE #1801: RemoveOrganizationMember leaves the member's custom-resource policies behind
## What happens

Removing a user from an organization with the `RemoveOrganizationMember` RPC deletes their org, project, and group policies and membership relations. But their policies on custom resources (for example `resource/aoi`) in the org's projects are not touched. The policy row and its SpiceDB rolebinding tuple stay live.

## Why

The handler calls `membership.RemoveOrganizationMember`, and the membership cascade only covers org, project, and group scopes. Custom-resource policy cleanup lives in `resource.RemovePrincipalAccess` (#1800), which is only called from the deleter during user deletion (`DeleteUser`). The org-removal path never reaches it.

## Consequence

- An ex-member keeps a live policy on a resource inside an org they were removed from. If they rejoin, or if the permission check does not require org membership, the old access still applies.
- If the user is later deleted with `DeleteUser`, the policy row is orphaned: the deleter cleans resource policies per org the user is still a member of, and the user is no longer in any. The user row and SpiceDB tuples are deleted, but the policy row stays in Postgres pointing at a deleted user.

## Reproduce (verified on a local server built from #1800)

1. Create an org, a project, and a custom resource (`resource/aoi`) in the project.
2. Add a user to the org. Create a custom role with `resource_aoi_*` permissions and a policy for the user on the resource.
3. Call `RemoveOrganizationMember` for the user.
4. The user's org and project policies are gone. The resource policy row and its `app/rolebinding` tuple remain.
5. Call `DeleteUser` for the same user. The user row and tuples are gone; the resource policy row remains, orphaned.

## Possible fix

Have the org-removal path clean the member's resource policies too, for example by calling `resource.RemovePrincipalAccess` with the org's project IDs from `membership.RemoveOrganizationMember` (or from the handler). Separately, `DeleteUser` could delete any remaining policies by principal id as a final sweep so no orphan rows survive.

whoAbhishekSah:
Fixes #1801

## Problem

When a member is removed from an organization, we delete their org, project, and group access. But we leave their access to custom resources inside that org's projects alone, in both the database and SpiceDB.

So a removed member keeps working access to resources in an org they are no longer part of. And if that user is deleted later, the leftover access row is never cleaned up, because by then they are not a member of any org.

## Fix

The member removal now clears custom-resource access as well. It happens in one shared place, so every path that removes a member gets it — the API call, service user deletion, and user deletion.

The cleanup runs only after the last-owner check passes, so a rejected removal still changes nothing.


whoAbhishekSah:
## Tested end to end on a local server

Drove the real RPCs against a running Frontier with Postgres and SpiceDB, and checked all three places the state lives: the `policies` table, the SpiceDB tuples, and a live permission check.

### Setup

- Custom resource type `widget/item`, with permissions `manage`, `get`, `delete`.
- Custom org role `widget-admin` holding `app.organization.update` + `widget.item.manage` + `widget.item.get`.
- **Org A** with two projects: `res1` in project 1, `res2` in project 2. Two projects on purpose, so a fix that only cleaned one would show up.
- **Org B** with one project holding `res3`. This is the control for an unrelated org.
- Four users, logged in over mail OTP: alice (owner), bob, charlie, dave.

Grants before the removal:

| Who | Org A membership | Custom-resource grants |
|---|---|---|
| bob | org viewer | `res1`, `res2` (Org A) and `res3` (Org B) |
| charlie | org viewer | `res1` |
| dave | org viewer | `res1` |
| alice | owner | `res1` |

### What we did

Called `RemoveOrganizationMember(Org A, bob, app/user)` as alice. Returned 200.

### What we got

| bob's grant | Before | After |
|---|---|---|
| Org A org-viewer policy | present | deleted |
| `res1` — project 1, Org A | present | deleted |
| `res2` — project 2, Org A | present | deleted |
| `res3` — Org B | present | kept |

- ✅ Both of bob's custom-resource policies inside Org A are gone — the policy row and the `granted` tuple on the resource.
- ✅ `res2` proves the cleanup covers every project in the org, not just the first one.
- ✅ `res3` kept. The cleanup does not reach into another org.
- ✅ charlie's and dave's grants on `res1` kept. It filters by person, not by resource.
- ✅ No orphan tuples. The `bearer` and `role` tuples on bob's rolebindings are gone too, so nothing dangles in SpiceDB.
- ✅ Rows are really deleted, not soft deleted — `deleted_at` is not set on anything, and the rows are absent. That covers the second half of #1801, where deleting the user later orphaned the policy row.
- ✅ `CheckResourcePermission` for `widget.item.manage` agrees: bob went from allow to deny on `res1` and `res2`, and still gets allow on `res3`. charlie and dave unchanged.

### Last-owner guard

Gave alice a grant on `res1`, then tried to remove her as the only owner of Org A.

- ✅ Rejected with `failed_precondition: this is the last owner of the organization`, and none of her four policies changed, including the one on the custom resource. So the new call really does sit after the guard, and a refused removal has no side effects.

### Unit tests

- ✅ `core/membership`, `core/deleter`, `core/resource` all pass.

### One thing found on the side, not from this change

Creating a policy on a custom resource whose namespace has no `delete` permission fails with a bare 500. The real cause only shows up in the server log:

```
relation/permission `delete` not found under definition `widget/item`
```

The authz check on `CreatePolicy` asks for `delete` on the target. Adding the permission fixes it. Worth a separate issue — the caller gets no hint about which permission is missing.

rohilsurana (approved):

PR #1809, opened 2026-07-27
refactor(schema): grant org and group access through roles only

whoAbhishekSah:
## Problem

An org owner gets their permissions in two ways at the same time:

1. **Through their role.** The Owner role holds the `app_organization_administer` permission.
2. **Through a direct `owner` link** between the user and the org, stored in SpiceDB. This link skips roles completely.

When someone joins an org or a group, Frontier writes both.

The second path is the problem. We can now change what a predefined role contains (using `frontier reconcile`). But shrinking the Owner role does nothing to real owners — their direct `owner` link still grants everything. The `member` link has the same issue. Custom permissions too: the schema generator adds the same `owner` shortcut to every custom permission it creates.

## Fix

Roles are now the only way to get org and group permissions.

- `base_schema.zed`: every org and group permission ended with `+ owner` (some also `+ member`). Removed those. Also deleted the `membership` permission — it was built only from these two links and nothing uses it.
- `generator.go`: custom permissions no longer get the `owner` shortcut at the org level. The creator of a single resource still keeps access to that resource — that is a different `owner` and is untouched.
- Deleted the unused `MembershipPermission` constants (Go and JS SDK).

Owners lose nothing: the Owner role has `app_organization_administer`, and every permission still accepts that.

**One real change:** the `member` link used to give every org member read access for free. Now a role must list `app_organization_get` for the user to see the org (`app_group_get` for groups).

The `owner` and `member` links are still declared in the schema and still written on every membership change. They just grant nothing now. Deleting them fully is a follow-up PR, because SpiceDB refuses to drop a relation while stored data still uses it.

### DIFF
--- a/core/organization/organization.go
+++ b/core/organization/organization.go
@@ -19,7 +19,6 @@ const (
 	Disabled State = "disabled"
 
 	AdminPermission = schema.UpdatePermission
-	AdminRelation   = schema.OwnerRelationName
 	AdminRole       = schema.RoleOrganizationOwner
 	MemberRole      = schema.RoleOrganizationViewer
 )
--- a/internal/bootstrap/generator.go
+++ b/internal/bootstrap/generator.go
@@ -139,7 +139,6 @@ func ApplyServiceDefinitionOverAZSchema(serviceDef *schema.ServiceDefinition, ex
 			{
 				// for org
 				nsRel, err := aznamespace.Relation(fqPermissionName, aznamespace.Union(
-					aznamespace.ComputedUserset("owner"),
 					aznamespace.TupleToUserset("platform", "superuser"),
 					aznamespace.TupleToUserset("granted", "app_organization_administer"),
 					aznamespace.TupleToUserset("granted", fqPermissionName),
--- a/internal/bootstrap/schema/base_schema.zed
+++ b/internal/bootstrap/schema/base_schema.zed
@@ -35,34 +35,32 @@ definition app/organization {
 
 	// permissions
     // org
-    permission membership = member + owner
-
-    permission delete = platform->superuser + granted->app_organization_administer + granted->app_organization_delete + owner
-    permission update = platform->superuser + granted->app_organization_administer + granted->app_organization_update + owner
-    permission get = platform->superuser + granted->app_organization_administer + granted->app_organization_get + granted->app_organization_update + owner + member
-    permission rolemanage = platform->superuser + granted->app_organization_administer + granted->app_organization_rolemanage + owner
-    permission policymanage = platform->superuser + granted->app_organization_administer + granted->app_organization_policymanage + owner
-    permission projectlist = platform->superuser + granted->app_organization_administer + granted->app_organization_projectlist + owner
-    permission grouplist = platform->superuser + granted->app_organization_administer + granted->app_organization_grouplist + owner
-    permission projectcreate = platform->superuser + granted->app_organization_administer + granted->app_organization_projectcreate + owner
-    permission groupcreate = platform->superuser + granted->app_organization_administer + granted->app_organization_groupcreate + owner
-	permission invitationlist = platform->superuser + granted->app_organization_administer + granted->app_organization_invitationlist + owner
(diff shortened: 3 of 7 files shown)


whoAbhishekSah:
Tested on a local Frontier (Postgres + SpiceDB), before and after the change.

**Setup:** one fresh org and five users, each covering a different access path:

- **alice** — a normal owner. She created the org, so she has both the Owner role and the direct `owner` link.
- **charlie** — owner through the role only. We created his Owner policy directly, so he has no `owner` link.
- **eve** — org member with a custom role that only grants `app_organization_projectlist`. Her role does not grant org read.
- **dave** — org viewer and member of a group. The group holds the `app_project_viewer` role on a project.
- **bob** — org viewer and *owner* of that same group.

We also added custom permissions (`resource/aoi`, `user/project:createaoi`) so the generator path is tested too.

**Test 1 — before the change.** We shrank the Owner role down to only `app_organization_get` using reconcile. charlie lost update and delete on the org. alice kept everything. Same role, different result — her direct `owner` link was bypassing the role. The custom permission behaved the same: alice kept it, charlie lost it.

**Test 2 — after the change, roles untouched.** alice, charlie, and dave: no difference, all access intact. eve can no longer read the org — her role never granted read, the `member` link was giving it to her for free. She still has `projectlist`, so her role itself works fine.

**Test 3 — after the change, Owner role shrunk again.** alice and charlie now behave exactly the same: both keep `get`, both lose update, delete, and the custom permission. alice's `owner` link is still stored in SpiceDB, but it grants nothing. Resetting the role gave both of them everything back.

**One more check:** dave (group member) can read the project through the group's role. bob (group owner) cannot. That is an old bug, not caused by this PR — group owners get an `owner` link instead of a `member` link, but roles granted to a group only reach its members. The follow-up PR fixes it.

Unit tests: full suite passes. The golden schema file was regenerated.

whoAbhishekSah:
Tested this locally end to end (server built from this branch, all flows driven as normal users through the ConnectRPC APIs).

**Setup**
- Fresh org created by a normal user (gets `app_organization_owner`).
- Two more users added through the invitation flow with roles attached (`invite_with_roles` enabled): one `app_organization_manager`, one `app_organization_viewer`. Roles verified via `ListOrganizationUsers`.
- A fourth user stayed outside the org for negative checks.
- New custom permission namespace `database/instance` (get/edit/delete) created after the schema change, so the generator path without the org-owner userset is what produced its definition.
- Projects and custom resources created by each role tier; one group with a member added via `SetGroupMemberRole`.

**Checks — `CheckResourcePermission`, 14 permissions x 4 users**

| check | owner | manager | viewer | outsider |
|---|---|---|---|---|
| org get | allow | allow | allow | deny |
| org update | allow | allow | deny | deny |
| org delete | allow | deny | deny | deny |
| org projectcreate | allow | allow | allow* | deny |
| project get/update | allow | allow | deny | deny |
| project delete | allow | deny | deny | deny |
| own project (creator) | allow | allow | allow | deny |
| custom resource (creator) | allow | allow | allow | deny |
| custom resource (non-creator) | allow | deny | deny | deny |

*viewer role on this server was widened with `app_organization_projectcreate` via reconcile — the allow came purely from the role, which is the point of this change.

Every allow traced to a role permission; every deny matched the role definition. The `owner`/`member` tuples (still written) contributed nothing.

**Listings**
- `ListOrganizationUsers`: all members can list, outsider denied.
- `ListProjectUsers`: shows only explicit project members; manager can view via org-scoped `app_project_get` without being listed; viewer denied on projects they are not in.
- `ListProjectsByCurrentUser` with/without `non_inherited`: inherited results follow role permissions (owner/manager see all, viewer only their own); explicit results collapse to the project each user created.
- Group listings: manager lists org groups via `grouplist` but cannot read inside one (no `app_group_get` in the role — same denial before this change, since it was never the group's owner/member relation); group member reads the group purely via `app_group_member`; outsider denied everywhere.
- Platform superuser path intact (bootstrap service account passes org delete and custom resource get).

No behavior regressions found. The pre-existing group-owner gap (owners miss group-granted roles because the bearer is `group#member`) is untouched here and lands in part 2.

rohilsurana (approved):

PR #1811, opened 2026-07-28
fix(billing): ignore inactive prices in checkout, plan matching, and invoicing

rohilsurana:
## What

Make every purchase and billing path ignore prices that are marked inactive. An inactive price is a retired one, and it must never be used for a new checkout, subscription, plan change, plan match, or invoice.

## Why

A price can be retired by marking it inactive rather than deleted, since provider (Stripe) prices are immutable and cannot be removed. But nothing in the code filtered by state, so a retired price would still flow into checkout line items, plan changes, and plan matching. Stripe rejects an archived price in a new checkout, so a product with a retired price could fail to sell or bill the wrong amount. This change makes the system safe to hold inactive prices, which is a prerequisite for managing a product's prices over time.

## Changes

- New `Price.IsActive()` helper and `PriceStateActive` / `PriceStateInactive` constants. Only the active state (or an empty state, which an in-memory price carries before it is stored) counts as usable; any other value fails closed.
- Checkout: plan checkout, one-time product checkout, and subscription sync skip inactive prices when building line items.
- Subscription: `ChangePlan` skips inactive prices when building the next subscription phase.
- Plan matching: `hasMatchingPrice` no longer counts an inactive price, and its error now says "no active prices".
- `Plan.IsFree()` ignores retired prices, so retiring a paid price can make a plan free.
- Credit overdraft invoice: the rate comes from the first active price, and the product must have at least one active price.
- Empty guards: if a plan or product has billable products but no active price to bill, the flow fails with a clear error instead of sending an empty request to Stripe. A credit-only plan still passes, since it has nothing to bill by design.

## Tests

`Price.IsActive()` and `Plan.IsFree()` are unit-tested, including the retired-price and unknown-state cases. The call-site changes apply those helpers as guards, and the existing billing test suites pass.

## Notes

- Behavior is unchanged today, since nothing sets a price inactive yet. This is groundwork for a follow-up that lets `UpdateProduct` retire a price.
- Two items are intentionally out of this PR and must land before any price is actually retired: the web SDK still reads `prices[0]` and would show a retired rate to users, and the plan boot loader matches a price by name including retired ones. Both belong with the retire feature.

### DIFF
--- a/billing/product/product.go
+++ b/billing/product/product.go
@@ -117,10 +117,12 @@ const (
 
 // IsActive reports whether the price can be used for new purchases. A retired
 // price is marked inactive, so an inactive price must be left out of checkout
-// line items, plan matching, and invoicing. An empty state is treated as active,
-// since that is the database default for a newly created price.
+// line items, plan matching, and invoicing. Only the active state counts as
+// usable, along with an empty state, which an in-memory price carries before it
+// is stored (the stored default is active). Any other value is treated as not
+// active, so an unexpected state fails closed rather than being billed.
 func (price Price) IsActive() bool {
-	return price.State != PriceStateInactive
+	return price.State == "" || price.State == PriceStateActive
 }
 
 type BillingScheme string
--- a/billing/product/product.go
+++ b/billing/product/product.go
@@ -108,6 +108,21 @@ func (price Price) IsLicensed() bool {
 	return price.UsageType == PriceUsageTypeLicensed
 }
 
+// Price states. A new price is active. A superseded price is retired by marking
+// it inactive rather than deleting it, since provider prices are immutable.
+const (
+	PriceStateActive   = "active"
+	PriceStateInactive = "inactive"
+)
+
+// IsActive reports whether the price can be used for new purchases. A retired
+// price is marked inactive, so an inactive price must be left out of checkout
+// line items, plan matching, and invoicing. An empty state is treated as active,
+// since that is the database default for a newly created price.
+func (price Price) IsActive() bool {
+	return price.State != PriceStateInactive
+}
+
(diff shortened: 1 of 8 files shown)


billing/product/product.go
```diff
@@ -108,6 +108,23 @@ func (price Price) IsLicensed() bool {
...
+)
+
+// IsActive reports whether the price can be used for new purchases. An inactive
+// price must be left out of checkout line items, plan matching, and invoicing.
+// Only the active state counts as usable, along with an empty state, which an
+// in-memory price carries before it is stored (the stored default is active).
+// Any other value is treated as not active, so an unexpected state fails closed
+// rather than being billed.
+func (price Price) IsActive() bool {
+	return price.State == "" || price.State == PriceStateActive
```
whoAbhishekSah, line 125: Is it alright to consider a zero value of the state field as an active Price?

whoAbhishekSah (approved):

billing/product/product.go
rohilsurana: Yes, and it is intentional. The `state` column is `text NOT NULL DEFAULT 'active'` (migration `20231012175641`), so a price that has been stored always has a concrete state, `active` or `inactive`, never empty. `transform()` reads that NOT NULL column, so a price loaded from the DB is never empty either.

An empty state only shows up on an in-memory `Price` that has not been persisted yet, for example one built from an API request or the plan file before insert (`CreatePrice` does not set it and relies on the DB default). For those, reading it as active is correct, since `active` is exactly what the row gets on insert.

The check is deliberately narrow: only empty or `active` count as active, so any other value, including an unexpected one, is treated as not active and fails closed rather than being billed.

PR #1812, opened 2026-07-28
refactor(membership): stop writing org owner/member relations

whoAbhishekSah:
## What this is

Part 2 of removing relation-based ownership, following #1809. That PR made the schema stop *reading* the `owner` and `member` links on orgs and groups; this one makes the membership paths stop *writing* them. The split was planned from the start: reading had to stop first so that a rollback of either PR never leaves anyone without access.

## Changes

- Adding an org member or changing their role writes only the policy — no org link. (The service-user identity link stays; the schema still reads it for the `manage` permission.)
- A group member's permissions come from their group role (Team Owner or Team Member), granted through a policy like everywhere else. Joining a group also writes the `member` link, whatever the role. That link is not a permission grant — it is how roles granted *to the group* (e.g. a project role) fan out to the people in it, which is why it stays. A role change inside the group touches only the policy.
- This fixes a bug that #1809 surfaced: the group's own owner used to get an `owner` link *instead of* `member`, so roles granted to the group never reached the group's own owner. Group owners now receive group-granted roles like every other member.
- Removal paths still delete the old links, so members from before this change leave no stale data behind.

Behavior change to be aware of: group owners gain access through roles granted to their group. They were wrongly excluded before.

### DIFF
--- a/core/membership/service.go
+++ b/core/membership/service.go
@@ -146,8 +146,7 @@ func (s *Service) AddOrganizationMember(ctx context.Context, orgID, principalID,
 		return err
 	}
 
-	fetchedRole, err := s.validateOrgRole(ctx, roleID, orgID)
-	if err != nil {
+	if _, err := s.validateOrgRole(ctx, roleID, orgID); err != nil {
 		return err
 	}
 
@@ -167,26 +166,6 @@ func (s *Service) AddOrganizationMember(ctx context.Context, orgID, principalID,
 
 	createdPolicy, err := s.createPolicy(ctx, orgID, schema.OrganizationNamespace, principalID, principalType, roleID)
 	if err != nil {
-		return err
-	}
-
-	// PATs don't get relations.
-	if principalType == schema.PATPrincipal {
-		s.auditOrgMemberAdded(ctx, org, principal, roleID)
-		return nil
-	}
-
-	relationName := orgRoleToRelation(fetchedRole)
-	if err := s.createRelation(ctx, orgID, schema.OrganizationNamespace, principalID, principalType, relationName); err != nil {
-		// best-effort cleanup to avoid orphaned policy
-		if deleteErr := s.policyService.Delete(ctx, createdPolicy.ID); deleteErr != nil {
-			s.log.WarnContext(ctx, "orphaned policy: relation creation failed and policy cleanup also failed",
-				"policy_id", createdPolicy.ID,
-				"org_id", orgID,
-				"principal_id", principalID,
-				"policy_delete_error", deleteErr,
-			)
-		}
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


whoAbhishekSah:
Tested live on a local Frontier (Postgres + SpiceDB) running this branch, using a real org with users covering each path. All checks via `CheckResourcePermission` as the affected user, and link inspection via the admin Relations API.

**1. Adding an org member writes no links.** Added a fresh user as org viewer. The Relations API shows zero `app/organization` links for them, and they can still read the org — the role policy alone does the work.

**2. Group owners now receive group-granted roles (the bug fix).** New group, made the user its Team Owner, granted the group `app_project_viewer` on a project. The user can read the project. Before this change the same setup was denied — verified during the #1809 testing. Their only group link is `member`; no `owner` link was written.

**3. Org role change works with no link step.** Promoted a viewer to owner — policy replaced, nothing else written, access correct.

**4. Group role change works with no link step.** The `member` link is written on join and never changes shape; a role change replaces the policy only. Pre-existing group `owner` links are converted to `member` by the follow-up data cleanup, not by this code.

**5. Removal cleans up fully.** Removed the user from the org: group member link, role bindings, everything gone; org and project access denied afterwards.

Also reproduced the bug itself end to end: with the old-style `owner`-only link, `ListProjectsByCurrentUser` lists the group's project (both with and without `nonInherited`) while `CheckResourcePermission` denies opening it — the list reads policies, the check reads SpiceDB, and they disagree exactly for group owners. With the `member` link in place the two agree.

Unit tests updated to match: no expectations on org link writes, group add asserts a `member`-only write, role changes assert policy-only writes.

whoAbhishekSah:
Tested this live on a local server built from this branch, with a fresh org and SpiceDB tuples checked directly in the datastore after each step.

| # | Test | Result |
|---|------|--------|
| 1 | Create org — only rolebinding tuple written for creator, no `organization#owner` | ✅ |
| 2 | Add 5 members (owner/manager/viewer roles) — policies only, zero org `owner`/`member` tuples | ✅ |
| 3 | Org access by role: viewer read-only, manager can update, outsider denied | ✅ |
| 4 | `ListOrganizationUsers` with roles — all members with correct role names | ✅ |
| 5 | Org role promote/demote — permissions flip, SpiceDB tuples unchanged | ✅ |
| 6 | Project members via `SetProjectMemberRole` — policy only; access matrix correct | ✅ |
| 7 | Group add: creator and Team Owner both get `member` tuple, zero `group#owner` | ✅ |
| 8 | No `organization#member@group#member` fan-in tuple on group creation; `group#org` link present | ✅ |
| 9 | Bug fix: project role granted to the group reaches the group owner (denied before grant, allowed after) | ✅ |
| 10 | Group role promote/demote — tuples stable, permissions flip both ways | ✅ |
| 11 | Custom resource (`database/instance`) with custom roles — get/edit/delete check matrix exact per role | ✅ |
| 12 | Org owner reaches project resource through role chain alone (no direct grant, no owner relation) | ✅ |
| 13 | Remove group member — tuple, policy, and group-granted access all gone | ✅ |
| 14 | Remove org member — project policies cascade away too | ✅ |
| 15 | Legacy cleanup: hand-written `organization#member` tuple deleted by the removal path | ✅ |
| 16 | Service user: `serviceuser#org` identity link written, nothing on the org side; `manage` works (token create ok for owner/manager, denied for viewer) | ✅ |

Listing calls exercised: `ListOrganizationUsers`, `ListProjectUsers`, `ListGroupUsers`, `ListProjectsByCurrentUser`, `ListCurrentUserGroups`, `ListOrganizationsByCurrentUser`. Check calls: `CheckResourcePermission` and `BatchCheckPermission`.

One pre-existing quirk (not from this PR): `CheckResourcePermission` on `app/serviceuser` returns not_found because no permissions are registered under that namespace.

rohilsurana (approved):

PR #1814, opened 2026-07-28
docs(local-setup): add backend, admin app, and client SDK dev guide

Shreyag02:
Document local dev setup for the IAM Admin console frontend: running the backend and deps, the admin app, and the client SDK demo.

  Also adds a pnpm start command, since the guide otherwise has to tell people to build the SDK before starting an app — a step that's easy to skip, and skipping it silently serves a stale sdk/*/dist.

  `pnpm start`

  - `pnpm start:admin` → build SDK, then admin on :5173
  - `pnpm start:client` → build SDK, then client-demo on :3000
  - `pnpm start` → both in parallel

  A turbo task with dependsOn: ["^build"], so Vite can't start before tsup finishes; the SDK build is cached, so repeat runs skip it. The existing `pnpm build --filter=@raystack/frontier + pnpm dev --filter=<app>` flow is unchanged and still documented.

  Env vars

  Dropped `FRONTIER_API_URL` (admin) and `FRONTIER_ENDPOINT` (client-demo) from the guide — neither app reads them. All traffic goes over Connect-RPC, so each app needs exactly one: `FRONTIER_CONNECTRPC_URL` and `FRONTIER_CONNECT_ENDPOINT`. The now-dead /api proxy and endpoint: '/api' config are left in the code for a follow-up.

### DIFF
--- a/docs/content/docs/local-setup.mdx
+++ b/docs/content/docs/local-setup.mdx
@@ -150,25 +150,22 @@ Now editing `frontier/web/sdk/client/...` rebuilds `dist` and Vite HMR reloads.
 
 ### Backend (optional)
 
-Switched by env only in `frontier/web/apps/client-demo/.env`. It ships pointing at remote dev, so it runs with nothing else started; swap the comments to use a local backend on `:8002`. Note the **different var names** than the admin app (`FRONTIER_ENDPOINT` / `FRONTIER_CONNECT_ENDPOINT`).
+Switched by env only in `frontier/web/apps/client-demo/.env`. It ships pointing at remote dev, so it runs with nothing else started; swap the comments to use a local backend on `:8002`. Note the **different var name** than the admin app (`FRONTIER_CONNECT_ENDPOINT`, not `FRONTIER_CONNECTRPC_URL`).
 
 ```bash
 # Remote dev (DEFAULT — no local backend needed)
-FRONTIER_ENDPOINT=https://<remote-frontier-host>/frontier
 FRONTIER_CONNECT_ENDPOINT=https://<remote-frontier-host>/frontier-connect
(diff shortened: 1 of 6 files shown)


whoAbhishekSah (approved):

docs/content/docs/local-setup.mdx
```diff
@@ -0,0 +1,215 @@
...
+  ```bash
+  FRONTIER_API_URL=https://<remote-frontier-host>/frontier
+  FRONTIER_CONNECTRPC_URL=https://<remote-frontier-host>/frontier-connect
+  ```
+
+- **Case B — with local backend:** points at `:8002`; requires the backend running first (§1); isolated local data.
+
+  ```bash
+  FRONTIER_API_URL=http://localhost:8002/frontier
+  FRONTIER_CONNECTRPC_URL=http://localhost:8002/
```
whoAbhishekSah, line 97: Do we need both of them, or does just CONNECTRPC suffice?

Shreyag02: `FRONTIER_CONNECTRPC_URL` alone is enough — `FRONTIER_API_URL` is unused, so I'll remove it.

Nothing in the repo reads it: the admin bundle only gets `NEXT_PUBLIC_FRONTIER_CONNECT_URL` and `FRONTIER_CONNECTRPC_URL`, the only proxy is `/frontier-connect`, and `axios` isn't imported anywhere — there's no REST client left to point it at. The app runs fine with just that one var in `.env`.

It's a leftover from before the Connect migration, still lingering in `.env.example`, the old admin README, and `turbo.json`. I'll drop it from §2 and §5 here and clear out those three separately.

PR #1816, opened 2026-07-29
feat(reconcile): add BillingProduct kind

rohilsurana:
## What

Adds a `BillingProduct` reconcile kind. It manages billing products — their title, description, behavior, config, prices, and features — from a desired-state YAML file, using the product APIs (`ListProducts`, `CreateProduct`, `UpdateProduct`). Part of the declarative reconcile flow in RFC&nbsp;0001, one kind per resource.

```yaml
apiVersion: v1
kind: BillingProduct
spec:
  - name: tokens
    title: Tokens
    behavior: credits
    config: {credit_amount: 1, min_quantity: 1, max_quantity: 100000}
    prices:
      - {name: default, amount: 100, currency: usd}
```

## How it follows the RFC's five rules

- **R1 scope & identity** — a product is keyed by `name`, a price by `name` within its product, a feature by `name`; identity is never rewritten. Provider ids, timestamps, price state, and metadata are out of scope. A product this kind cannot represent (a tiered price, an empty title, or a name under three characters) is left untouched, and a file that names it fails the plan.
- **R2 the file is the desired state** — `title`, `description`, and `config` are full-write; a missing title fails the plan (the provider uses it as the product name and requires it). Behavior is create-only, so a change to it fails the plan. Prices converge: a new name is added, a retired name is reactivated when its fields match, an omitted active price is retired; an immutable-field change fails the plan. A product missing from the file fails the plan; `delete: true` is rejected (there is no product delete API).
- **R3 check before apply** — `Validate()` runs each entry server-free, including the proto's own enum and format rules via `protovalidate`, so a bad value fails the whole file up front rather than at apply.
- **R4 converge, not transact** — one API call per product, adds before updates, no rollback, re-apply converges.
- **R5 export inverts reconcile** — export is sorted and writes only active prices; reconciling an export plans zero changes, proven by round-trip tests.

## Depends on (all merged)

- **#1810** (UpdateProduct converges a product's prices), **#1811** (checkout, plan matching, and invoicing ignore inactive prices), **#1821** (UpdateProduct resets omitted title/description/config), **#1824** (plan_id removed from the product RPCs).
- Proton: raystack/proton#494 (allow an empty interval for one-time prices) and #495 (drop plan_id from `ProductRequestBody`), both merged and pinned.

## Notes

- Metadata is out of scope for this kind: it is never set, read, or exported here.
- Prices are a converging set of immutable provider objects, so a product's last price cannot be removed through the file; retire it by hand.

## Testing

Unit-tested at the validate, diff, and reconciler layers (convergence, retire/reactivate, immutable-change and out-of-scope failures, the export round-trip). Verified end to end against a local server plus a Stripe mock, and reviewed by an independent pass against all five rules. Build, tests, gofmt, vet, and lint are green.


internal/reconcile/billingproduct.go
```diff
@@ -0,0 +1,322 @@
...
+	"strings"
+)
+
+// KindBillingProduct is the desired-state document kind for billing products.
+const KindBillingProduct = "BillingProduct"
+
+// Known enum values for a product and its prices. They are checked at plan time
+// so a typo fails before any call, matching what the billing service accepts.
+var (
+	billingProductBehaviors = map[string]bool{"basic": true, "credits": true, "per_seat": true}
```
rohilsurana: BY having these here means we will need to update he api as well as here when we add a new behaviour or interval or schemes.

rohilsurana: Fixed in 98fea42d. Dropped the hardcoded behavior/interval/usage/scheme lists. The proto `buf.validate` constraints plus the server's `validate.NewInterceptor()` already reject bad values, and the `billing/product` package holds the type constants, so there is no longer a second place to update when a value is added.

The one value the kind still names is the tiered billing scheme, and only to reject it up front, since this kind cannot represent tiered prices (there is no `tiers` field). That is a scope decision for the kind, not a duplicated enum list.

Tradeoff worth noting: a typo'd enum (say `behavior: weird`) now fails at apply through the server's validation rather than at plan time, because a dry run does not send the write. If you'd rather catch enum typos in the plan, I can build the set from the `product` package constants instead (compiler-checked, so a rename is caught, though adding a value would still touch this file). Say the word.

internal/reconcile/billingproduct.go
```diff
@@ -0,0 +1,392 @@
...
+		return m
+	}
+	dm, am, rm := index(desired), index(active), index(retired)
+	// first reject any immutable-field change on a name that already exists,
+	// whether it is active or retired. This pass runs over every matched name
+	// before adds, reactivations, or deactivations are considered, so the
+	// outcome never depends on map iteration order.
+	for name, d := range dm {
+		if c, ok := am[name]; ok {
+			if d != c {
```
whoAbhishekSah, line 343: This compares every field, so a `metered_aggregate` difference fails the plan even for a licensed price. The server only checks that field for metered prices (`checkImmutablePriceFields` in `billing/product/service.go`). Both sides default to `"sum"`, so this usually matches. But a licensed price stored with some other value would fail the plan here for an update the server would accept. One option: blank `metered_aggregate` in `normalizeBillingPrice` when the usage type is licensed, same as the server's rule.

internal/reconcile/billingproduct.go
```diff
@@ -0,0 +1,392 @@
...
+	// "credits" when credit_amount > 0) and never changes it on update. A file
+	// that asks to change it to something the product is not cannot apply, so fail
+	// the plan rather than skip it in silence. A credit product's behavior is
+	// always "credits", so fold that in to avoid a false failure when the file
+	// still names a behavior alongside a credit amount.
+	expectedBehavior := s.Behavior
+	if s.Config.CreditAmount > 0 {
+		expectedBehavior = "credits"
+	}
+	if expectedBehavior != "" && expectedBehavior != cur.Behavior {
```
whoAbhishekSah, line 272: When the file leaves `behavior` out, this check is skipped, so it means "keep what's there". On create the server fills in `basic`. That is fine for a create-only field, but the docs page says a left-out field resets. Can we add one line to `reconcile.mdx` saying what an omitted behavior means (basic on create, keep on update)?

internal/reconcile/billingproduct_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+	specs := make([]BillingProductSpec, 0, len(current))
+	for _, c := range current {
+		entry := BillingProductSpec{
+			Name:        c.Name,
+			Title:       c.Title,
+			Description: c.Description,
+			Behavior:    c.Behavior,
+			Config:      c.Config,
+		}
+		entry.Prices = append(entry.Prices, c.Prices...)
```
whoAbhishekSah: With the new default in `CreatePrice`, licensed prices also store `metered_aggregate = "sum"`. Export copies the field as-is, so a plain seat price shows up in the YAML with `metered_aggregate: sum`, which reads like the price is metered. Could drop the field here when `usage_type` is not metered. The diff fills the default back in, so the round trip would still plan nothing.

internal/reconcile/billingproduct_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+		for _, name := range uniqueSorted(c.Features) {
+			entry.Features = append(entry.Features, BillingFeatureRef{Name: name})
+		}
+		specs = append(specs, entry)
+	}
+	return specs, nil
+}
+
+func (r *BillingProductReconciler) fetchCurrent(ctx context.Context) ([]currentBillingProduct, []string, error) {
+	resp, err := r.client.ListProducts(ctx, authReq(&frontierv1beta1.ListProductsRequest{}, r.header))
```
whoAbhishekSah, line 152: The "every product must be in the file" rule works because `ListProducts` returns everything in one response. If it ever gets pagination, products past page one would look new here and the plan would try to create them again, which fails on the unique name. Maybe worth a short comment here so that link isn't lost.

internal/reconcile/billingproduct_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+	default:
+		return fmt.Errorf("unknown op action %q", op.action)
+	}
+}
+
+// billingProductBody builds the request body for a create or update. The whole
+// desired product is sent: UpdateProduct converges the prices and features on the
+// server, so the reconciler does not compute per-price calls itself. Metadata is
+// out of scope for this kind, so it is never sent, read, or exported.
+func billingProductBody(s BillingProductSpec) (*frontierv1beta1.ProductRequestBody, error) {
```
whoAbhishekSah: This never returns an error; every path returns nil. Dropping the error return would simplify the three call sites.

internal/reconcile/billingproduct.go
```diff
@@ -0,0 +1,392 @@
...
+	if strings.TrimSpace(s.Title) == "" {
+		return fmt.Errorf("product %q must have a title", name)
+	}
+	if s.Delete {
+		return fmt.Errorf("product %q cannot be deleted: there is no product delete or deactivate API; remove the entry and archive the product by hand", s.Name)
+	}
+
+	seenPrice := map[string]struct{}{}
+	for _, p := range s.Prices {
+		name := strings.ToLower(strings.TrimSpace(p.Name))
```
whoAbhishekSah: `name` already holds the product name at the top of this function. Reusing it for the price here (and for the feature below) is easy to misread. `priceName` / `featureName` would be clearer.

internal/reconcile/billingproduct.go
```diff
@@ -0,0 +1,392 @@
...
+// KindBillingProduct is the desired-state document kind for billing products.
+const KindBillingProduct = "BillingProduct"
+
+// The set of valid behaviors, intervals, usage types, and schemes lives in the
+// proto (buf.validate) and the billing/product package, and the server rejects
+// bad values through its validate interceptor. This kind does not re-list them,
+// so a new value does not need a change here. The one value it must know is the
+// tiered scheme, which it cannot represent (there is no tiers field), so it is
+// rejected up front.
+const billingSchemeTiered = "tiered"
```
whoAbhishekSah: `billing/product` already exports `BillingSchemeTiered`. The comment's point about not copying the valid-value lists makes sense, but this one constant could come from that package so the two can't drift.

whoAbhishekSah (commented):
Went through the new kind. A few comments below, nothing blocking.

internal/reconcile/billingproduct_test.go
```diff
@@ -0,0 +1,234 @@
...
+		assert.ErrorContains(t, err, "immutable")
+	})
+
+	t.Run("fails the plan deterministically when an add and an immutable change coexist", func(t *testing.T) {
+		s := newBillingProduct()
+		s.Prices[0].Amount = 200 // an immutable change on the existing "default"
+		s.Prices = append(s.Prices, BillingPriceSpec{Name: "yearly", Amount: 1000, Currency: "usd", Interval: "year"})
+		// map iteration order must not change the outcome: the immutable check
+		// runs over every name before any add is considered, so this always fails.
+		for i := 0; i < 25; i++ {
```
AmanGIT07: Can use `range 25`

internal/reconcile/billingproduct_reconciler.go
rohilsurana: ahh right missed this, earlier this had metadata which could give error. letme remove the error return

internal/reconcile/billingproduct.go
rohilsurana: Fixed in 23d27498. normalizeBillingPrice now blanks metered_aggregate for non-metered prices, so a licensed price no longer false-rejects. Matches the server, which only checks the field for metered prices.

internal/reconcile/billingproduct_reconciler.go
rohilsurana: Fixed in abf6e7f1. Export now drops metered_aggregate for non-metered prices. The diff fills the default back in, so the round trip still plans nothing.

internal/reconcile/billingproduct_reconciler.go
rohilsurana: Done in abf6e7f1. billingProductBody returns just the body now, and the two call sites are simplified.

internal/reconcile/billingproduct_reconciler.go
rohilsurana: Added a note in abf6e7f1 in fetchCurrent, calling out that the every-product-in-the-file rule relies on ListProducts returning all products in one page.

internal/reconcile/billingproduct.go
rohilsurana: Fixed in 23d27498. Renamed to priceName and featureName.

internal/reconcile/billingproduct.go
rohilsurana: Fixed in 23d27498. The const now derives from product.BillingSchemeTiered, so the two cannot drift.

internal/reconcile/billingproduct.go
rohilsurana: Documented in 35df3f19. Added a line to reconcile.mdx: an omitted behavior becomes basic on create (credits when credit_amount is set) and is left as is on update.

internal/reconcile/billingproduct_test.go
rohilsurana: Done in fe3c316d.

Makefile
```diff
@@ -4,7 +4,7 @@ TAG := $(shell git rev-list --tags --max-count=1)
 VERSION := $(shell git describe --tags ${TAG})
 .PHONY: build check fmt lint test test-race vet test-cover-html help install proto admin-app compose-up-dev
 .DEFAULT_GOAL := build
-PROTON_COMMIT := "f90d337079080c48c6d071982baf175e9f76f033"
+PROTON_COMMIT := "91eaffcdc8435ee129f9f93b43ad957c32efee62"
```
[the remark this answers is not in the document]
rohilsurana: The regen already landed via #1824 (merged into this branch): the checked-in proto has no PlanId on ProductRequestBody (verified). Resolved.

internal/reconcile/billingproduct.go
rohilsurana: Update: enum typos are now caught at plan time too. Validate() runs each entry through the proto protovalidate rules, so a bad behavior/interval/usage_type/scheme fails the whole file up front, not only at apply.

whoAbhishekSah (approved):

PR #1822, opened 2026-07-30
refactor(project): move project creation grants into membership

whoAbhishekSah:
## What

Project creation used to write two things directly: the project→org relation and the creator's owner policy. This moves both writes behind a new `membership.OnProjectCreated`, the same way group creation already goes through `membership.OnGroupCreated`.

## Why

Membership grants were still written by hand in the project service. With this change, the membership package handles all of them, and the project service no longer creates policies or relations for access.

### DIFF
--- a/core/membership/service.go
+++ b/core/membership/service.go
@@ -954,6 +954,53 @@ func (s *Service) RemoveProjectMember(ctx context.Context, projectID, principalI
 	return nil
 }
 
+// OnProjectCreated runs right after a project is created. It links the project
+// to its parent org in SpiceDB and gives the creator the project owner role.
+// There is no org-membership check on the creator here: platform superusers
+// can create projects in orgs they are not members of.
+func (s *Service) OnProjectCreated(ctx context.Context, projectID, orgID, creatorID, creatorType string) error {
+	if err := s.linkProjectToOrg(ctx, projectID, orgID); err != nil {
+		return err
+	}
+	if _, err := s.createPolicy(ctx, projectID, schema.ProjectNamespace, creatorID, creatorType, schema.RoleProjectOwner); err != nil {
+		if cleanupErr := s.unlinkProjectFromOrg(ctx, projectID, orgID); cleanupErr != nil {
+			s.log.WarnContext(ctx, "project hierarchy cleanup failed after owner add failure",
+				"project_id", projectID,
+				"org_id", orgID,
+				"error", cleanupErr,
+			)
+		}
+		return err
+	}
+	return nil
+}
+
+// linkProjectToOrg creates the project->org relation in SpiceDB. Org-level
+// project permissions (e.g. project delete via the org) resolve through it.
+func (s *Service) linkProjectToOrg(ctx context.Context, projectID, orgID string) error {
+	if _, err := s.relationService.Create(ctx, relation.Relation{
+		Object:       relation.Object{ID: projectID, Namespace: schema.ProjectNamespace},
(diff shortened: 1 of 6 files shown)


whoAbhishekSah:
Tested live on a local server (fresh org, real ConnectRPC calls, verified against the policies table and SpiceDB tuples):

| # | Actor | Action | Result |
|---|-------|--------|--------|
| 1 | Org owner (member) | CreateProject | Created; creator got `app_project_owner` policy; `project → org` tuple written |
| 2 | Superadmin, not an org member | CreateProject in that org | Created; same owner policy + org tuple; no membership required |
| 3 | Org owner, no direct policy on the superadmin's project | get/update/delete checks | Allowed — inherited through the `project → org` link |
| 4 | Authenticated user with no relation to the org | CreateProject | `permission_denied`; no orphan project row left |
| 5 | Same outsider | Permission checks on projects/resources | All false |
| 6 | Org owner / superadmin | Create custom resource (`resource/aoi`, `resource/workflow`) under the new projects | Created; `owner` + `project` tuples written; CheckResourcePermission matrix correct (owner true, outsider false, inherited org access true) |
| 7 | Org owner / superadmin | DeleteProject | Deleted; zero live project tuples left in SpiceDB |

No regressions found.

core/membership/service.go
```diff
@@ -954,6 +954,53 @@ func (s *Service) RemoveProjectMember(ctx context.Context, projectID, principalI
...
+	}
+	if _, err := s.createPolicy(ctx, projectID, schema.ProjectNamespace, creatorID, creatorType, schema.RoleProjectOwner); err != nil {
+		if cleanupErr := s.unlinkProjectFromOrg(ctx, projectID, orgID); cleanupErr != nil {
+			s.log.WarnContext(ctx, "project hierarchy cleanup failed after owner add failure",
+				"project_id", projectID,
+				"org_id", orgID,
+				"error", cleanupErr,
+			)
+		}
+		return err
```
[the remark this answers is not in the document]
whoAbhishekSah: A warning log is fine for alerting developers.

core/membership/service.go
```diff
@@ -954,6 +954,53 @@ func (s *Service) RemoveProjectMember(ctx context.Context, projectID, principalI
...
+}
+
+// unlinkProjectFromOrg removes the project->org relation. Used to clean up
+// when project creation fails partway.
+func (s *Service) unlinkProjectFromOrg(ctx context.Context, projectID, orgID string) error {
+	if err := s.relationService.Delete(ctx, relation.Relation{
+		Object:       relation.Object{ID: projectID, Namespace: schema.ProjectNamespace},
+		Subject:      relation.Subject{ID: orgID, Namespace: schema.OrganizationNamespace},
+		RelationName: schema.OrganizationRelationName,
+	}); err != nil && !errors.Is(err, relation.ErrNotExist) {
```
AmanGIT07, line 998: will we get `relation.ErrNotExist` on this path?

whoAbhishekSah: Normally no. We only call this right after the link was created, so the row is there.

Even if the row is already gone, `relation.Service.Delete` returns nil, not `ErrNotExist`. It first lists the matching rows, and for a missing row it just gets an empty list and does nothing.

There is one case where the error can show up: the row is found by the list step, but something else deletes it before `DeleteByID` runs. `DeleteByID` returns `ErrNotExist` when its delete removes zero rows. The check is there for that case, so the cleanup never fails on a row that is already gone. `unlinkGroupFromOrg` has the same check for the same reason.

AmanGIT07 (approved):

PR #1823, opened 2026-07-30
refactor(project): remove unused SetMemberRole and RemoveMember

whoAbhishekSah:
## What

Deletes two methods on the project service that are not called anymore: `SetMemberRole` and `RemoveMember`. The API handlers switched to `membership.SetProjectMemberRole` and `membership.RemoveProjectMember` a while back, so these were dead code doing their own policy reads and writes.

Stacked on #1822.


AmanGIT07 (approved):

whoAbhishekSah:
Tested live on a local server (fresh org + project, bob joined via invite → accept, verified against the policies table and SpiceDB tuples):

| # | Case | Call | Result |
|---|------|------|--------|
| 1 | Org member gets a project role | SetProjectMemberRole → `app_project_viewer` | Policy written; checks: `get` true, `update`/`delete` false |
| 2 | Role upsert | SetProjectMemberRole again → `app_project_owner` | Old policy replaced (one row left); all checks flipped to true |
| 3 | Principal not in the org | SetProjectMemberRole | `failed_precondition: principal is not a member of the resource` |
| 4 | Org-scoped role on a project | SetProjectMemberRole | `invalid_argument: role is not valid for project scope` |
| 5 | Remove member | RemoveProjectMember | Policies gone, all checks false, `ListProjectUsers` correct; org membership untouched |
| 6 | Remove someone who was never a member | RemoveProjectMember | `not_found` |
| 7 | Wiring regression (`cmd/serve.go` changed) | CreateProject / DeleteProject | Both work; zero live tuples left after cleanup |

No regressions found.

One note from testing, unrelated to this PR: the `app_project_manager` role in my local data had empty `scopes`, so assigning it fails case-4 style. The predefined role definition needs `scopes: [app/project]` in the deployment's role config for manager assignment to work.

PR #1824, opened 2026-07-30
refactor(billing): remove plan_id from the product create/update RPCs

rohilsurana:
## What

Remove `plan_id` from the product create and update RPCs. `ProductRequestBody` is shared by `CreateProduct` and `UpdateProduct`, and this drops the field from both:

- Regenerates against proton, which reserves `plan_id` (field 4) on `ProductRequestBody`.
- Updates the `CreateProduct` handler to stop reading `plan_id` (it no longer links a plan on create).
- `UpdateProduct` already ignored `plan_id`, so it needs no code change beyond the regen.

## Why

Plan membership is owned by the plan side. `UpsertPlans` creates a product and then links it with `AddPlan`, after checking the product has an active price at the plan's interval. The request-body `plan_id` was a way around that check: creating a product with a `plan_id` wrote the link straight to the database and skipped the matching-price check, so you could link a product to a plan that has no valid price for its interval. Removing it makes plan membership a single-owner concern (the plans) and closes that back door.

Nothing in Frontier set `plan_id` on create: the plan bootstrap and the reconcile flow both create products without it. The internal `Product.PlanIDs` and `AddPlan` are untouched; they remain the real linking path.

## Depends on

- **raystack/proton#495** (reserve `plan_id` on `ProductRequestBody`) is merged. `PROTON_COMMIT` is pinned to its merge commit `91eaffc`.
- Stacked on **#1816** (which owns the frontier proton pin bump), so the regen here is only the `plan_id` removal. Base is `feat/reconcile-billing-product`; rebase onto `main` once #1816 lands.

## Tests

Updated the connect handler tests (dropped the `plan_id` request field and the `plan_id=` fragment from the error messages) and the e2e billing regression test (dropped the empty `PlanId` on each `ProductRequestBody`). Connect, billing, and reconcile suites pass; build, gofmt, vet, and lint are clean.


internal/api/v1beta1connect/billing_product.go
```diff
@@ -82,14 +82,9 @@ func (h *ConnectHandler) CreateProduct(ctx context.Context, request *connect.Req
 			MaxQuantity:  request.Msg.GetBody().GetBehaviorConfig().GetMaxQuantity(),
 		}
 	}
-	// only link a plan when one is given; an empty plan id would persist a junk
-	// [""] entry that later breaks plan lookup by id.
-	var planIDs []string
-	if planID := request.Msg.GetBody().GetPlanId(); planID != "" {
-		planIDs = []string{planID}
-	}
+	// a product is not linked to a plan here; plan membership is managed from the
```
rohilsurana: no need to mention these details here

AmanGIT07 (approved):

internal/api/v1beta1connect/billing_product.go
rohilsurana: Removed in 4720da57.

PR #1827, opened 2026-07-30
refactor(membership): deduplicate repeated patterns

whoAbhishekSah:
## What this PR does

It removes duplicated code inside the `core/membership` package. It builds on #1826, which split the one big service file into smaller files. With the code side by side, the copies were easy to see.

No results change. Every public function keeps its name and signature. The package shrinks by 165 lines. One function makes one less database query (change #9, from review).

## The duplications, one by one

**1. Two functions built the same policy filter.**
`resourcePolicyFilter` (list.go) and `policyFilterForResource` (service.go) did the same job: take a resource ID and type, and set the matching field on `policy.Filter` (`OrgID`, `ProjectID`, or `GroupID`). Now only `resourcePolicyFilter` exists. `removeAllPolicies` uses it, and then calls the existing `removePoliciesByFilter` instead of keeping its own copy of the list-then-delete loop.

One side effect to know about: the filter built by `removeAllPolicies` now also sets `ResourceType`. The query result is the same. The repository already pins the resource type whenever `OrgID`/`ProjectID`/`GroupID` is set (see `applyListFilter` in `internal/store/postgres/policy_repository.go`), so the new condition is redundant, not different.

**2. Three role validators were the same function.**
`validateOrgRole`, `validateProjectRole`, and `validateGroupRole` all did the same three steps: fetch the role, check it is scoped to the right namespace, then accept it if it is a platform-wide role or a custom role of this org. Only the namespace and the returned error differed. The shared steps now live in `validateRoleForScope`. The three old functions stay as one-line wrappers, so call sites read the same as before.

**3. Two "don't remove the last owner" checks were the same function.**
`validateMinOwnerConstraint` (org) and `validateMinGroupOwnerConstraint` (group) ran the same algorithm: fetch the owner role, skip the check if the member is being promoted to owner or isn't an owner, otherwise count the owner policies and fail if this member is the only one. Both now call the shared `validateMinRoleConstraint`. Each wrapper passes its own owner role name, filter, and error.

**4. Five audit helpers repeated the same ~30 lines.**
The org and group helpers for member added, role changed, and member removed each built the same audit record (target, metadata with `role_id` and `email`) and then wrote the same legacy audit log. That shared body is now one function, `auditMemberChange`. The five helpers became short wrappers that only supply the event names and attributes. Adding the next membership event is a five-line wrapper.

**5. A repeated condition got a name.**
`len(existing) == 1 && existing[0].RoleID == roleID` appeared at four call sites. It now reads `hasExactlyRole(existing, roleID)`, which says what the check means: the member already has exactly this role, so there is nothing to change.

**6. String-glued map keys became a struct.**
Two places built map keys by joining strings: `principalType + "\x00" + principalID`. They now use a small struct, `principalKey{Type, ID}`. Go structs work directly as map keys, and there is no separator character to reason about.

**7. The longest function got split.**
`ListPrincipalsByResource` was ~85 lines doing two jobs: find the members, then fetch every role each member holds. The second job moved into its own function, `enrichMemberRoles`.

**8. Hand-written loops replaced by an existing helper.**
`listOrgsForPrincipal` and `listGroupsForPrincipal` each looped over policies to collect resource IDs and dedupe them. The helper `policyResourceIDs` already did exactly that, so they call it now.

**9. One less query in `ListPrincipalsByResource` (from review).**
CodeRabbit spotted a pre-existing inefficiency that the split made visible: when the caller passes no role filter, the role-enrichment query was identical to the member query, so the function hit the database twice for the same rows. Now the first result is reused; only a role-filtered listing issues the second, unfiltered query. Same results, one query instead of two on the common path.

## What I looked at but did not merge

`validateOrgMembership` and `validatePrincipal` overlap, but they return different errors (`ErrNotOrgMember` vs `ErrPrincipalNotInOrg`), and only one of them checks whether a service user is disabled. Merging them would change behavior, and this PR promises not to. Left as is.

## Tests

Two kinds of test changes, both matching the changes above:

- The four mock expectations in `TestService_RemoveProjectMember` now expect `ResourceType: schema.ProjectNamespace` in the filter (change #1).
- The four no-role-filter cases in `TestService_ListPrincipalsByResource` now pin the policy `List` call with `.Once()` instead of `.Times(2)` (change #9), proving the duplicate query is gone.

Everything else passes untouched. `go test ./core/... ./internal/api/...` is green.

### DIFF
--- a/core/membership/group.go
+++ b/core/membership/group.go
@@ -174,12 +174,12 @@ func (s *Service) RemoveAllGroupMembers(ctx context.Context, groupID string) err
 	// First pass: delete every policy. Track which principals had any
 	// delete failure so we don't strip their SpiceDB relations while a
 	// surviving policy still references them.
-	principals := make(map[principalKey]policy.Policy, len(policies))
+	principals := make(map[principalKey]struct{}, len(policies))
 	failed := make(map[principalKey]struct{}, len(policies))
 	var errs error
 	for _, p := range policies {
 		key := policyPrincipalKey(p)
-		principals[key] = p
+		principals[key] = struct{}{}
 		if delErr := s.policyService.Delete(ctx, p.ID); delErr != nil {
 			failed[key] = struct{}{}
 			errs = errors.Join(errs, fmt.Errorf("delete policy %s: %w", p.ID, delErr))
@@ -189,12 +189,12 @@ func (s *Service) RemoveAllGroupMembers(ctx context.Context, groupID string) err
 	// Second pass: clean up direct relations only for principals whose
 	// policies were all deleted successfully. The rest get retried on the
 	// next attempt once their lingering policies are removed.
-	for key, p := range principals {
+	for key := range principals {
 		if _, hadFailure := failed[key]; hadFailure {
 			continue
 		}
-		if relErr := s.removeGroupMemberRelation(ctx, groupID, p.PrincipalID, p.PrincipalType); relErr != nil {
-			errs = errors.Join(errs, fmt.Errorf("remove relations for %s:%s: %w", p.PrincipalType, p.PrincipalID, relErr))
+		if relErr := s.removeGroupMemberRelation(ctx, groupID, key.ID, key.Type); relErr != nil {
+			errs = errors.Join(errs, fmt.Errorf("remove relations for %s:%s: %w", key.Type, key.ID, relErr))
 		}
 	}
 
--- a/core/membership/group.go
+++ b/core/membership/group.go
@@ -4,13 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"slices"
 
 	"github.com/raystack/frontier/core/policy"
 	"github.com/raystack/frontier/core/relation"
 	"github.com/raystack/frontier/core/role"
 	"github.com/raystack/frontier/internal/bootstrap/schema"
-	"github.com/raystack/frontier/pkg/utils"
 )
 
 // removeGroupMemberRelation deletes the member relation for a principal on a group.
@@ -83,7 +81,7 @@ func (s *Service) SetGroupMemberRole(ctx context.Context, groupID, principalID, 
 	}
 
 	// change path: skip if the principal already has exactly this role
-	if len(existing) == 1 && existing[0].RoleID == resolvedRoleID {
+	if hasExactlyRole(existing, resolvedRoleID) {
 		return nil
 	}
 
@@ -176,11 +174,11 @@ func (s *Service) RemoveAllGroupMembers(ctx context.Context, groupID string) err
 	// First pass: delete every policy. Track which principals had any
 	// delete failure so we don't strip their SpiceDB relations while a
 	// surviving policy still references them.
-	principals := make(map[string]policy.Policy, len(policies))
-	failed := make(map[string]struct{}, len(policies))
+	principals := make(map[principalKey]policy.Policy, len(policies))
+	failed := make(map[principalKey]struct{}, len(policies))
 	var errs error
 	for _, p := range policies {
-		key := p.PrincipalType + "\x00" + p.PrincipalID
+		key := policyPrincipalKey(p)
 		principals[key] = p
 		if delErr := s.policyService.Delete(ctx, p.ID); delErr != nil {
 			failed[key] = struct{}{}
@@ -311,20 +309,7 @@ func (s *Service) unlinkGroupFromOrg(ctx context.Context, groupID, orgID string)
 //   - a platform-wide role scoped to groups, or
 //   - a custom role created for the group's parent organization.
 func (s *Service) validateGroupRole(ctx context.Context, roleID, orgID string) (role.Role, error) {
-	fetchedRole, err := s.roleService.Get(ctx, roleID)
-	if err != nil {
-		return role.Role{}, err
-	}
-	if !slices.Contains(fetchedRole.Scopes, schema.GroupNamespace) {
-		return role.Role{}, ErrInvalidGroupRole
-	}
-	if fetchedRole.OrgID == orgID {
-		return fetchedRole, nil
-	}
-	if utils.IsNullUUID(fetchedRole.OrgID) {
-		return fetchedRole, nil
-	}
-	return role.Role{}, ErrInvalidGroupRole
+	return s.validateRoleForScope(ctx, roleID, orgID, schema.GroupNamespace, ErrInvalidGroupRole)
 }
 
 // validateMinGroupOwnerConstraint ensures the group keeps at least one owner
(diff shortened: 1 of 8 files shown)


whoAbhishekSah:
Applied the CodeRabbit suggestion in c3464d72: `ListPrincipalsByResource` now reuses the first policy query for role enrichment when no role filter is set; only a role-filtered listing issues the second, unfiltered query. The four no-role-filter test cases pin the single call with `.Once()`.

whoAbhishekSah:
Tested every code path this PR touches, on a live server. No behavior changes found.

**Setup**
- Built this branch and ran it on a local sandbox (port 8003), pointed at the same local Postgres and SpiceDB as a second server running the code before this PR — so every read could be compared between the two.
- Fresh org with 4 new users: alice (creator/owner), bob, charlie, dave. A project and two groups under the org.
- A second org owned by dave, with a custom role in it, for the cross-org role test.
- Admin RPCs used the bootstrap service account. User RPCs used normal login sessions.
- Audit checks read the `audit_records` and `auditlogs` tables directly.
- For the review follow-up (duplicate query fix), counted the actual `policies` queries per request using Postgres statement logging, comparing the commit before the fix against the fix.

| # | Area | Test | Expected | Result |
|---|------|------|----------|--------|
| 1 | Org members | Add bob with the org viewer role | Added, audit written | ✅ |
| 2 | Org members | Add with a project-scoped role | Rejected: "role is not valid for organization scope" | ✅ |
| 3 | Org members | Add with another org's custom role | Rejected, same error | ✅ |
| 4 | Org members | Set the same role again | Nothing changes, no duplicate role | ✅ |
| 5 | Org members | Change bob viewer → manager | Role replaced, audit written | ✅ |
| 6 | Org members | Demote an owner while another owner exists | Allowed | ✅ |
| 7 | Org members | Demote the only remaining owner | Blocked: "org must have at least one owner..." | ✅ |
| 8 | Org lists | ListOrganizationUsers, no filter | All members with their full role lists, same as before | ✅ |
| 9 | Org lists | ListOrganizationUsers, filtered by role | Only matching members, but each still shows all their roles | ✅ |
| 10 | Org lists | ListOrganizationAdmins | Returns the owner | ✅ |
| 11 | Org lists | ListOrganizationsByCurrentUser as bob | Shows his org, same as before | ✅ |
| 12 | Org lists | ListOrganizationsByUser | As a normal user: blocked (superuser-only, by design). As superuser: returns the orgs | ✅ |
| 13 | Project members | Add bob as project viewer | Added | ✅ |
| 14 | Project members | Set the same role again | Nothing changes | ✅ |
| 15 | Project members | Add with an org-scoped role | Rejected: "role is not valid for project scope" | ✅ |
| 16 | Project members | ListProjectUsers with bob holding 2 roles | Bob listed once, both roles shown | ✅ |
| 17 | Project members | Remove bob, who had 2 policies on the project | Both policies deleted (0 rows left in DB); his org membership untouched | ✅ |
| 18 | Group members | Add bob as group member | Added, audit written | ✅ |
| 19 | Group members | Set the same role again | Nothing changes | ✅ |
| 20 | Group members | Add with a project-scoped role | Rejected: "role is not valid for group scope" | ✅ |
| 21 | Group members | Promote bob member → group owner | Works, audit written | ✅ |
| 22 | Group members | Demote a group owner while another owner exists | Allowed | ✅ |
| 23 | Group members | Demote the last group owner | Blocked: "group must have at least one owner..." | ✅ |
| 24 | Group members | Remove a member | Removed; audit written with email and no role_id (same as before) | ✅ |
| 25 | Group lists | ListGroupUsers / ListCurrentUserGroups | Correct members and groups | ✅ |
| 26 | Group delete | Delete a group with 3 members | Every member policy on the group deleted (0 rows left) | ✅ |
| 27 | Audit | All added / role_changed / removed events in both audit tables | Same events, email, role_id and group_id fields as before | ✅ |
| 28 | Review fix | Policy queries per list call, no role filter | 2 queries before the fix → 1 after; results identical | ✅ |
| 29 | Review fix | Policy queries per list call, with a role filter | 2 queries in both versions (the second one is needed for full role lists) | ✅ |

Unit tests pass (`go test ./core/membership/...`), and the mocks now pin the list call to run once on the unfiltered path, so a regression back to double-querying would fail CI.

core/membership/service.go
```diff
@@ -210,34 +212,102 @@ func (s *Service) createRelation(ctx context.Context, resourceID, resourceType,
...
+
+// validateMinRoleConstraint ensures at least one holder of guardRoleName
+// remains on the resource after demoting or removing one. resourceFilter
+// scopes the holder count to the resource (the role ID is filled in here);
+// errLast is returned when the principal is the last holder. Returns the
+// resolved guard role ID for reuse as an atomic delete guard.
+func (s *Service) validateMinRoleConstraint(ctx context.Context, guardRoleName string, resourceFilter policy.Filter, newRoleID string, existing []policy.Policy, errLast error) (string, error) {
+	guardRole, err := s.roleService.Get(ctx, guardRoleName)
+	if err != nil {
+		return "", fmt.Errorf("get owner role: %w", err)
```
AmanGIT07: nit: the old group path said "get group owner role", so org and group failures were distinguishable in logs. The helper already has `guardRoleName`, so we can keep that distinction without the shared code knowing about org or group:          
```
return "", fmt.Errorf("get role %s: %w", guardRoleName, err)
```

core/membership/service.go
```diff
@@ -210,34 +212,102 @@ func (s *Service) createRelation(ctx context.Context, resourceID, resourceType,
...
+	if !holdsGuardRole {
+		return guardRole.ID, nil
+	}
+
+	// principal holds the guarded role and is losing it — make sure at least
+	// one other holder remains
+	resourceFilter.RoleID = guardRole.ID
+	holderPolicies, err := s.policyService.List(ctx, resourceFilter)
+	if err != nil {
+		return "", fmt.Errorf("list owner policies: %w", err)
```
AmanGIT07: Same as above, keeps the two owner lookups distinguishable
```
return "", fmt.Errorf("list policies for role %s: %w", guardRoleName, err)
```

core/membership/group.go
```diff
@@ -176,11 +174,11 @@ func (s *Service) RemoveAllGroupMembers(ctx context.Context, groupID string) err
 	// First pass: delete every policy. Track which principals had any
 	// delete failure so we don't strip their SpiceDB relations while a
 	// surviving policy still references them.
-	principals := make(map[string]policy.Policy, len(policies))
-	failed := make(map[string]struct{}, len(policies))
+	principals := make(map[principalKey]policy.Policy, len(policies))
```
AmanGIT07: nit: now that the key is a struct, `key.ID` and `key.Type` are the same strings as `p.PrincipalID` and `p.PrincipalType`, and the second pass reads nothing else from the value — so the stored `policy.Policy` is no longer needed. It's also a small trap: when a principal has several policies, the value is whichever policy was written last, so only the two identity fields are safe to read from it anyway.                                      
                                                                                                                                                                                                                                         
  This could be a set, like `failed` on the next line:                                                                                                                                                                                   
                                                                                                                                                                                                                                         
  ```go                                                                                                                                                                                                                                  
  principals := make(map[principalKey]struct{}, len(policies))                                                                                                                                                                           
  ...                                                                                                                                                                                                                                    
  principals[key] = struct{}{}                                                                                                                                                                                                           
  ...                                                                                                                                                                                                                                    
  for key := range principals {                                                                                                                                                                                                          
        ...                                                                                                                                                                                                                              
        s.removeGroupMemberRelation(ctx, groupID, key.ID, key.Type)                                                                                                                                                                      
  }                                                                                                                                                                                                                                      
  ```

AmanGIT07 (approved):

core/membership/service.go
whoAbhishekSah: Done in 423e594e — `get role %s: %w` with the guard role name.

core/membership/service.go
whoAbhishekSah: Done in 423e594e — `list policies for role %s: %w`.

core/membership/group.go
whoAbhishekSah: Done in 423e594e. principals is a set now and the relation pass reads key.ID/key.Type — good catch on the last-write value being arbitrary.

PR #1834, opened 2026-08-03
fix(billing): include plans without products in plan listings

### LINKED ISSUE #1832: bug(billing): ListAllPlans and ListPlans omit plans that have no products
## What

`BillingPlanRepository.ListWithProducts` (`internal/store/postgres/billing_plan_repository.go`, the join near line 357) INNER JOINs `billing_products` on `plan.id = ANY(product.plan_ids)`. A plan with no products never appears in the result. `UpsertPlans` allows creating a plan with an empty product set, so such plans can exist.

This affects both `FrontierService.ListPlans` and the new `AdminService.ListAllPlans`. The point of `ListAllPlans` (and the future BillingPlan reconcile export) is to surface every plan, including misconfigured or product-less ones, so this is a correctness gap for that endpoint.

## Fix

Change the INNER JOIN to a LEFT JOIN and tolerate NULL product columns:

- `PlanProductRow`'s product fields are non-nullable today (`string`, `time.Time`, `BehaviorConfig`, `pq.StringArray`), so scanning NULLs from a LEFT JOIN fails. The fields need to become nullable, or the SELECT needs `COALESCE`, and the row loop must skip appending an empty product.
- This is a shared read path used by `ListPlans` too, so its behavior changes as well (product-less active plans start appearing).
- Ship it with a Postgres/integration test for a product-less plan. No seed today creates one.

## Context

Deferred from the #1830 review to keep that PR focused and low-risk. The rework touches shared read code and cannot be unit-tested without Postgres, and the BillingPlan reconciler that needs this does not exist yet.

rohilsurana:
## What

`BillingPlanRepository.ListWithProducts` INNER JOINed `billing_products` on `plan.id = ANY(product.plan_ids)`, so any plan with no products was dropped from the result. `UpsertPlans` allows creating a plan with an empty product set, so such plans can exist. This switches the join to a LEFT JOIN so those plans are returned.

- INNER JOIN -> LEFT JOIN in `ListWithProducts`.
- The product columns on `PlanProductRow` become pointers, because a left join leaves them null for a plan with no products. `getProduct` dereferences them.
- The row loop skips appending a product when `product_id` is null, and always stores the accumulated plan back into the map (this also fixes a latent store-back bug where a multi-product plan could lose products after the first).
- `plan_deleted_at` was selected from the product table by mistake, so a plan could report its product's deletion time as its own. It now comes from the plans table.

## Why

This affects both `FrontierService.ListPlans` and, once raystack/frontier#1830 lands, `AdminService.ListAllPlans`. An admin listing (and the future BillingPlan reconcile export) must be able to see every plan, including a misconfigured one with no products, so it can be reported or fixed.

## Testing

- `go build ./...`, `go vet`, and `golangci-lint` are clean.
- The SQL change needs a live Postgres, so it is not exercised by unit tests. A repository test that creates a product-less plan and asserts it is listed should follow (the harness exists in `billing_product_repository_test.go`).

Closes #1832.

### DIFF
--- a/internal/store/postgres/billing_plan_repository.go
+++ b/internal/store/postgres/billing_plan_repository.go
@@ -54,20 +54,22 @@ type PlanProductRow struct {
 	PlanUpdatedAt time.Time  `db:"plan_updated_at"`
 	PlanDeletedAt *time.Time `db:"plan_deleted_at"`
 
-	ProductID          string         `db:"product_id"`
-	ProductProviderID  string         `db:"product_provider_id"`
+	// product columns are pointers because a left join leaves them null for a
+	// plan that has no products
+	ProductID          *string        `db:"product_id"`
+	ProductProviderID  *string        `db:"product_provider_id"`
 	ProductPlanIDs     pq.StringArray `db:"product_plan_ids"`
-	ProductName        string         `db:"product_name"`
+	ProductName        *string        `db:"product_name"`
 	ProductTitle       *string        `db:"product_title"`
 	ProductDescription *string        `db:"product_description"`
 
-	ProductBehavior string             `db:"product_behavior"`
+	ProductBehavior *string            `db:"product_behavior"`
 	ProductConfig   BehaviorConfig     `db:"product_config"`
-	ProductState    string             `db:"product_state"`
+	ProductState    *string            `db:"product_state"`
 	ProductMetadata types.NullJSONText `db:"product_metadata"`
 
-	ProductCreatedAt time.Time  `db:"product_created_at"`
-	ProductUpdatedAt time.Time  `db:"product_updated_at"`
+	ProductCreatedAt *time.Time `db:"product_created_at"`
+	ProductUpdatedAt *time.Time `db:"product_updated_at"`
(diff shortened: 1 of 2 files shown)


internal/store/postgres/billing_plan_repository.go
```diff
@@ -91,20 +93,34 @@ func (pr PlanProductRow) getPlan() (plan.Plan, error) {
...
+	return *s
+}
+
+func derefTime(t *time.Time) time.Time {
+	if t == nil {
+		return time.Time{}
+	}
+	return *t
+}
+
```
AmanGIT07: Please check if we can use `internal/store/postgres/null_converters.go` or go 1.26 semantics here

rohilsurana: Switched to the existing ptrToString and added ptrToTime next to it in null_converters.go (81de9441). Go 1.26 new(expr) does not apply here: it takes a value and returns a pointer to it, while these fields need the opposite, a possibly-nil pointer turned back into a value, which still needs the nil check.

AmanGIT07 (approved):

Issue #1837, opened 2026-08-04
Organization delete: check all blockers up front and tell the user what to fix


whoAbhishekSah:
## The problem

Before deleting an org, the only check we run today is: does the billing account have any invoice, ever? If yes, we refuse — and because of the error mapping issue (#1836), the caller sees a plain `internal` error with no explanation.

This check is wrong in both directions:

- **Too strict**: even a fully paid, closed invoice blocks deletion forever. The user has no way to remove invoice history, so the org becomes undeletable and the error cannot guide them anywhere.
- **Too loose**: an active subscription does not block anything. The delete flow silently cancels it on Stripe mid-delete. Token balances are not looked at either — unused paid tokens vanish, and unpaid token debt is written off.

## The fix

Run one pre-flight check before deleting anything, collect **all** blockers, and return them together as a `failed_precondition` error. Each blocker names the fix, and each fix is something the caller can do themselves through the API:

| Blocker | What the user does |
|---|---|
| Active or trialing subscription | cancel it (`CancelSubscription`), then retry |
| Open or unpaid invoice | pay it via the invoice's hosted URL, then retry |
| Postpaid account with negative token balance | buy tokens to clear the debt, then retry |
| Prepaid account with unused tokens | re-send the delete with the acknowledgment flag (see below) |

Details of the behavior:

- **Paid invoices stop blocking.** Stripe keeps its own permanent copy of every invoice, so removing our rows loses nothing. Only open/unpaid invoices block.
- **All blockers come back in one response**, using standard `PreconditionFailure` error details (one violation per blocker, with a machine-readable type like `ACTIVE_SUBSCRIPTION` and a human message).
- The check runs **before any deletion starts**, so a blocked delete changes nothing (this depends on the ordering work in #1835).
- The handler maps this error to `failed_precondition`; only unexpected failures stay `internal`.

## Proto change: acknowledging unused tokens

Unused prepaid tokens are the one blocker with no clean-up action — the user already paid for them, and there is nothing to cancel or pay. What we need from the user is informed consent: "I understand deleting this org forfeits my remaining tokens."

A confirmation popup in a UI is not enough on its own. The API is public — SDKs, scripts, curl — and the server cannot see a popup. For the server to enforce the consent, it has to arrive inside the request. So:

- Add a field to `DeleteOrganizationRequest` in the proto definitions (raystack/proton), for example `bool acknowledge_token_forfeit`.
[...]

PR #1838, opened 2026-08-04
fix(deleter): clean up all billing and kyc records during org delete

### LINKED ISSUE #1835: Organization delete can fail halfway and leave the org in a broken state
## The problem

When we delete an organization, `deleter.Service.DeleteOrganization` removes things in this order:

1. Policies (this includes the org owners)
2. Projects, groups, service users, roles, invitations
3. Billing data: subscriptions, invoices, the Stripe customer, then the `billing_customers` row
4. The org itself

Step 3 can fail because of two tables we never clean up:

- `billing_checkouts` has a foreign key to `billing_customers`. We never delete checkout rows — the checkout repository does not even have a delete method. So if the org ever opened a Stripe checkout page (almost every paying org has), the hard delete of the customer row fails with a foreign key error.
- `org_kyc` has a foreign key to `organizations`. We never delete the KYC row, so the org row delete can fail the same way.

When this happens, the damage is already done. The owners and projects were deleted in steps 1 and 2, and the Stripe customer was deleted just before the failing row. The org is now half-deleted: it still exists, but nobody owns it, and its billing record points to a Stripe customer that no longer exists. Retrying does not help — the same foreign key fails every time, and every billing API call for the org keeps failing with `No such customer` from Stripe.

Two smaller gaps in the same flow:

- For billing accounts not linked to Stripe (offline accounts), we skip deleting their local subscription and invoice rows. If such rows exist, the customer delete fails the same way.
- Token transactions (`billing_transactions`) have no foreign key, so they are silently left behind after a delete.

## The fix

1. Add a delete method for checkouts and call it in `DeleteCustomers` before deleting the billing customer row.
2. Delete the `org_kyc` row as part of the org delete.
3. Always delete local subscription and invoice rows, whether or not the account is on Stripe.
4. Delete the org's token transactions along with the billing customer.
5. Reorder the flow: billing teardown first, identity (policies, projects, and so on) after, org row last. Billing talks to an external system, so it is the step most likely to fail. If it fails first, the org still has its owners and a retry can finish the job cleanly.
6. Make every step safe to retry. If Stripe says a customer or subscription is already gone, treat that as success. The customer delete already does this; the subscription path should too.

## Why this way

Deleting an org touches many tables and an external system, and we cannot wrap Stripe calls in a database transaction. So the realistic goal is not "all or nothing" — it is "any failure leaves the org retryable, never broken". Cleaning up every referencing table removes the foreign key failures, and the reordering plus retry-safety means a failed delete can always be run again to completion.

## Note: these are hard deletes

All the deletes above remove rows permanently. We considered soft delete (keeping rows and marking them with `deleted_at`) and decided against it:

- Some tables have a `deleted_at` column, but no query in the codebase filters on it. Soft delete would require updating every Get/List query in every affected repository, and missing one means deleted orgs keep showing up in lists and sync jobs.
- Names are globally unique (`organizations.name`, `projects.name`). A soft-deleted org would hold its name forever, so nobody could reuse it without new partial unique indexes.
- The foreign key chain makes it all-or-nothing: `billing_customers.org_id` points to `organizations.id`, so we cannot keep billing rows while hard-deleting the org. Either the whole chain goes soft or the whole chain goes hard.
- Frontier already has a reversible option: disabling an org keeps everything and can be undone. Delete is the permanent purge, and it should behave like one.

If a deleted record is ever needed later, there are still two places to find it:

1. **Audit records** — the delete flow writes an audit record for each deleted entity (who, what, when, and key fields).
2. **Stripe** — Stripe permanently keeps its own copy of every customer, subscription, invoice, and checkout, even after deletion.

whoAbhishekSah:
Fixes #1835. Top layer of the cleanup stack — uses the delete methods added in #1841, #1842, #1843 and relies on the retryable teardown from #1844.

## Problem

Organization delete could fail halfway and leave the org in a broken state: no owners, and a billing account pointing to a Stripe customer that was already deleted. Checkout and kyc rows were never removed, so their foreign keys made the hard delete of `billing_customers` and `organizations` fail after the Stripe customer was gone. Retrying hit the same error forever.

## Changes

- The deleter now removes checkout rows and credit transactions before the billing customer row, and the `org_kyc` row before the org row.
- Reordered the teardown: billing first, org policies near the end. A failure at any step now leaves the org owned and the delete retryable.
- Always remove local subscription and invoice rows, including for offline billing accounts.

## Testing

- Rewrote `core/deleter/service_test.go` to cover the new deletion order and the checkout, credit, and kyc steps.


core/deleter/service.go
```diff
@@ -289,15 +320,20 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
...
-				return fmt.Errorf("failed to delete org while deleting a billing account invoices[%s]: %w", c.ID, err)
-			}
+		// cancels active subscriptions on the billing provider and removes local records
+		if err := d.subService.DeleteByCustomer(ctx, c); err != nil {
+			return fmt.Errorf("failed to delete org while deleting a billing account subscriptions[%s]: %w", c.ID, err)
+		}
+		if err := d.invoiceService.DeleteByCustomer(ctx, c); err != nil {
+			return fmt.Errorf("failed to delete org while deleting a billing account invoices[%s]: %w", c.ID, err)
+		}
+		// checkouts and credit transactions reference the billing account row,
```
whoAbhishekSah: Redundant comment. Can be removed.

core/deleter/service_test.go
```diff
@@ -22,306 +22,343 @@ import (
...
+		// roles
+		m.roleSvc.EXPECT().List(mock.Anything, role.Filter{OrgID: "org-1"}).
+			Return([]role.Role{{ID: "role-1", Name: "r1"}}, nil)
+		m.roleSvc.EXPECT().Delete(mock.Anything, "role-1").Return(nil)
 
-		// finally delete org model
-		orgSvc.EXPECT().DeleteModel(mock.Anything, "org-1").Return(nil)
+		// kyc, then the org model
+		m.kycSvc.EXPECT().DeleteKyc(mock.Anything, "org-1").Return(nil)
+		m.orgSvc.EXPECT().DeleteModel(mock.Anything, "org-1").Return(nil)
```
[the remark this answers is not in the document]
whoAbhishekSah: Fixed in 9505aacd: the kyc delete now runs before the org policy loop, and a new test ("kyc delete failure keeps owner policies") asserts via strict mocks that a kyc failure triggers no policy, role, or org-model deletion.

core/deleter/service.go
whoAbhishekSah: Removed in 9505aacd.

core/deleter/service.go
```diff
@@ -242,33 +259,47 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
...
+			return fmt.Errorf("failed to delete org while deleting a role[%s]: %w", p.Name, err)
 		}
 	}
 
-	// delete all billing accounts
-	if err := d.DeleteCustomers(ctx, id); err != nil {
-		return err
+	// delete the org's kyc record; its row references the org row
+	if err := d.kycService.DeleteKyc(ctx, id); err != nil {
+		return fmt.Errorf("failed to delete org while deleting its kyc record: %w", err)
```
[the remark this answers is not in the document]
whoAbhishekSah: Partly addressed by 9505aacd, which moves the kyc delete ahead of the policy loop. Full atomicity for the tail is not possible here: policies and roles live in both Postgres and SpiceDB, and the two stores cannot share a transaction. After the reorder, everything past the policy loop (roles, org row) is local-only work with no provider calls, so a failure there is rare and the delete stays retryable by a superuser — which is the recovery path this PR is designed around. Restoring owner policies on failure would need a compensation log across both stores, which is out of scope for this fix; #1837 (pre-flight blocker checks) further shrinks the window for late failures.

core/deleter/service.go
```diff
@@ -289,15 +321,18 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
...
-			// delete all subscription first
-			if err := d.subService.DeleteByCustomer(ctx, c); err != nil {
-				return fmt.Errorf("failed to delete org while deleting a billing account subscriptions[%s]: %w", c.ID, err)
-			}
-			// delete all invoices
-			if err := d.invoiceService.DeleteByCustomer(ctx, c); err != nil {
-				return fmt.Errorf("failed to delete org while deleting a billing account invoices[%s]: %w", c.ID, err)
-			}
+		// cancels active subscriptions on the billing provider and removes local records
+		if err := d.subService.DeleteByCustomer(ctx, c); err != nil {
```
rohilsurana, line 325: This now runs for offline accounts too, which fixes the case where their local subscription rows blocked the customer hard delete. It depends on the offline skip added in #1844, so make sure the stack lands in order and this never merges ahead of it.

core/deleter/service.go
```diff
@@ -129,7 +144,8 @@ func NewCascadeDeleter(orgService OrganizationService, projService ProjectServic
 	userPATService UserPATService,
 	serviceUserService ServiceUserService,
 	customerService CustomerService, subService SubscriptionService,
-	invoiceService InvoiceService) *Service {
+	invoiceService InvoiceService, checkoutService CheckoutService,
+	creditService CreditService, kycService KycService) *Service {
```
rohilsurana, line 148: Not for this PR, but `NewCascadeDeleter` is now at 17 positional dependencies. A config struct would make the next addition safer than another positional append. Fine as a follow-up.

rohilsurana (commented):
Verified the new order against the schema FKs: subscriptions, invoices, checkouts, and transactions all go before the customer hard delete, and `org_kyc` goes before the owner policies. Every earlier step is idempotent on retry, so a failure at any point leaves the org owned and the delete can just run again. The test rewrite with strict mocks pins exactly that ordering, nice work. Two notes inline.

rohilsurana (approved):

core/deleter/service.go
whoAbhishekSah: The stack enforces this mechanically: this PR's base branch is #1844's branch, so GitHub will not let it merge first. When #1844 merges, this one gets retargeted automatically.

core/deleter/service.go
whoAbhishekSah: Agreed, will do it as a follow-up so this stack stays reviewable. Tracking in #1848.

PR #1839, opened 2026-08-04
feat(deleter): write an audit record for each checkout removed during org delete

whoAbhishekSah:
Stacked on #1838. Review only the last commit here; the rest is the base PR.

## Problem

#1838 hard-deletes the `billing_checkouts` rows during org delete. Those rows hold the only local reference to the Stripe checkout sessions (`provider_id`). Once they are gone, there is no local trace of what was removed.

## Changes

- New audit event `app.billing.checkout.deleted`, registered as a system event.
- During `DeleteCustomers`, the deleter lists the checkouts before removing them and writes one audit record per checkout. Each record carries the provider session id, customer id, plan/product id, state, and payment status.
- Audit writes are best effort: a failed write logs a warning and does not block the delete, same as the existing `organization.deleted` event.
- The audits table is not part of the delete cascade, so these records survive the org delete. The Stripe sessions can still be found later from `provider_id`.

## Testing

- Extended `core/deleter/service_test.go`: the delete flow now expects the checkout list call, and a new case checks that a list failure stops the customer delete.

### DIFF
--- a/core/deleter/service.go
+++ b/core/deleter/service.go
@@ -339,20 +339,28 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 			return fmt.Errorf("failed to delete org while deleting a billing account checkouts[%s]: %w", c.ID, err)
 		}
 		// the checkout rows are gone after this; the audit records keep the
-		// provider references so the sessions can still be found on the provider
+		// provider references so the sessions can still be found on the provider.
+		// The records are only written when ctx carries the audit service (the
+		// API path seeds it); otherwise audit.NewLogger falls back to a noop
 		auditLogger := audit.NewLogger(ctx, id)
 		for _, ch := range checkouts {
-			if err := auditLogger.LogWithAttrs(audit.BillingCheckoutDeletedEvent, audit.Target{
-				ID:   ch.ID,
-				Type: "billing_checkout",
-			}, map[string]string{
+			attrs := map[string]string{
 				"provider_id":    ch.ProviderID,
 				"customer_id":    ch.CustomerID,
 				"plan_id":        ch.PlanID,
 				"product_id":     ch.ProductID,
 				"state":          ch.State,
 				"payment_status": ch.PaymentStatus,
-			}); err != nil {
+			}
+			for k, v := range attrs {
+				if v == "" {
+					delete(attrs, k)
+				}
+			}
+			if err := auditLogger.LogWithAttrs(audit.BillingCheckoutDeletedEvent, audit.Target{
+				ID:   ch.ID,
+				Type: "billing_checkout",
+			}, attrs); err != nil {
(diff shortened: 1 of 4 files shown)


rohilsurana (commented):
The shape is right: list before delete, log after the delete succeeds, and a retry lists zero checkouts so there are no duplicate records. Two notes inline. One minor thought: as more steps get audited, a small helper for this loop would keep DeleteCustomers readable.

core/deleter/service.go
```diff
@@ -328,9 +331,31 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
...
+		checkouts, err := d.checkoutService.List(ctx, checkout.Filter{CustomerID: c.ID})
+		if err != nil {
+			return fmt.Errorf("failed to delete org while listing billing account checkouts[%s]: %w", c.ID, err)
+		}
 		if err := d.checkoutService.DeleteByCustomer(ctx, c.ID); err != nil {
 			return fmt.Errorf("failed to delete org while deleting a billing account checkouts[%s]: %w", c.ID, err)
 		}
+		// the checkout rows are gone after this; the audit records keep the
+		// provider references so the sessions can still be found on the provider
+		auditLogger := audit.NewLogger(ctx, id)
```
rohilsurana, line 345: `audit.GetService` falls back to a noop repository when the context has no audit service, and it only prints a line when that happens. So these records only survive when org delete runs through the API path that seeds the context. If a background or admin job ever calls the deleter, the trail silently disappears. Worth a short comment here, since the whole point of this PR is that the records outlive the rows.

core/deleter/service.go
```diff
@@ -328,9 +331,31 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
...
+		for _, ch := range checkouts {
+			if err := auditLogger.LogWithAttrs(audit.BillingCheckoutDeletedEvent, audit.Target{
+				ID:   ch.ID,
+				Type: "billing_checkout",
+			}, map[string]string{
+				"provider_id":    ch.ProviderID,
+				"customer_id":    ch.CustomerID,
+				"plan_id":        ch.PlanID,
```
rohilsurana, line 350: Most checkouts will have an empty `plan_id` or `product_id`. Skipping empty values would keep the records tidy.

rohilsurana (approved):

core/deleter/service.go
whoAbhishekSah: Added a comment in 9d514916 noting that the records are only written when ctx carries the audit service (the API path seeds it) and that the logger falls back to a noop otherwise. Today the deleter is only reached through the handlers, so the trail is intact for every current caller.

core/deleter/service.go
whoAbhishekSah: Done in 9d514916 — empty values are dropped from the attrs before logging.

PR #1841, opened 2026-08-04
feat(checkout): add delete of checkout records by billing account

whoAbhishekSah:
Part of #1835. Bottom of the org-delete cleanup stack.

Checkout rows hold a foreign key to the billing customer row, and nothing could remove them. That FK is what blocks the hard delete of a billing account during org delete.

## Changes

- `DeleteByCustomerID` on the checkout Postgres repository.
- `DeleteByCustomer` on the checkout service.

No behavior change: nothing calls these yet. The deleter starts using them later in the stack. Checkout sessions on the billing provider are not touched; they expire on their own.

### DIFF
--- a/internal/store/postgres/billing_checkout_repository.go
+++ b/internal/store/postgres/billing_checkout_repository.go
@@ -304,6 +304,23 @@ func (r BillingCheckoutRepository) UpdateByID(ctx context.Context, toUpdate chec
 	return customerModel.transform()
 }
 
+func (r BillingCheckoutRepository) DeleteByCustomerID(ctx context.Context, customerID string) error {
+	query, params, err := dialect.Delete(TABLE_BILLING_CHECKOUTS).Where(goqu.Ex{
+		"customer_id": customerID,
+	}).ToSQL()
+	if err != nil {
+		return fmt.Errorf("%w: %s", errParse, err)
+	}
+
+	if err = r.dbc.WithTimeout(ctx, TABLE_BILLING_CHECKOUTS, "DeleteByCustomerID", func(ctx context.Context) error {
+		_, err := r.dbc.ExecContext(ctx, query, params...)
+		return err
+	}); err != nil {
+		return fmt.Errorf("%w: %s", errDB, err)
+	}
+	return nil
+}
+
 func (r BillingCheckoutRepository) List(ctx context.Context, flt checkout.Filter) ([]checkout.Checkout, error) {
 	stmt := dialect.Select().From(TABLE_BILLING_CHECKOUTS).Order(goqu.I("created_at").Desc())
 	if flt.CustomerID != "" {
--- a/internal/store/postgres/billing_checkout_repository.go
+++ b/internal/store/postgres/billing_checkout_repository.go
@@ -309,14 +309,14 @@ func (r BillingCheckoutRepository) DeleteByCustomerID(ctx context.Context, custo
 		"customer_id": customerID,
 	}).ToSQL()
 	if err != nil {
-		return fmt.Errorf("%w: %s", errParse, err)
+		return fmt.Errorf("%w: %w", errParse, err)
 	}
 
 	if err = r.dbc.WithTimeout(ctx, TABLE_BILLING_CHECKOUTS, "DeleteByCustomerID", func(ctx context.Context) error {
 		_, err := r.dbc.ExecContext(ctx, query, params...)
 		return err
 	}); err != nil {
-		return fmt.Errorf("%w: %s", errDB, err)
+		return fmt.Errorf("%w: %w", errDB, err)
 	}
 	return nil
 }
--- a/billing/checkout/service.go
+++ b/billing/checkout/service.go
@@ -63,6 +63,7 @@ type Repository interface {
 	Create(ctx context.Context, ch Checkout) (Checkout, error)
 	UpdateByID(ctx context.Context, ch Checkout) (Checkout, error)
 	List(ctx context.Context, filter Filter) ([]Checkout, error)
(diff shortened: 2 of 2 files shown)


whoAbhishekSah:
## End-to-end testing of the whole stack

Posting here because this is the PR that lands on `main`, so it carries the whole series.

Tested at the tip of the stack, which contains all six PRs: #1841 → #1842 → #1843 → #1844 → #1838 → #1839. That covers checkout delete, credit delete, kyc delete, the retryable subscription teardown, the deleter changes, and the new checkout audit records.

Run against a real Stripe test-mode account, so every provider call actually went to Stripe. No mock and no recorded cassettes.

### Setup

- Fresh Postgres and SpiceDB databases, both migrated from scratch.
- Platform permissions, roles, platform users and preferences applied with `frontier reconcile`.
- A Stripe test-mode key in `billing.stripe_key`.
- One plan created through `CreatePlan` ($10/month, 100 start credits). Frontier registered the product and price in Stripe.
- Seven orgs seeded through the normal RPCs to cover the different billing shapes. Billing accounts, usage, checkout sessions and KYC records were all created the way a client would create them.
- Every delete was driven through `FrontierService/DeleteOrganization`.
- After each delete I checked three places: the local tables, the `auditlogs` rows, and the Stripe API directly.

### Results

| Case | What the org had | Expected | Result |
| --- | --- | --- | --- |
| Invoice guard | Online account, 1 paid invoice, 1 subscription, 1 checkout | Refused, nothing removed | Pass. Org, subscription and checkout all untouched |
| Simple account | Online account, nothing else | Deleted | Pass. Customer shows `deleted: true` in Stripe |
| Full billing data | 2 checkouts, 2 credit rows, KYC record | Deleted, 1 audit record per checkout | Pass. Both rows gone, 2 audit records written |
| Provider drift | 1 checkout, 2 credit rows; customer deleted in Stripe behind Frontier's back | Deleted anyway | Pass. `resource_missing` treated as already gone |
| Active subscription | Active subscription + paid invoice | Refused while the invoice exists | Pass. After clearing the invoice row the delete ran and the subscription went to `canceled` in Stripe |
| Offline account | Offline account (no provider id), 2 credit rows, KYC record | Deleted with no provider calls | Pass |
| Failure then retry | 2 checkouts, 2 credit rows, KYC record. A database trigger blocked the billing account delete | Fails safely, retry finishes | Pass, details below |

Extra checks:

- Orphan sweep across the whole database after all deletes: 0 orphaned checkouts, subscriptions, invoices, KYC rows and billing accounts.
- Unit tests pass for all five changed packages.

### What the retry case showed

I forced a realistic mid-teardown failure with a trigger on `billing_customers`. The behaviour matched what the code comments claim:

- Checkouts and credit rows were removed and the audit records were written.
- The org row survived **and kept its owner policy**, so the org was still owned and still reachable.
- After removing the trigger, the same call finished the delete.
- The checkout audit count stayed at 2. Retrying does not write duplicate records, because the checkout list is empty the second time.

### Audit record contents

Each `app.billing.checkout.deleted` record carried the Stripe session id (`cs_test_...`), the customer id, the plan id, the state and the payment status. Six checkouts were removed across the test orgs and six records were written, one each. So the sessions can still be traced on the provider after the local rows are gone, which is the point of #1839.

### Two issues found, neither introduced by this stack

**1. A refused delete looks like a crash to the caller.** When the invoice guard blocks a delete, the client gets:

```json
{"code":"internal","message":"internal server error"}
```

The real reason never reaches the caller. `internal/api/v1beta1connect/deleter.go` wraps every error as `CodeInternal`, so `deleter.ErrDeleteNotAllowed` is lost and a normal business rule reads as a server fault. Mapping that error to `CodeFailedPrecondition` would fix it. Can raise this separately.

**2. Credit ledger rows on the system account stay behind.** Each usage debit writes two rows: one on the customer account and one on the system account (`00000000-0000-0000-0000-000000000000`). `DeleteByAccountID` clears the customer side only, so the system-side rows remain and still carry descriptions naming orgs that no longer exist. Nothing breaks and the customer delete is not blocked, since there is no foreign key on `billing_transactions.account_id`. Flagging it because it leaves the ledger one-sided. It may well be intended for accounting history.

internal/store/postgres/billing_checkout_repository.go
```diff
@@ -304,6 +304,23 @@ func (r BillingCheckoutRepository) UpdateByID(ctx context.Context, toUpdate chec
 	return customerModel.transform()
 }
 
+func (r BillingCheckoutRepository) DeleteByCustomerID(ctx context.Context, customerID string) error {
+	query, params, err := dialect.Delete(TABLE_BILLING_CHECKOUTS).Where(goqu.Ex{
+		"customer_id": customerID,
+	}).ToSQL()
+	if err != nil {
+		return fmt.Errorf("%w: %s", errParse, err)
```
rohilsurana: Small nit: this wraps the inner error with `%s`, so the cause drops out of the `errors.Is` chain. The credit and kyc PRs in this stack use `%w: %w`. Consider doing the same here (and on the `errDB` return below), since the stack is all about making teardown retryable and inspectable.

rohilsurana (commented):
Reviewed as part of the stack (#1841 through #1839). The logic looks good: parameterized delete, the customer_id column is indexed, and there is no existing mock for the checkout Repository, so nothing needed regenerating. One small nit inline.

rohilsurana (approved):

internal/store/postgres/billing_checkout_repository.go
whoAbhishekSah: Fixed in 9deb1dc7 — both returns now use `%w: %w`.

PR #1842, opened 2026-08-04
feat(credit): add delete of credit transactions by billing account

whoAbhishekSah:
Part of #1835. Stacked on the checkout delete PR.

Credit transactions also hold a foreign key to the billing customer row. Without a delete method, an account teardown would either orphan them or fail on the constraint.

## Changes

- `DeleteByAccountID` on the transactions Postgres repository.
- `DeleteByAccountID` on the credit service.
- Regenerated the transaction repository mock.

No behavior change: nothing calls these yet.


billing/credit/service.go
```diff
@@ -175,6 +176,12 @@ func (s Service) GetByID(ctx context.Context, id string) (Transaction, error) {
 	return s.transactionRepository.GetByID(ctx, id)
 }
 
+// DeleteByAccountID removes all credit transactions of a billing account. It is
+// meant for account teardown; the deletion is recorded in audit records.
```
rohilsurana: This says the deletion is recorded in audit records, but nothing in the stack writes audit records for credit transactions. Only checkout deletes get them (#1839). Suggest dropping the sentence, or adding the audit record if it was intended.

rohilsurana (commented):
One inline note on the doc comment. Also, the PR description says credit transactions hold a foreign key to the billing customer row. `billing_transactions.account_id` has no REFERENCES constraint (see `20231012175641_create_billing_tables.up.sql`). The delete is still worth having, since it avoids orphaned rows, but the description overstates the constraint. Worth updating so reviewers do not take the FK claim at face value.

rohilsurana (approved):

billing/credit/service.go
whoAbhishekSah: Dropped the sentence in 8ac4120e. Only checkout deletes get audit records in this stack. If we decide credit deletions need them too (related: the system-account side of the ledger survives the org delete), that can be its own small PR on the same pattern as #1839.

PR #1843, opened 2026-08-04
feat(kyc): add delete of an org kyc record

whoAbhishekSah:
Part of #1835. Stacked on the credit delete PR.

The `org_kyc` row references the organization row, and nothing could remove it, which blocked the hard delete of an organization.

## Changes

- `Delete` on the kyc Postgres repository.
- `DeleteKyc` on the kyc service.
- Regenerated the kyc repository mock.

No behavior change: nothing calls these yet.


rohilsurana (commented):
Reviewed as part of the stack. No suggestions here. The idempotent delete is the right call for a retryable teardown, and the error mapping matches the rest of the file.

rohilsurana (approved):

PR #1844, opened 2026-08-04
fix(subscription): make billing account teardown retryable

whoAbhishekSah:
Part of #1835. Stacked on the kyc delete PR.

`DeleteByCustomer` failed and could not be run again when the provider side was already gone. That mattered for org delete: after a halfway failure, the Stripe customer was deleted but local rows remained, and every retry died on the first provider call.

## Changes

- A subscription already canceled on the provider now just syncs the local state instead of failing the cancel call.
- A subscription missing on the provider counts as canceled.
- Offline accounts skip the provider entirely; only local records are removed.
- Dropped the `SyncWithProvider` call, which failed outright when the provider customer was already deleted.

### DIFF
--- a/billing/subscription/service.go
+++ b/billing/subscription/service.go
@@ -304,9 +304,18 @@ func (s *Service) Cancel(ctx context.Context, id string, immediate bool) (Subscr
 	}
 
 	// check if schedule exists
-	_, stripeSchedule, err := s.createOrGetSchedule(ctx, sub)
+	stripeSubscription, stripeSchedule, err := s.createOrGetSchedule(ctx, sub)
 	if err != nil {
 		return sub, err
+	}
+
+	if stripeSubscription != nil && stripeSubscription.Status == stripe.SubscriptionStatusCanceled {
+		// already canceled on the provider, just sync the local state
+		sub.State = string(stripeSubscription.Status)
+		if stripeSubscription.CanceledAt > 0 {
+			sub.CanceledAt = utils.AsTimeFromEpoch(stripeSubscription.CanceledAt)
+		}
+		return s.repository.UpdateByID(ctx, sub)
 	}
 
 	if immediate || stripeSchedule == nil {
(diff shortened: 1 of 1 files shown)


billing/subscription/service.go
```diff
@@ -304,11 +304,20 @@ func (s *Service) Cancel(ctx context.Context, id string, immediate bool) (Subscr
 	}
 
 	// check if schedule exists
-	_, stripeSchedule, err := s.createOrGetSchedule(ctx, sub)
+	stripeSubscription, stripeSchedule, err := s.createOrGetSchedule(ctx, sub)
 	if err != nil {
 		return sub, err
 	}
 
+	if stripeSubscription != nil && stripeSubscription.Status == stripe.SubscriptionStatusCanceled {
```
rohilsurana: `createOrGetSchedule` also returns a nil schedule for `incomplete` and `incomplete_expired` statuses. Those fall through to `Subscriptions.Cancel` below. Stripe rejects a cancel of an `incomplete_expired` subscription, and that error is not `ErrSubscriptionOnProviderNotFound`, so `DeleteByCustomer` would stop there. It needs local state to say active while the provider says expired, so it is rare, but treating those statuses like canceled here would close the gap.

rohilsurana (commented):
The teardown logic checks out: the early return syncs local state when the provider already canceled, and dropping SyncWithProvider is safe because deleting the Stripe customer later cancels anything still live on the provider. One edge case inline.

Also, no tests cover the new Cancel branch, which is the one piece of real logic in this stack. The concrete Stripe client makes unit tests hard here, so a short note in the PR on how this was verified against Stripe test mode would help.

rohilsurana (approved):

billing/subscription/service.go
whoAbhishekSah: Good catch. Fixed in d2242b34: `incomplete_expired` now takes the sync-local-state path, same as provider-canceled, since it is terminal and Stripe rejects a cancel for it. `incomplete` still goes through the real cancel because Stripe allows canceling those, and we want it actually terminated rather than left able to activate later.

PR #1847, opened 2026-08-04
feat(admin): list pending invites in user details side panel

Shreyag02:
## Summary

The user details side panel now lists the organizations a user has been invited to but hasn't joined yet, sitting below their existing memberships. Data comes from `FrontierService/ListUserInvitations`.

## Changes

- **New `Invitations` section** in the user details side panel — one block per invite, showing the org, the role offered, a `Status` (`Pending` / `Expired`), and a relative expiry (`in 7 days`, flipping to `5 days ago` once lapsed).
- **New `useOrganizationRoles` hook** (`admin/hooks/`) that fetches the platform's default org roles plus the org's custom roles, returning both a `roles` list and an id → title map.
- **`membership-dropdown.tsx` now uses that hook** (+2 / −40). It previously had its own copy of the same two queries, so both places resolve role titles identically from here on.
- **Side panel header colour** — `List.Header` renders its text in a tertiary-coloured inner span; overridden to `--rs-color-foreground-base-primary` to match design, applied to all three sections.

## Technical Details

- **`ListUserInvitations` is keyed by the invitee's email, not the user uuid** — an invitation can exist before the user does. Worth knowing because passing the uuid returns `200` with an empty list rather than an error, so getting it wrong fails silently. The query uses `user?.email` and is gated on it.
- **The API returns expired invitations too.** There's no server-side `expires_at` filter; rows live until the cleanup cron sweeps them. So `Pending` / `Expired` is derived client-side from `expires_at` vs. now.
- **Expiry text comes from dayjs `relativeTime` with the stock `en` locale**, so it reads `in 7 days` / `5 days ago`. The new `formatInviteExpiry` helper in `admin/utils/connect-timestamp.ts` returns the text and the `isExpired` flag together, and that module is where `dayjs.extend(relativeTime)` is installed (a one-time global extend, so any other dayjs consumer in the bundle gains `fromNow` too). An earlier revision registered a custom `en-invite` locale to get `7 days left` phrasing; that's been dropped — the wording gain wasn't worth a locale registration living in a shared util, and stock output already distinguishes the two directions.
- An invitation carries only `org_id` and `role_ids`, so the org title/avatar comes from the existing `useOrganizationLookup` and role titles from the new hook. react-query caches both per key, so multiple invite blocks share one fetch.
- Invitations and memberships have independent loading/error branches — a failure in one won't blank out the other section.
- Role fetch errors now log once from a `useEffect` inside the hook. The dropdown previously logged them from the render body, which repeated on every render.

## Test Plan

Verified:
- `tsc --noEmit`: 21 errors, all pre-existing and in untouched files (`organizations/details/edit/billing.tsx`, `client/utils`, `tsup.config.ts`, …); 0 in the files this PR touches.
- `eslint` clean on all touched files; `pnpm build` succeeds in `web/sdk`.
- Expiry rendering checked across the boundaries:

  | Case | Renders | Status |
  | --- | --- | --- |
  | 7-day invite, created today | `in 7 days` | Pending |
  | 1 day remaining | `in a day` | Pending |
  | 3 hours remaining | `in 3 hours` | Pending |
  | 30 minutes remaining | `in 30 minutes` | Pending |
  | 30 minutes lapsed | `30 minutes ago` | Expired |
  | lapsed 5 days ago | `5 days ago` | Expired |

Not applicable — frontend only, no Go or query-builder changes.


web/sdk/admin/views/users/details/layout/side-panel-invitation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const hours = to.diff(from, "hour");
+  if (hours >= 1) {
+    return {
+      text: `${hours} hour${hours === 1 ? "" : "s"} ${suffix}`,
+      isExpired,
+    };
+  }
+
+  return { text: `Less than an hour ${suffix}`, isExpired };
+}
```
rohanchkrabrty: We can use dayjs relativeTime plugin 
https://day.js.org/docs/en/plugin/relative-time

rohanchkrabrty (changes requested):

web/sdk/admin/views/users/details/layout/side-panel-invitation.tsx
Shreyag02: updated and also dropped custom invite expiry locale

rohanchkrabrty (approved):

Issue #1836, opened 2026-08-04
Billing APIs return "internal" for errors the caller can act on


whoAbhishekSah:
## The problem

Almost every billing handler wraps failures in `connect.CodeInternal`. That code means "something broke inside Frontier". But many of these failures are not Frontier bugs — they describe the state of the caller's billing account, and often the caller can fix them.

Returning `internal` for these cases causes two problems:

- **For the user**: the error says nothing. They cannot tell whether they did something wrong, whether their account is in a bad state, or whether Frontier is down. There is nothing to act on, so they either retry blindly or file a support ticket.
- **For anyone running Frontier**: `internal` errors stop meaning anything. Real bugs and ordinary account states look identical in logs and metrics, so genuine failures are hard to spot.

The clearest example: if an org's billing account points to a Stripe customer that no longer exists (see #1835 for how that happens), `GetUpcomingInvoice` returns `internal` on every single call — even though nothing in Frontier is broken.

The root cause is that we inspect Stripe errors in only four narrow places in the billing services. Everywhere else, the raw `stripe.Error` bubbles up to the handler, and the handler's fallback for any unrecognized error is `internal`. There are ~88 such fallback wraps across the billing handlers.

RPCs affected by state-caused failures that show as `internal` today:

| RPC | Example trigger |
|---|---|
| `GetUpcomingInvoice` | billing account points to a deleted Stripe customer |
| `GetBillingAccount` (with payment methods) | same |
| `CreateCheckout`, `DelegatedCheckout` | checkout against a deleted Stripe customer, payment failures |
| `CancelSubscription` | subscription already canceled, or already gone on Stripe |
| `ChangeSubscription` | plan change already in progress ("try again later") |
| `CreateBillingAccount`, `UpdateBillingAccount`, `RegisterBillingAccount` | Stripe rejects the input, or the Stripe customer is gone |

## The fix

Fix it once at each layer instead of patching 88 call sites:

1. **A Stripe error translator in the billing package.** One helper that every service wraps its Stripe calls with. It turns `*stripe.Error` into typed domain errors:
   - `resource_missing` → `ErrProviderResourceMissing`
   - card and payment errors → `ErrPaymentFailed` (keep Stripe's human-readable message)
   - rate limits and Stripe outages → `ErrProviderUnavailable`
   - anything else stays as-is
   After this, no raw Stripe error ever leaves the billing services.

2. **A shared error mapper for billing handlers.** Used as the fallback instead of bare `CodeInternal`:
[...]

PR #1849, opened 2026-08-05
feat(billing): add stripe error translator

whoAbhishekSah:
## What

Adds a `billing/errors` package (imported as `billingerrors`, same convention as `core/userpat/errors`) with three typed errors and one function, `TranslateStripeError`, that classifies a stripe error into them:

- `resource_missing` → `ErrProviderResourceMissing` (the record is gone on Stripe)
- card errors and decline codes → `ErrPaymentFailed`
- rate limits, stripe server errors, HTTP 429/5xx → `ErrProviderUnavailable`

Anything else is returned unchanged.

The translated error is a `ProviderError`. Its message keeps Stripe's human-readable text (`stripe.Error.Msg`) instead of the raw JSON blob that `stripe.Error.Error()` prints. The original error stays in the chain, so `errors.As` can still reach the `stripe.Error` when needed.

## Why

Almost every billing failure today reaches the caller as a bare "internal server error", even when the problem is the state of their billing account (see #1836). This translator is the first layer of the fix: it gives the services and handlers a typed error they can act on.

Part of #1836. Stack: this PR → services adopt the translator → handlers map the errors to proper codes.

### DIFF
--- a/billing/errors/errors.go
+++ b/billing/errors/errors.go
@@ -1,0 +1,59 @@
+package errors
+
+import (
+	"errors"
+	"net/http"
+
+	stripe "github.com/stripe/stripe-go/v79"
+)
+
+var (
+	ErrProviderResourceMissing = errors.New("record no longer exists on the billing provider")
+	ErrPaymentFailed           = errors.New("payment failed")
+	ErrProviderUnavailable     = errors.New("billing provider is unavailable")
+)
+
+// ProviderError is a billing provider failure classified as one of the
+// Err* kinds above. Message carries the provider's human-readable message.
+type ProviderError struct {
+	Kind    error
+	Message string
+	cause   error
+}
+
+func (e *ProviderError) Error() string {
+	if e.Message == "" {
+		return e.Kind.Error()
+	}
+	return e.Kind.Error() + ": " + e.Message
+}
+
+func (e *ProviderError) Unwrap() []error {
+	return []error{e.Kind, e.cause}
+}
+
+// TranslateStripeError converts a stripe error into a *ProviderError so
+// callers can match it with errors.Is against the Err* kinds. Errors that
+// don't match a known kind are returned unchanged.
+func TranslateStripeError(err error) error {
+	var stripeErr *stripe.Error
+	if err == nil || !errors.As(err, &stripeErr) {
+		return err
+	}
+
+	var kind error
+	switch {
+	case stripeErr.Code == stripe.ErrorCodeResourceMissing:
+		kind = ErrProviderResourceMissing
+	case stripeErr.Type == stripe.ErrorTypeCard || stripeErr.DeclineCode != "":
+		kind = ErrPaymentFailed
+	case stripeErr.Code == stripe.ErrorCodeRateLimit,
+		stripeErr.Type == stripe.ErrorTypeAPI,
+		stripeErr.HTTPStatusCode == http.StatusTooManyRequests,
+		stripeErr.HTTPStatusCode >= http.StatusInternalServerError:
+		kind = ErrProviderUnavailable
+	default:
+		return err
+	}
+	return &ProviderError{Kind: kind, Message: stripeErr.Msg, cause: err}
+}
--- a/billing/errors/errors.go
+++ b/billing/errors/errors.go
@@ -1,8 +1,11 @@
 package errors
 
 import (
+	"context"
 	"errors"
+	"net"
 	"net/http"
+	"net/url"
 
 	stripe "github.com/stripe/stripe-go/v79"
 )
@@ -14,11 +17,13 @@ var (
 )
 
 // ProviderError is a billing provider failure classified as one of the
-// Err* kinds above. Message carries the provider's human-readable message.
+// Err* kinds above. Message carries the provider's human-readable message,
+// RequestID the provider's id for the failed request (for support tickets).
 type ProviderError struct {
-	Kind    error
-	Message string
-	cause   error
+	Kind      error
+	Message   string
+	RequestID string
+	cause     error
 }
 
 func (e *ProviderError) Error() string {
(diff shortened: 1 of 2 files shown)


whoAbhishekSah:
Tested the full stack (#1849 translator -> #1850 service translation -> #1851 handler mapper) live against a local Frontier with real Stripe test mode. Fresh org owned by a normal user; org-level calls made with that user's session, admin calls with a platform admin. Note: DelegatedCheckout is an AdminService RPC — the caller in those rows is a platform admin or internal automation, not the end user.

**New error codes, all verified live** (every one of these was `internal` before the stack):

| Trigger | Result | What the caller can do now |
|---|---|---|
| Stripe customer deleted, then UpdateBillingAccount / CreateCheckout (setup + subscription) / DelegatedCheckout (admin) / GetUpcomingInvoice / GetBillingAccount with payment methods | `failed_precondition` | Stop retrying (it will never succeed). Show "billing account is unlinked" instead of a generic error. The org admin can delete and recreate the billing account, or contact support. Before: retries and a dead-end error toast. |
| DelegatedCheckout (admin) with no payment method on a paid plan | `failed_precondition` | The admin or automation driving the checkout learns the org has no card on file and can ask the org to add a payment method (checkout setup session) before granting the plan. |
| DelegatedCheckout (admin) with a coupon id that does not exist on Stripe | `failed_precondition` with the Stripe message ("No such coupon") | The admin sees the bad coupon id in the error and fixes the request — no log-diving needed. |
| CancelSubscription / cancel upcoming phase when the Stripe subscription is gone | `failed_precondition` "subscription no longer exists on the billing provider" | Treat the subscription as dead: stop retrying, re-subscribe or contact support. |
| New billing account while the previous one has a negative credit balance | `failed_precondition` "existing account with pending dues found" | Show the user the real blocker: clear the dues first, then create the account. |
| DeleteBillingAccount when the Stripe customer is already deleted | still succeeds (tolerance kept) | Cleanup and org-delete flows don't get stuck on a half-deleted account. |

The common thread: `internal` tells a client "server bug, retry later" — wrong on both counts for these cases, and the interceptor masks the message so the caller learns nothing. `failed_precondition` says "your request was fine, the account state is not; retrying won't help, do X first", which is what all of these are.

`ErrPaymentFailed` and `ErrProviderUnavailable` cannot be raised synchronously against real Stripe (subscriptions are created with `allow_incomplete`, and Stripe 5xx/rate limits can't be forced), so those two are covered by the unit tests only. Same for `ErrPhaseIsUpdating`, which only occurs in a race at a phase boundary.

**Happy paths regressed nothing**: account create/get/list/update, balance, checkout sessions (real Stripe URLs), delegated subscription (went active, card charged), plan change with scheduled phase, cancel upcoming phase, cancel subscription, upcoming invoice, HasTrialed. Unchanged codes confirmed: `not_found`, `unauthenticated`, `permission_denied` for a non-member, active-account conflict.

Testing also surfaced four cases that still returned `internal` or a misleading message. Fixed on #1851 (commit 8252b15):

| Case | Now | What the caller can do now |
|---|---|---|
| Missing coupon / payment method reported as "billing account is no longer linked" | provider's own message passes through | See the actual cause instead of a wrong diagnosis. |
| CreateCheckout on an org with no billing account | `not_found` "customer doesn't exist" | Create the billing account first, then check out. |
| DelegatedCheckout (admin) for a plan the org is already on | `already_exists` | The admin or automation treats it as done / shows "already subscribed"; no support ticket for a non-error. |
| GetProduct / CheckFeatureEntitlement with unknown id | `not_found` | Fix the product/feature name in the request. |

All re-verified live after the fix.

whoAbhishekSah:
One follow-up on the test report above (commit 8709ad9 on #1851): provider messages passed through to the caller now mask provider-generated object ids. A deleted customer reads as "No such customer: 'cus_*****'" instead of showing the real Stripe customer id. Caller-supplied values don't match the id shape, so "No such coupon: 'SUMMER20'" is unchanged. Covered by unit tests on the mapper.

billing/errors/errors_test.go
```diff
@@ -0,0 +1,99 @@
...
+				assert.Equal(t, tt.err, got)
+				return
+			}
+			assert.ErrorIs(t, got, tt.wantKind)
+			if tt.wantMsg != "" {
+				assert.Equal(t, tt.wantMsg, got.Error())
+			}
+
+			var stripeErr *stripe.Error
+			assert.ErrorAs(t, got, &stripeErr)
```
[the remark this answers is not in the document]
whoAbhishekSah: Done in 65732eae: passthrough cases assert identity with assert.Same, translated cases assert the original error stays in the chain.

billing/errors/errors.go
```diff
@@ -0,0 +1,59 @@
...
+	var kind error
+	switch {
+	case stripeErr.Code == stripe.ErrorCodeResourceMissing:
+		kind = ErrProviderResourceMissing
+	case stripeErr.Type == stripe.ErrorTypeCard || stripeErr.DeclineCode != "":
+		kind = ErrPaymentFailed
+	case stripeErr.Code == stripe.ErrorCodeRateLimit,
+		stripeErr.Type == stripe.ErrorTypeAPI,
+		stripeErr.HTTPStatusCode == http.StatusTooManyRequests,
+		stripeErr.HTTPStatusCode >= http.StatusInternalServerError:
```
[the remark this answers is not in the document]
whoAbhishekSah: Keeping >= 500 as is. Status codes above 599 don't occur in real HTTP responses, and if a broken proxy ever produced one, classifying it as provider-unavailable (retryable) is more useful to the caller than internal. Capping at 599 would only change behavior for a status that can't happen.

billing/errors/errors.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (e *ProviderError) Error() string {
+	if e.Message == "" {
+		return e.Kind.Error()
+	}
+	return e.Kind.Error() + ": " + e.Message
+}
+
+func (e *ProviderError) Unwrap() []error {
+	return []error{e.Kind, e.cause}
```
rohilsurana: If someone builds a `ProviderError` outside this package, `cause` stays nil (it is unexported), and this returns a slice with a nil entry. `errors.Is` and `errors.As` skip nil entries today, but the stdlib docs say the slice from `Unwrap() []error` should not contain nil. Cheap to guard:

```go
func (e *ProviderError) Unwrap() []error {
	errs := []error{e.Kind}
	if e.cause != nil {
		errs = append(errs, e.cause)
	}
	return errs
}
```

billing/errors/errors.go
```diff
@@ -0,0 +1,59 @@
...
+		return err
+	}
+
+	var kind error
+	switch {
+	case stripeErr.Code == stripe.ErrorCodeResourceMissing:
+		kind = ErrProviderResourceMissing
+	case stripeErr.Type == stripe.ErrorTypeCard || stripeErr.DeclineCode != "":
+		kind = ErrPaymentFailed
+	case stripeErr.Code == stripe.ErrorCodeRateLimit,
```
rohilsurana, line 76: Note: network failures (timeout, connection refused) do not produce a `*stripe.Error`, so a hard outage often passes through untranslated and stays internal instead of unavailable. Fine for this layer, just flagging it in case you want a follow-up for that case.

billing/errors/errors.go
```diff
@@ -0,0 +1,59 @@
...
+		stripeErr.Type == stripe.ErrorTypeAPI,
+		stripeErr.HTTPStatusCode == http.StatusTooManyRequests,
+		stripeErr.HTTPStatusCode >= http.StatusInternalServerError:
+		kind = ErrProviderUnavailable
+	default:
+		return err
+	}
+	return &ProviderError{Kind: kind, Message: stripeErr.Msg, cause: err}
```
rohilsurana: The message keeps `stripeErr.Msg` but drops `stripeErr.RequestID`. That request id is what Stripe support asks for when you report a problem. Today the raw JSON from `stripe.Error.Error()` lands in logs and includes it. After this stack, this message is all that reaches the logs (the logger interceptor logs the error string the handler returns), so the request id is gone everywhere.

Consider adding it to the message, or logging the raw stripe error at the call sites. If it goes into the message, #1851 should redact `req_...` too.

rohilsurana (commented):
Reviewed the whole stack (#1849 -> #1850 -> #1851). This one looks good. Small package with one clear job, and the table tests cover the classification well, including wrapped errors and keeping the kinds apart.

Notes:
- Package name `errors` with the `billingerrors` import alias matches the `core/userpat/errors` convention. Works for me.
- Classification precedence looks right: resource missing first, then card/decline, then availability. I checked stripe-go v79: it has no rate limit error type, so matching on the code plus HTTP 429/5xx is the right way with this SDK.

A few small comments inline.

rohilsurana (approved):
LGTM

PR #1850, opened 2026-08-05
fix(billing): translate stripe errors in billing services

whoAbhishekSah:
## What

Wraps every stripe call in the billing services with `TranslateStripeError` from #1849, so a typed provider error leaves the service instead of a raw stripe error. Touches the customer, subscription, checkout, product, and invoice services — all 30 stripe call sites.

Along the way:

- The old `err.(*stripe.Error)` type assertions become `errors.Is` / `errors.As`, which also match wrapped errors. The type assertion silently failed on any error that had been wrapped.
- The product and invoice calls that returned bare errors now carry context ("failed to create price at billing provider: ...").
- `ListPaymentMethods` never checked the stripe iterator's error, so a failed list looked like an empty list. It now returns the error.

Behavior kept as before: a customer already deleted on Stripe is still tolerated during delete, and a missing subscription still maps to `ErrSubscriptionOnProviderNotFound`.

Part of #1836. Stack: translator (#1849) → this PR → handler error mapping.

### DIFF
--- a/billing/customer/service.go
+++ b/billing/customer/service.go
@@ -2,6 +2,7 @@ package customer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"math/rand"
@@ -14,6 +15,7 @@ import (
 	"github.com/stripe/stripe-go/v79"
 
 	"github.com/raystack/frontier/billing"
+	billingerrors "github.com/raystack/frontier/billing/errors"
 	"github.com/raystack/frontier/internal/metrics"
 
 	"slices"
@@ -134,14 +136,11 @@ func (s *Service) RegisterToProvider(ctx context.Context, customer Customer) (*s
 		TestClock: customer.StripeTestClockID,
 	})
 	if err != nil {
-		if stripeErr, ok := err.(*stripe.Error); ok {
-			switch stripeErr.Code {
-			case stripe.ErrorCodeParameterMissing:
-				// stripe error
-				return nil, fmt.Errorf("missing parameter while registering to biller: %s", stripeErr.Error())
-			}
+		var stripeErr *stripe.Error
+		if errors.As(err, &stripeErr) && stripeErr.Code == stripe.ErrorCodeParameterMissing {
+			return nil, fmt.Errorf("missing parameter while registering to biller: %s: %w", stripeErr.Msg, err)
 		}
-		return nil, fmt.Errorf("failed to register in billing provider: %w", err)
+		return nil, fmt.Errorf("failed to register in billing provider: %w", billingerrors.TranslateStripeError(err))
 	}
 
 	return stripeCustomer, nil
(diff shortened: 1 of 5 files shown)


billing/customer/service.go
```diff
@@ -134,14 +136,11 @@ func (s *Service) RegisterToProvider(ctx context.Context, customer Customer) (*s
...
 	if err != nil {
-		if stripeErr, ok := err.(*stripe.Error); ok {
-			switch stripeErr.Code {
-			case stripe.ErrorCodeParameterMissing:
-				// stripe error
-				return nil, fmt.Errorf("missing parameter while registering to biller: %s", stripeErr.Error())
-			}
+		var stripeErr *stripe.Error
+		if errors.As(err, &stripeErr) && stripeErr.Code == stripe.ErrorCodeParameterMissing {
+			return nil, fmt.Errorf("missing parameter while registering to biller: %s", stripeErr.Error())
```
rohilsurana: This branch still prints the raw stripe JSON blob (`stripeErr.Error()`) and uses `%s`, which cuts the error chain. Since these lines were rewritten anyway, `stripeErr.Msg` would give a readable message and `%w` on the original `err` would keep the chain for `errors.As`. Same pattern at line 198 in `Update`.

billing/customer/service.go
```diff
@@ -352,6 +340,9 @@ func (s *Service) ListPaymentMethods(ctx context.Context, id string) ([]PaymentM
 
 		paymentMethods = append(paymentMethods, pm)
 	}
+	if err := stripePaymentMethodItr.Err(); err != nil {
```
rohilsurana, line 343: Good catch on the missed iterator error. One behavior note: `GetBillingAccount` with `with_payment_methods=true` now fails the whole request when this list fails, where before the caller got the account with an empty list. I think failing is more correct, just calling it out.

rohilsurana (commented):
I verified the coverage claim on this branch: 32 `stripeClient` call sites across the five services, and every one either wraps its error with `TranslateStripeError` or reaches it through an iterator `.Err()` check that wraps. No `err.(*stripe.Error)` type assertions are left under `billing/`.

The two tolerated cases keep their old behavior: a customer already deleted on Stripe is still tolerated during Delete, and a missing subscription still becomes `ErrSubscriptionOnProviderNotFound`. The `ErrorCodeInvoiceUpcomingNone` special case in `GetUpcoming` is also preserved.

The `ListPaymentMethods` fix is a real bug fix. A failed list used to look like an empty list.

Two small comments inline.

rohilsurana (approved):
LGTM, verified all stripe call sites are covered

billing/customer/service.go
whoAbhishekSah: Yes, intentional — failing the request is more correct than silently returning the account with an empty payment method list. No change made here.

PR #1851, opened 2026-08-05
fix(api): map billing errors to actionable codes instead of internal

### LINKED ISSUE #1836: Billing APIs return "internal" for errors the caller can act on

whoAbhishekSah:
## What

Adds one shared fallback, `mapBillingError`, and uses it in place of the ~87 bare `connect.CodeInternal` wraps across the billing handlers:

| error | code | message |
|---|---|---|
| `billing.ErrProviderResourceMissing` | `failed_precondition` | billing account is no longer linked to the payment provider |
| `billing.ErrPaymentFailed` | `failed_precondition` | payment failed: \<Stripe's message\> |
| `billing.ErrProviderUnavailable` | `unavailable` | billing provider is unavailable, retry later |
| `subscription.ErrSubscriptionOnProviderNotFound` | `failed_precondition` | subscription no longer exists on the billing provider |
| `subscription.ErrPhaseIsUpdating` | `failed_precondition` | phase is in the middle of a change, please try again later |
| `customer.ErrExistingAccountWithPendingDues` | `failed_precondition` | existing account with pending dues found |
| anything else | `internal` | unchanged |

The existing specific mappings in the handlers (`errors.Is` ladders for not-found, invalid-argument, and so on) are untouched; only the final fallback changes. The rewrite was done with `gofmt -r`, so every call keeps its context string.

## Why

`internal` should mean "a bug in Frontier". Today it also covers ordinary account states the caller can fix — the clearest case being `GetUpcomingInvoice` returning `internal` forever when the org's Stripe customer was deleted (#1835). With this change the caller learns whose move it is: fix the account, retry later, or report a bug.

Fixes #1836. Stack: translator (#1849) → services (#1850) → this PR.


internal/api/v1beta1connect/billing_errors.go
```diff
@@ -0,0 +1,62 @@
...
+// redactedProviderError hides provider object ids in a message shown to the
+// caller, keeping the rest of the provider's text.
+func redactedProviderError(providerErr *billingerrors.ProviderError) error {
+	return errors.New(providerIDPattern.ReplaceAllString(providerErr.Error(), "${1}_*****"))
+}
+
+// mapBillingError is the fallback for billing handlers in place of a bare
+// CodeInternal. Provider and account-state problems reach the caller as
+// codes they can act on; everything else stays internal.
+func mapBillingError(err error) *connect.Error {
```
rohilsurana: This drops the handler context for every mapped kind, and with it most of what lands in server logs.

The chain today: the handler builds `fmt.Errorf("GetUpcomingInvoice.GetUpcoming: org_id=%s billing_id=%s: %w", ...)`, and this function throws that away for the known kinds and returns a fresh error with only the clean message. The logger interceptor (`pkg/server/connect_interceptors/logger.go`) logs the error string the handler returns. So for exactly the failures this stack is about, the log line loses the method name, the org and billing ids, and the unredacted provider id. Before this PR these were internal errors, so the full wrapped string was logged at error level and the sanitizer hid it from the client.

Suggestion: log the original error before mapping. For example, make this `mapBillingError(ctx, err)`, log `err` here with the request-scoped logger, then return the clean error. That keeps the client-facing win without losing the debugging detail.

internal/api/v1beta1connect/billing_errors.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/raystack/frontier/billing/checkout"
+	"github.com/raystack/frontier/billing/customer"
+	billingerrors "github.com/raystack/frontier/billing/errors"
+	"github.com/raystack/frontier/billing/product"
+	"github.com/raystack/frontier/billing/subscription"
+)
+
+// provider-generated object ids (random alphanumeric after the prefix).
+// Caller-supplied values like coupon codes don't match this shape.
+var providerIDPattern = regexp.MustCompile(`\b(cus|sub_sched|sub|in|ii|il|pi|pm|seti|si|price|prod|cs|ch|coup|promo)_(?:(?:test|live)_)?[A-Za-z0-9]{8,}\b`)
```
rohilsurana: This allowlist fails open: any id whose prefix is not listed reaches the caller unmasked. Stripe has more prefixes that can show up in these messages (`txn`, `evt`, `plan`, `tok`, `src`, `ba`, `re`, `py`). Two options: add the missing ones, or flip to a general pattern like `[a-z]{2,10}_(?:(?:test|live)_)?[A-Za-z0-9]{8,}` and accept that an odd caller-supplied value gets masked too. Right now a new object type in a stripe message leaks its id.

internal/api/v1beta1connect/billing_errors.go
```diff
@@ -0,0 +1,62 @@
...
+// CodeInternal. Provider and account-state problems reach the caller as
+// codes they can act on; everything else stays internal.
+func mapBillingError(err error) *connect.Error {
+	switch {
+	case errors.Is(err, billingerrors.ErrProviderResourceMissing):
+		var providerErr *billingerrors.ProviderError
+		if errors.As(err, &providerErr) {
+			return connect.NewError(connect.CodeFailedPrecondition, redactedProviderError(providerErr))
+		}
+		return connect.NewError(connect.CodeFailedPrecondition, ErrBillingProviderResourceMissing)
```
rohilsurana: Small inconsistency: with a `ProviderError` in the chain the caller sees `record no longer exists on the billing provider: ...`, and on this fallback path they see `billing account is no longer linked to the payment provider`. Two different texts for the same condition, and the PR table only documents this one. This text also guesses the missing object is the billing account, which may be wrong (it could be a subscription or a checkout session). Fine if intentional, just flagging it.

rohilsurana (commented):
Reviewed with the rest of the stack. The mapping table is sensible, the mechanical rewrite kept every context string (spot checked), and the test table is thorough: redaction shapes, test-mode ids, and caller-supplied coupon codes are all covered. Promoting `already subscribed to the plan` to a sentinel with `already_exists` is a nice touch, and the `UpdateProduct` refactor keeps the `ErrInvalidDetail` to `invalid_argument` path intact.

One real concern about what ends up in server logs, and two smaller points, all inline.

Also checked `BillingWebhookCallback`: it now answers the webhook sender with 412/503 instead of 500 for provider problems. Stripe retries on any non-2xx, so retry behavior does not change.

rohilsurana (approved):
LGTM overall. Please look at the logging comment on mapBillingError before merge.

PR #1855, opened 2026-08-05
fix(db): report rollback outcome correctly in WithTxn

### LINKED ISSUE #1765: WithTxn reports rollback errors incorrectly and double-wraps the returned error
**Where:** `pkg/db/db.go`, `Client.WithTxn`

When the transaction callback returns an error, the deferred cleanup rolls the transaction back and wraps the error before returning it:

```go
} else if err != nil {
    if rlbErr := txn.Rollback(); err != nil {
        err = fmt.Errorf("rollback error: %s while executing: %w", rlbErr, err)
    } else {
        err = fmt.Errorf("rollback: %w", err)
    }
    err = fmt.Errorf("rollback: %w", err)
}
```

Two problems:

1. The inner `if` condition checks `err`, which is always non-nil in this branch. It should check `rlbErr`. Because of this the `else` branch is unreachable, and when the rollback succeeds the message still embeds the nil rollback error:
   `rollback: rollback error: %!s(<nil>) while executing: <original error>`
2. The error gets a second `rollback:` wrap after the inner `if`/`else`, so every error is prefixed twice.

Expected behavior:

- Rollback succeeds: return the original error with a single wrap, e.g. `rollback: <original error>`.
- Rollback fails: return both errors with a single wrap, e.g. `rollback error: <rollback error> while executing: <original error>`.
- The original error stays matchable with `errors.Is`/`errors.As` (keep `%w`).

AmanGIT07:
## Summary
`WithTxn` in `pkg/db` reported every failed transaction as a rollback failure: the message embedded a nil rollback error (`%!s(<nil>)`) and carried a duplicate `rollback:` prefix.

Fixes #1765

## Changes
- Check the rollback's own error (`rlbErr`) instead of the already non-nil callback error when building the returned message.
- Wrap the returned error once instead of twice.
- Remove unused error assignments in the panic path; it rolls back and re-panics as before.
- Add `pkg/db/db_test.go` covering commit, begin failure, rollback success and failure messages, commit failure, and panic handling, using a fake `database/sql/driver` (no new dependencies).
- Fix four inverted error assertions in `internal/store/postgres/user_repository_test.go`: `errors.Unwrap(err) == tc.Err` failed the test when the error matched, and the old double wrap kept it from firing. Now `!errors.Is(err, tc.Err)`.

## Technical Details
The callback error stays wrapped with `%w` on all paths, so `errors.Is`/`errors.As` matching is unchanged. Messages now are:
- rollback succeeds: `rollback: <callback error>`
- rollback fails: `rollback error: <rollback error> while executing: <callback error>`

## Test Plan
- [x] `go test -race -count=2 ./pkg/db/` passes
- [x] `go test ./internal/store/postgres/` passes
- [x] `golangci-lint run ./pkg/db/... ./internal/store/postgres/...` reports no issues

### DIFF
--- a/internal/store/postgres/user_repository_test.go
+++ b/internal/store/postgres/user_repository_test.go
@@ -366,8 +366,8 @@ func (s *UserRepositoryTestSuite) TestUpdateByEmail() {
 		s.Run(tc.Description, func() {
 			got, err := s.repository.UpdateByEmail(s.ctx, tc.UserToUpdate)
 			if tc.Err != nil && tc.Err.Error() != "" {
-				if errors.Unwrap(err) == tc.Err {
-					s.T().Fatalf("got error %s, expected was %s", err.Error(), tc.Err)
+				if !errors.Is(err, tc.Err) {
+					s.T().Fatalf("got error %v, expected was %v", err, tc.Err)
 				}
 			}
 
@@ -451,8 +451,8 @@ func (s *UserRepositoryTestSuite) TestUpdateByID() {
 		s.Run(tc.Description, func() {
 			got, err := s.repository.UpdateByID(s.ctx, tc.UserToUpdate)
 			if tc.Err != nil && tc.Err.Error() != "" {
-				if errors.Unwrap(err) == tc.Err {
-					s.T().Fatalf("got error %s, expected was %s", err.Error(), tc.Err)
+				if !errors.Is(err, tc.Err) {
+					s.T().Fatalf("got error %v, expected was %v", err, tc.Err)
 				}
 			}
 			// TODO(kushsharma): remove metadata field from ignore once metadata is refactored
@@ -559,8 +559,8 @@ func (s *UserRepositoryTestSuite) TestGetByName() {
 		s.Run(tc.Description, func() {
 			got, err := s.repository.GetByName(s.ctx, tc.Name)
 			if tc.Err != nil && tc.Err.Error() != "" {
-				if errors.Unwrap(err) == tc.Err {
-					s.T().Fatalf("got error %s, expected was %s", err.Error(), tc.Err)
+				if !errors.Is(err, tc.Err) {
+					s.T().Fatalf("got error %v, expected was %v", err, tc.Err)
 				}
 			}
 			if !cmp.Equal(got, tc.ExpectedUser) {
@@ -610,8 +610,8 @@ func (s *UserRepositoryTestSuite) TestUpdateByName() {
 		s.Run(tc.Description, func() {
 			got, err := s.repository.UpdateByName(s.ctx, tc.UserToUpdate)
 			if tc.Err != nil && tc.Err.Error() != "" {
-				if errors.Unwrap(err) == tc.Err {
-					s.T().Fatalf("got error %s, expected was %s", err.Error(), tc.Err)
+				if !errors.Is(err, tc.Err) {
+					s.T().Fatalf("got error %v, expected was %v", err, tc.Err)
 				}
 			}
 
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -80,7 +80,7 @@ func (c Client) WithTxn(ctx context.Context, txnOptions sql.TxOptions, txFunc fu
 			panic(p)
 		} else if err != nil {
 			if rlbErr := txn.Rollback(); rlbErr != nil {
-				err = fmt.Errorf("rollback error: %s while executing: %w", rlbErr, err)
+				err = fmt.Errorf("rollback error: %w while executing: %w", rlbErr, err)
 			} else {
 				err = fmt.Errorf("rollback: %w", err)
 			}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -76,21 +76,14 @@ func (c Client) WithTxn(ctx context.Context, txnOptions sql.TxOptions, txFunc fu
 
 	defer func() {
 		if p := recover(); p != nil {
-			switch p := p.(type) {
-			case error:
-				err = p
-			default:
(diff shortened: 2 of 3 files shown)


pkg/db/db.go
```diff
@@ -76,21 +76,14 @@ func (c Client) WithTxn(ctx context.Context, txnOptions sql.TxOptions, txFunc fu
 
 	defer func() {
 		if p := recover(); p != nil {
-			switch p := p.(type) {
-			case error:
-				err = p
-			default:
-				err = fmt.Errorf("%s", p)
-			}
-			err = txn.Rollback()
```
AmanGIT07, line 85: this is dead code, cleaned up

pkg/db/db.go
```diff
@@ -76,21 +76,14 @@ func (c Client) WithTxn(ctx context.Context, txnOptions sql.TxOptions, txFunc fu
...
+			_ = txn.Rollback()
 			panic(p)
 		} else if err != nil {
-			if rlbErr := txn.Rollback(); err != nil {
+			if rlbErr := txn.Rollback(); rlbErr != nil {
 				err = fmt.Errorf("rollback error: %s while executing: %w", rlbErr, err)
```
rohilsurana: Minor, non-blocking: `rlbErr` is formatted with `%s`, so `errors.Is(err, rlbErr)` will not match it. This module is on Go 1.26, which allows multiple `%w` verbs, so you could make both errors matchable:

```go
err = fmt.Errorf("rollback error: %w while executing: %w", rlbErr, err)
```

The callback error is the one callers actually match on and it is already wrapped, so this is just a nice-to-have.

pkg/db/db.go
```diff
@@ -76,21 +76,14 @@ func (c Client) WithTxn(ctx context.Context, txnOptions sql.TxOptions, txFunc fu
...
 	defer func() {
 		if p := recover(); p != nil {
-			switch p := p.(type) {
-			case error:
-				err = p
-			default:
-				err = fmt.Errorf("%s", p)
-			}
-			err = txn.Rollback()
+			_ = txn.Rollback()
```
rohilsurana, line 79: Confirming for the review: dropping the old `err = ...` and `err = txn.Rollback()` assignments here is safe. The deferred func re-panics, so the named `err` return is never observed by the caller. Behavior is unchanged and the code is cleaner.

pkg/db/db_test.go
```diff
@@ -0,0 +1,139 @@
...
+	beginErr    error
+	commitErr   error
+	rollbackErr error
+	commits     int
+	rollbacks   int
+}
+
+func (c *fakeConn) Prepare(string) (driver.Stmt, error) { return nil, errors.New("not implemented") }
+func (c *fakeConn) Close() error                        { return nil }
+func (c *fakeConn) Begin() (driver.Tx, error) {
```
rohilsurana, line 29: Note for future tests: `fakeConn` implements only `driver.Conn` (plain `Begin`), not `driver.ConnBeginTx`. That works here because the subtests pass `sql.TxOptions{}` (default isolation, not read-only), which lets `database/sql` fall back to `Begin()`. If a later test needs a non-default isolation level or read-only, the fake would need to implement `ConnBeginTx`.

rohilsurana (commented):
## Review

Clean, well-scoped bugfix. `WithTxn` used to report every failed transaction as a rollback failure. It printed the nil rollback error as `%!s(<nil>)` and added a duplicate `rollback:` prefix. This PR fixes all three underlying problems: the wrong variable in the condition, the double wrap, and the dead assignments in the panic path. It also adds a focused test and corrects the inverted assertions in `user_repository_test.go`.

### What I checked
- **`errors.Is` / `errors.As` matching is preserved.** The callback error stays wrapped with `%w` on every path, so downstream matching does not change.
- **No caller depends on the old message.** The only `rollback:` match in non-test code is an unrelated comment in `core/serviceuser/service.go`.
- **Panic path behavior is identical.** The deferred func re-panics, so the named `err` return is never observed by the caller. Dropping those assignments changes nothing.
- **Test coverage is strong.** Commit, begin failure, rollback success and failure with exact messages, commit failure, and panic are all covered. The exact-message asserts guard against the `%!s(<nil>)` regression directly.

### Verdict
Looks good to merge. There is one minor, non-blocking suggestion inline about making the rollback error matchable with `errors.Is`. The other inline notes are just clarifications for the review. Nice catch on the inverted `errors.Unwrap(err) == tc.Err` asserts too.

internal/store/postgres/user_repository_test.go
```diff
@@ -366,8 +366,8 @@ func (s *UserRepositoryTestSuite) TestUpdateByEmail() {
 		s.Run(tc.Description, func() {
 			got, err := s.repository.UpdateByEmail(s.ctx, tc.UserToUpdate)
 			if tc.Err != nil && tc.Err.Error() != "" {
-				if errors.Unwrap(err) == tc.Err {
-					s.T().Fatalf("got error %s, expected was %s", err.Error(), tc.Err)
+				if !errors.Is(err, tc.Err) {
```
rohilsurana, line 369: Good fix. The old `if errors.Unwrap(err) == tc.Err { Fatalf }` was inverted: it failed the test when the error matched. The double-wrap bug in `WithTxn` was masking it because `Unwrap` returned the intermediate wrapper, not `tc.Err`. `!errors.Is(err, tc.Err)` is the correct assertion and also handles multi-level wrapping.

pkg/db/db.go
AmanGIT07: Applied in 7cdc2157 — both errors are wrapped with %w now, message text unchanged. Also added an errors.Is assert for the rollback error in the rollback-failure test.

rohilsurana (approved):

PR #1857, opened 2026-08-05
feat(deleter): clear or report org delete blockers up front

### LINKED ISSUE #1837: Organization delete: check all blockers up front and tell the user what to fix

whoAbhishekSah:
Closes https://github.com/raystack/frontier/issues/1837. Based on #1865; the forfeit email (#1880) and the check RPC (#1893/#1894) stack on top.

`DeleteOrganization` now checks everything that blocks the delete before touching any data, and returns every reason together as one `failed_precondition` response with a `PreconditionFailure` detail (one violation per blocker). The proto ack field from the issue text was dropped (https://github.com/raystack/proton/pull/497 closed) — unused tokens never block.

What blocks the delete:

- **Active subscription on a paid plan** (`ACTIVE_SUBSCRIPTION`): any running subscription whose plan has a priced product. The caller downgrades it to the standard plan through the normal plan-change flow, then retries. Plans resolve lazily — only when a running subscription references one — so a broken plan reference can only affect orgs that actually hold such a subscription.
- **Invoice that still asks for money** (`UNPAID_INVOICE`): open and uncollectible invoices the caller can pay, and **drafts** — the provider finalizes those shortly, and deleting inside that window would silently lose the charge, so they block until payable. "Asks for money" is judged by the provider's `amount_remaining`, so a negative-total credit note or an invoice fully covered by the customer's credit balance never blocks.
- **A running subscription on a plan that no longer resolves** blocks with its own message (cancel the subscription, then retry) instead of failing the delete forever.
- **Negative token balance** (`NEGATIVE_TOKEN_BALANCE`): the account owes tokens (overdraft). Support settles it.

What does not block:

- **A running subscription on a free plan**: once nothing else blocks, the delete cancels it itself (immediate, unbilled usage invoiced on the spot), then re-checks the invoices. Every account is judged again before anything is canceled, so a paid subscription created mid-delete blocks without costing another account its subscription first, and a subscription whose provider copy is already gone is skipped rather than failing the delete forever. Paid subscriptions are never canceled by the delete, so a blocked delete never costs the caller a plan they pay for.
- **Unused tokens**: the delete forfeits them (the client confirms with the user first) and writes the amount to an `app.billing.tokens.forfeited` audit record. #1880 adds the owner email.

Other changes:

- Invoices are judged straight from the billing provider through a new `invoice.ListPayableOnProvider` — three status-filtered pages, no expansion, no local writes — instead of a full `SyncWithProvider`, keeping the delete's latency independent of the org's invoice history. An invoice the sync has not stored yet is reported by its provider reference.
- The billing accounts are listed once per delete and passed through the whole flow, and plan lookups share one per-delete cache.
- Deleting an org that is already gone returns `not_found` before any checks run. Disabled orgs stay deletable.
- New invoice state constant `UncollectibleState`; first use of `errdetails` in the repo (`google.golang.org/genproto/googleapis/rpc` becomes a direct dependency).

### DIFF
--- a/billing/invoice/service.go
+++ b/billing/invoice/service.go
@@ -244,6 +244,59 @@ func (s *Service) isCreditOverdraftDayOfInvoice() bool {
 	return time.Now().UTC().Day() == s.creditOverdraftInvoiceDay
 }
 
+// ListPayableOnProvider returns the customer's invoices that still ask for
+// money — open, uncollectible, or being prepared as drafts — read straight
+// from the billing provider so the answer is current. Zero-amount invoices
+// are skipped. Where a local row exists for the invoice it keeps its local
+// id; an invoice the sync has not seen yet is returned with an empty id and
+// only its provider reference. Unlike SyncWithProvider this touches no local
+// rows, expands nothing, and reads one status-filtered listing per payable
+// state — a customer with very many payable invoices pages through more
+// requests, but the cost scales with what is still owed, not with the whole
+// invoice history.
+func (s *Service) ListPayableOnProvider(ctx context.Context, customr customer.Customer) ([]Invoice, error) {
+	localInvoices, err := s.repository.List(ctx, Filter{
+		CustomerID: customr.ID,
+	})
+	if err != nil {
+		return nil, err
+	}
+	localByProviderID := make(map[string]Invoice, len(localInvoices))
+	for _, inv := range localInvoices {
+		localByProviderID[inv.ProviderID] = inv
+	}
+
+	var payable []Invoice
+	for _, status := range PayableStates {
+		stripeInvoices := s.stripeClient.Invoices.List(&stripe.InvoiceListParams{
+			Customer: stripe.String(customr.ProviderID),
+			Status:   stripe.String(string(status)),
+			ListParams: stripe.ListParams{
+				Context: ctx,
+			},
+		})
+		for stripeInvoices.Next() {
+			stripeInvoice := stripeInvoices.Invoice()
+			// AmountRemaining is what the customer still owes; Total is not:
+			// a negative total is a credit owed to the customer, and an open
+			// invoice fully covered by their credit balance has a positive
+			// total with nothing due. Neither asks for money.
+			if stripeInvoice.AmountRemaining <= 0 {
+				continue
+			}
+			inv := stripeInvoiceToInvoice(customr.ID, stripeInvoice)
+			if local, ok := localByProviderID[stripeInvoice.ID]; ok {
+				inv.ID = local.ID
+			}
+			payable = append(payable, inv)
+		}
+		if err := stripeInvoices.Err(); err != nil {
+			return nil, fmt.Errorf("failed to list %s invoices: %w", status, billingerrors.TranslateStripeError(err))
+		}
+	}
+	return payable, nil
+}
+
 func (s *Service) SyncWithProvider(ctx context.Context, customr customer.Customer) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
--- a/core/deleter/deleter.go
+++ b/core/deleter/deleter.go
@@ -1,7 +1,38 @@
 package deleter
 
-import "fmt"
+import "strings"
 
-var (
-	ErrDeleteNotAllowed = fmt.Errorf("deletion not allowed for billed accounts")
+// Reasons an organization delete can be blocked. The API error carries
+// them as violation types, so a client can tell the reasons apart
+// without reading the message text.
+const (
+	BlockerActiveSubscription   = "ACTIVE_SUBSCRIPTION"
+	BlockerUnpaidInvoice        = "UNPAID_INVOICE"
+	BlockerNegativeTokenBalance = "NEGATIVE_TOKEN_BALANCE"
 )
+
+// Blocker is one reason an organization cannot be deleted right now.
+type Blocker struct {
+	// Type is one of the Blocker* constants.
+	Type string
+	// Subject is the id of the entity behind the reason.
+	Subject string
+	// Message says what blocks the delete and how to clear it.
+	Message string
+}
+
+// BlockedError carries every blocker the up-front check found, so the
+// caller gets one checklist instead of discovering blockers one retry at
+// a time.
+type BlockedError struct {
+	OrgID    string
+	Blockers []Blocker
+}
+
+func (e *BlockedError) Error() string {
+	msgs := make([]string, 0, len(e.Blockers))
+	for _, b := range e.Blockers {
+		msgs = append(msgs, b.Message)
+	}
+	return "organization cannot be deleted yet: " + strings.Join(msgs, "; ")
+}
--- a/core/deleter/service.go
+++ b/core/deleter/service.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"strconv"
 
 	"github.com/raystack/frontier/core/audit"
 
@@ -15,6 +16,10 @@ import (
 	"github.com/raystack/frontier/billing/invoice"
 
 	"github.com/raystack/frontier/billing/customer"
+
+	"github.com/raystack/frontier/billing/plan"
+
+	"github.com/raystack/frontier/billing/subscription"
 
 	"github.com/raystack/frontier/core/organization"
 
@@ -32,10 +37,6 @@ import (
 	"github.com/raystack/frontier/core/project"
 	"github.com/raystack/frontier/core/resource"
 	"github.com/raystack/frontier/core/serviceuser"
-)
-
-const (
-	DisableDeleteIfBilled = true
 )
 
 type ProjectService interface {
@@ -98,11 +99,14 @@ type CustomerService interface {
 }
 
 type SubscriptionService interface {
+	List(ctx context.Context, filter subscription.Filter) ([]subscription.Subscription, error)
+	Cancel(ctx context.Context, id string, immediate bool) (subscription.Subscription, error)
 	DeleteByCustomer(ctx context.Context, customr customer.Customer) error
 }
 
 type InvoiceService interface {
 	List(ctx context.Context, flt invoice.Filter) ([]invoice.Invoice, error)
+	ListPayableOnProvider(ctx context.Context, customr customer.Customer) ([]invoice.Invoice, error)
 	DeleteByCustomer(ctx context.Context, customr customer.Customer) error
 }
 
@@ -112,11 +116,16 @@ type CheckoutService interface {
 }
 
 type CreditService interface {
+	GetBalance(ctx context.Context, accountID string) (int64, error)
 	DeleteByAccountID(ctx context.Context, accountID string) error
 }
 
 type KycService interface {
 	DeleteKyc(ctx context.Context, orgID string) error
+}
+
+type PlanService interface {
+	GetByID(ctx context.Context, id string) (plan.Plan, error)
 }
 
 type Service struct {
@@ -137,6 +146,7 @@ type Service struct {
 	checkoutService    CheckoutService
 	creditService      CreditService
 	kycService         KycService
+	planService        PlanService
 }
 
 func NewCascadeDeleter(orgService OrganizationService, projService ProjectService,
@@ -148,7 +158,8 @@ func NewCascadeDeleter(orgService OrganizationService, projService ProjectServic
 	serviceUserService ServiceUserService,
 	customerService CustomerService, subService SubscriptionService,
 	invoiceService InvoiceService, checkoutService CheckoutService,
-	creditService CreditService, kycService KycService) *Service {
+	creditService CreditService, kycService KycService,
+	planService PlanService) *Service {
 	return &Service{
 		projService:        projService,
 		orgService:         orgService,
@@ -167,6 +178,7 @@ func NewCascadeDeleter(orgService OrganizationService, projService ProjectServic
 		checkoutService:    checkoutService,
 		creditService:      creditService,
 		kycService:         kycService,
+		planService:        planService,
 	}
 }
 
@@ -215,15 +227,32 @@ func (d Service) DeleteGroup(ctx context.Context, id string) error {
 // likely to fail. Identity (projects, policies, and so on) goes after, with
 // org policies (the org owners) near the end. This way a failure at any step
 // leaves the org owned and the delete can simply be run again. Every step
-// treats already-deleted data as success for the same reason.
+// treats already-deleted data as success for the same reason. That applies
+// to data inside a half-deleted org; an org whose row is already fully gone
+// answers not found instead, so a caller can tell a finished delete from a
+// repeatable one (and a mistyped org id from a success).
 func (d Service) DeleteOrganization(ctx context.Context, id string) error {
-	// check if delete is allowed
-	if err := d.canDelete(ctx, id); err != nil {
-		return fmt.Errorf("%s: %w", err.Error(), ErrDeleteNotAllowed)
+	// an org that is already gone has nothing left to check or tear down;
+	// disabled orgs stay deletable
+	if _, err := d.orgService.Get(ctx, id); err != nil && !errors.Is(err, organization.ErrDisabled) {
+		return err
+	}
+
+	customers, err := d.customerService.List(ctx, customer.Filter{
+		OrgID: id,
+	})
+	if err != nil {
+		return err
+	}
+
+	// clear what we can and collect what still blocks the delete, before
+	// touching any data
+	if err := d.ensureDeletable(ctx, id, customers); err != nil {
+		return err
 	}
 
 	// delete all billing accounts
-	if err := d.DeleteCustomers(ctx, id); err != nil {
+	if err := d.deleteCustomers(ctx, id, customers); err != nil {
 		return err
 	}
 
@@ -316,6 +345,8 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
 	return nil
 }
 
+// DeleteCustomers lists the org's billing accounts itself; DeleteOrganization
+// goes through deleteCustomers with the accounts it already listed.
 func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 	customers, err := d.customerService.List(ctx, customer.Filter{
 		OrgID: id,
@@ -323,6 +354,10 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 	if err != nil {
 		return err
 	}
+	return d.deleteCustomers(ctx, id, customers)
+}
+
+func (d Service) deleteCustomers(ctx context.Context, id string, customers []customer.Customer) error {
 	for _, c := range customers {
 		// cancels active subscriptions on the billing provider and removes local records
 		if err := d.subService.DeleteByCustomer(ctx, c); err != nil {
@@ -362,6 +397,22 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 				Type: "billing_checkout",
 			}, attrs); err != nil {
 				slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.BillingCheckoutDeletedEvent, "checkout_id", ch.ID)
+			}
+		}
+		// tokens still on the account are forfeited by this delete, so
+		// record the amount before the transactions are removed
+		balance, err := d.creditService.GetBalance(ctx, c.ID)
+		if err != nil {
+			return fmt.Errorf("failed to delete org while checking balance of billing account[%s]: %w", c.ID, err)
+		}
+		if balance > 0 {
+			if err := auditLogger.LogWithAttrs(audit.BillingTokensForfeitedEvent, audit.Target{
+				ID:   c.ID,
+				Type: "billing_account",
+			}, map[string]string{
+				"amount": strconv.FormatInt(balance, 10),
+			}); err != nil {
+				slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.BillingTokensForfeitedEvent, "customer_id", c.ID)
 			}
 		}
 		if err := d.creditService.DeleteByAccountID(ctx, c.ID); err != nil {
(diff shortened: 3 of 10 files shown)


core/deleter/deleter.go
```diff
@@ -1,7 +1,41 @@
 package deleter
 
-import "fmt"
+import "strings"
 
-var (
-	ErrDeleteNotAllowed = fmt.Errorf("deletion not allowed for billed accounts")
+// Blocker types returned by the delete's up-front check. They are
```
whoAbhishekSah: Org deletes up-front check. Remove machine-readable word. clients can branch on them is weird english. Keep it in simple language.

core/deleter/deleter.go
```diff
@@ -1,7 +1,41 @@
...
+// machine-readable and end up as PreconditionFailure violation types on
+// the API error, so clients can branch on them.
+const (
+	BlockerActiveSubscription   = "ACTIVE_SUBSCRIPTION"
+	BlockerUnpaidInvoice        = "UNPAID_INVOICE"
+	BlockerNegativeTokenBalance = "NEGATIVE_TOKEN_BALANCE"
 )
+
+// Blocker is one reason an organization cannot be deleted right now. The
```
whoAbhishekSah: Blocker struct is generic. so may be we can write the comment in that way

core/deleter/service.go
```diff
@@ -137,6 +147,11 @@ type Service struct {
 	checkoutService    CheckoutService
 	creditService      CreditService
 	kycService         KycService
+	planService        PlanService
+	// defaultPlan names the plan a subscription may still be on when the org
+	// is deleted; subscriptions on any other plan block the delete until the
+	// caller downgrades them
+	defaultPlan string
```
whoAbhishekSah: how does the billing services decide what is the default plan? does it come from billing reconciler ? if yes what is the format of the config

whoAbhishekSah: is this plan id or name or title ?

core/deleter/service.go
```diff
@@ -364,6 +389,22 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 				slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.BillingCheckoutDeletedEvent, "checkout_id", ch.ID)
 			}
 		}
+		// tokens still on the account are forfeited by this delete, so
+		// record the amount before the transactions are removed
+		balance, err := d.creditService.GetBalance(ctx, c.ID)
+		if err != nil {
+			return fmt.Errorf("failed to delete org while checking balance of billing account[%s]: %w", c.ID, err)
+		}
+		if balance > 0 {
```
whoAbhishekSah, line 408: Can balance be less than 0 ? What about postpaid cases? credit_min ?

core/deleter/service.go
```diff
@@ -412,23 +453,157 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+				Subject: c.ID,
+				Message: fmt.Sprintf("billing account[%s] owes %d tokens: contact support to settle the balance, then retry the delete", c.ID, -balance),
+			})
+		}
+	}
+	if len(blockers) > 0 {
+		return &BlockedError{OrgID: id, Blockers: blockers}
+	}
+
+	// nothing blocks the delete; the only subscriptions still running are on
```
whoAbhishekSah: write it in simple language... "nothing blocks" is not correct english

core/deleter/service_test.go
```diff
@@ -188,20 +192,120 @@ func TestDeleteOrganization(t *testing.T) {
...
+
+		err := m.build().DeleteOrganization(context.Background(), "org-1", false)
+
+		var blocked *deleter.BlockedError
+		assert.ErrorAs(t, err, &blocked)
+		assert.Len(t, blocked.Blockers, 1)
+		assert.Equal(t, deleter.BlockerNegativeTokenBalance, blocked.Blockers[0].Type)
+		// the ack flag cannot bypass a debt
+		err = m.build().DeleteOrganization(context.Background(), "org-1", true)
+		assert.ErrorAs(t, err, &blocked)
```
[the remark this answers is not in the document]
whoAbhishekSah: The design dropped the acknowledgement flag entirely — unused tokens no longer block the delete and DeleteOrganization has no ack parameter anymore, so this test no longer exists. The debt case now asserts the blocker type and message directly.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+		subs, err := d.subService.List(ctx, subscription.Filter{CustomerID: c.ID})
+		if err != nil {
+			return fmt.Errorf("failed to check subscriptions for billing account[%s]: %w", c.ID, err)
+		}
+		canceled := false
+		for _, sub := range subs {
+			if !sub.IsActive() {
+				continue
+			}
+			if _, err := d.subService.Cancel(ctx, sub.ID, true); err != nil {
```
whoAbhishekSah: Cancel here treats every error as fatal, but `subscription.Service.DeleteByCustomer` — one step later in this same flow — tolerates `ErrSubscriptionOnProviderNotFound`. A subscription row that is locally active while its Stripe copy is gone (provider drift) makes this return an error on every attempt, so the org becomes permanently undeletable. Skipping the cancel on `ErrSubscriptionOnProviderNotFound` matches what the teardown itself already does.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+
+	var blockers []Blocker
+	for _, c := range customers {
+		if !c.IsOffline() {
+			subs, err := d.subService.List(ctx, subscription.Filter{CustomerID: c.ID})
+			if err != nil {
+				return fmt.Errorf("failed to check subscriptions for billing account[%s]: %w", c.ID, err)
+			}
+			for _, sub := range subs {
+				if sub.IsActive() && sub.PlanID != standardPlanID {
```
whoAbhishekSah: The paid-plan test is identity with the one configured plan id. Two consequences:

1. With `default_plan` unset (it has no default), `standardPlanID` is `""` and every active sub blocks with "downgrade to the standard plan" — advice that can never unblock, since after any downgrade the sub is still active with a non-empty plan id.
2. A free plan that is not the configured default also blocks, though there is nothing to bill.

`plan.IsFree()` looks like the check this actually wants — block on any active sub that costs money, cancel the free ones.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+// configured default plan every active subscription blocks the delete. The
+// lookup is skipped when no billing account talks to a provider.
+func (d Service) standardPlanID(ctx context.Context, customers []customer.Customer) (string, error) {
+	if d.defaultPlan == "" {
+		return "", nil
+	}
+	if !slices.ContainsFunc(customers, func(c customer.Customer) bool { return !c.IsOffline() }) {
+		return "", nil
+	}
+	standardPlan, err := d.planService.GetByID(ctx, d.defaultPlan)
```
whoAbhishekSah: This lookup runs before we know whether the org has any subscriptions at all, and a failure aborts the delete as an internal error. If `default_plan` names a plan that was never seeded (or was renamed), every delete of every org with an online billing account 500s — including orgs with zero subscriptions that were deletable before this change. Resolving the plan lazily, only once an active sub is actually found, contains the blast radius of a config/seed mismatch.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+		if c.IsOffline() {
+			continue
+		}
+		subs, err := d.subService.List(ctx, subscription.Filter{CustomerID: c.ID})
+		if err != nil {
+			return fmt.Errorf("failed to check subscriptions for billing account[%s]: %w", c.ID, err)
+		}
+		canceled := false
+		for _, sub := range subs {
+			if !sub.IsActive() {
```
whoAbhishekSah, line 546: Subs in provider states outside `IsActive()` — `past_due`, `unpaid`, `paused` — fall through everywhere: they raise no ACTIVE_SUBSCRIPTION blocker above, get skipped by this cancel loop, and `DeleteByCustomer` skips its provider cancel for them too. A paid-plan org that is past_due can delete without ever downgrading or settling, relying only on Stripe's customer-delete to kill the sub.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+	if err := d.invoiceService.SyncWithProvider(ctx, c); err != nil {
+		return nil, fmt.Errorf("failed to sync invoices for billing account[%s]: %w", c.ID, err)
+	}
+	invoices, err := d.invoiceService.List(ctx, invoice.Filter{CustomerID: c.ID, NonZeroOnly: true})
+	if err != nil {
+		return nil, fmt.Errorf("failed to check invoices for billing account[%s]: %w", c.ID, err)
+	}
+	var blockers []Blocker
+	for _, inv := range invoices {
+		if inv.State == invoice.OpenState || inv.State == invoice.UncollectibleState {
```
whoAbhishekSah: Draft invoices with amount > 0 no longer block (the old code blocked on any invoice), and the Stripe customer deletion later in the teardown discards drafts. Stripe creates renewal invoices as drafts and auto-finalizes them about an hour later — a delete inside that window silently loses the charge. Same window applies to the `InvoiceNow` cancellation invoice if this re-check catches it while still draft.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+			if !sub.IsActive() {
+				continue
+			}
+			if _, err := d.subService.Cancel(ctx, sub.ID, true); err != nil {
+				return fmt.Errorf("failed to cancel subscription[%s] of billing account[%s]: %w", sub.ID, c.ID, err)
 			}
+			canceled = true
 		}
+		if canceled {
+			bs, err := d.invoiceBlockers(ctx, c)
```
whoAbhishekSah: If the cancel's own final invoice shows up open here (auto-charge settles async, so with unbilled metered usage there is a window), the delete returns UNPAID_INVOICE — but the subscription is already gone and nothing re-creates it if the caller abandons the delete. That contradicts the comment above ("a delete that stays blocked does not cost the caller their subscription") and the message tells them to pay via the hosted page, which is not theirs to do on an auto-charging invoice.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+
+// invoiceBlockers returns a blocker for every invoice of the account the
+// caller can still pay; paid, void, and draft invoices don't block. The
+// billing provider keeps its own permanent copy of every invoice, so
+// deleting our rows later loses nothing.
+func (d Service) invoiceBlockers(ctx context.Context, c customer.Customer) ([]Blocker, error) {
+	// the local invoice rows sync from the provider on a timer, so pull
+	// them fresh first: a just-paid invoice must not block, a just-created
+	// one must
+	if err := d.invoiceService.SyncWithProvider(ctx, c); err != nil {
```
whoAbhishekSah: This inline sync makes the delete RPC's latency proportional to the org's full invoice history: `SyncWithProvider` paginates every Stripe invoice with line expansion, sleeps up to 1s per invoice, and holds the mutex shared with the background sync loop — and it can run twice per account in one delete. It also makes Stripe availability a hard dependency of the delete path. A status-filtered query for open/uncollectible invoices only (no expansion, no sleeps) gives the same answer much cheaper.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
-				return fmt.Errorf("cannot delete organization with billing account[%s]", c.ID)
+		if c.IsOffline() {
+			continue
+		}
+		subs, err := d.subService.List(ctx, subscription.Filter{CustomerID: c.ID})
+		if err != nil {
+			return fmt.Errorf("failed to check subscriptions for billing account[%s]: %w", c.ID, err)
+		}
+		canceled := false
+		for _, sub := range subs {
```
whoAbhishekSah, line 545: This loop cancels every `IsActive` sub without re-checking the plan against `standardPlanID`. A paid-plan checkout that completes between the blocker pass's List and this List gets `Cancel(immediate=true)` instead of an ACTIVE_SUBSCRIPTION blocker — silently and irreversibly. Re-checking the plan here (or reusing the first List's result) closes the window.

core/deleter/service.go
```diff
@@ -364,6 +391,22 @@ func (d Service) DeleteCustomers(ctx context.Context, id string) error {
 				slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.BillingCheckoutDeletedEvent, "checkout_id", ch.ID)
 			}
 		}
+		// tokens still on the account are forfeited by this delete, so
+		// record the amount before the transactions are removed
+		balance, err := d.creditService.GetBalance(ctx, c.ID)
```
whoAbhishekSah, line 404: A `GetBalance` failure here hard-aborts the teardown midway (subscription and invoice rows for this customer are already gone), yet the value only feeds the audit attribute — and the audit write's own failure three lines down is deliberately warn-and-continue. `ensureDeletable` read the same balance moments earlier in this call; threading it through avoids both the extra read and the inverted severity.

core/deleter/service.go
```diff
@@ -412,23 +455,162 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
 	for _, c := range customers {
-		if invoices, err := d.invoiceService.List(ctx, invoice.Filter{CustomerID: c.ID}); err != nil {
-			return fmt.Errorf("failed to check invoices for billing account[%s]: %w", c.ID, err)
-		} else if len(invoices) > 0 {
-			if DisableDeleteIfBilled {
-				return fmt.Errorf("cannot delete organization with billing account[%s]", c.ID)
+		if c.IsOffline() {
+			continue
+		}
+		subs, err := d.subService.List(ctx, subscription.Filter{CustomerID: c.ID})
```
whoAbhishekSah, line 541: Duplicated I/O within one delete: this List already ran for the same customers in the blocker pass above, and `GetBalance` — a serializable full-history aggregate the repo comment itself flags as slow — runs again in `DeleteCustomers`. Nothing mutates in between; collecting the active sub ids and balances once in the first pass and reusing them drops the second round of reads (and keeps the audit amount consistent with what was checked).

whoAbhishekSah (commented):
Reviewed the deleter changes; 10 inline notes below. The two biggest for me: provider-drifted subs making an org permanently undeletable, and draft invoices slipping through the blocker check and getting discarded by the customer delete. None of these are fixed by the later commits in the stack.

core/deleter/service.go
whoAbhishekSah: No code change here, reasoning: the money in these states is guarded by the invoice blockers, not the subscription blocker. past_due and unpaid mean a finalized invoice went unpaid — that invoice is open (or uncollectible after write-off), and both block; drafts block now too. A paused subscription bills nothing while paused. At teardown, the Stripe customer deletion cancels whatever subscription state remains, so nothing keeps charging after the org is gone. The provider-cancel skip for non-active states inside DeleteByCustomer is pre-existing teardown behavior from the earlier cascade work, not something this PR changes.

core/deleter/service.go
whoAbhishekSah: Partly fixed, partly accepted: the window is real but now only free-plan subscriptions can be in it — paid plans are never canceled by the delete, so the corrected comment says a blocked delete never costs a plan the caller pays for. If the free plan's final usage invoice shows up open here, the delete blocks with the sub already canceled; the caller loses a free subscription on an org they were deleting, which seems acceptable. On the message: a finalized auto-charge invoice can also be paid from its hosted page, and Stripe retries the charge itself, so the advice still works — retrying the delete after the charge settles also passes.

core/deleter/service.go
whoAbhishekSah: Replaced the inline sync with a new invoice.ListPayableOnProvider: three status-filtered pages (open, uncollectible, draft), no line expansion, no sleeps, no mutex, no local writes — delete latency no longer scales with invoice history. One semantic note: an invoice the background sync has not stored yet has no local id, so its blocker carries the provider reference instead.

core/deleter/service.go
whoAbhishekSah: Fixed in the email PR commit on this stack, where the delete already collects per-account balances up front for the forfeit notice: those numbers now thread into the teardown (deleteCustomers), so the audit uses what the pre-check read and no balance read can abort mid-teardown. The exported DeleteCustomers keeps reading for itself when called directly.

core/deleter/service.go
whoAbhishekSah: GetBalance duplication is gone — the pre-collected amounts thread into the teardown (see the email PR commit). The second subscription List stays on purpose: reusing the first pass's list would reopen the race in the other thread — a paid subscription created between the passes must be seen and blocked, not carried over as absent. The plan lookups it triggers are cached per call.

billing/invoice/service.go
```diff
@@ -244,6 +244,53 @@ func (s *Service) isCreditOverdraftDayOfInvoice() bool {
...
+		stripeInvoices := s.stripeClient.Invoices.List(&stripe.InvoiceListParams{
+			Customer: stripe.String(customr.ProviderID),
+			Status:   stripe.String(string(status)),
+			ListParams: stripe.ListParams{
+				Context: ctx,
+			},
+		})
+		for stripeInvoices.Next() {
+			stripeInvoice := stripeInvoices.Invoice()
+			if stripeInvoice.Total == 0 {
```
rohilsurana: This skips an invoice only when its Total is exactly 0. Two normal Stripe cases still get through and turn into fake blockers: a negative-total invoice (a proration credit the customer is owed), and an open invoice already fully covered by the customer's credit balance (Total is positive but nothing is due). Both leave the org stuck with an unpaid-invoice message the caller cannot act on. The field that means "still owes money" is AmountRemaining (or AmountDue), not Total.

rohilsurana (commented):
Stack-aware review (base org-delete-preflight-mocks). A few things to flag. The invoice one is the most important; the rest are smaller. I dropped two earlier notes after reading the stack: the empty-PlanID-is-paid case is a documented choice, and the mid-teardown balance read is reworked in #1880.

core/deleter/service.go
```diff
@@ -412,23 +447,201 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+// the length of one delete. A subscription without a plan counts as paid:
+// blocking it is recoverable, canceling it is not.
+func (d Service) isPaidPlan(ctx context.Context, planID string, cache map[string]bool) (bool, error) {
+	if planID == "" {
+		return true, nil
+	}
+	if paid, ok := cache[planID]; ok {
+		return paid, nil
+	}
+	pln, err := d.planService.GetByID(ctx, planID)
```
rohilsurana, line 652: If a running subscription points at a plan that no longer resolves (a hard-deleted plan, or GetByID skipping soft-deleted ones), this returns the error and the whole delete aborts on every attempt. The caller cannot clear a dangling plan reference, so the org can never be deleted. It may be safer to treat an unresolvable plan as a clear, actionable blocker than as a hard error.

core/deleter/service.go
```diff
@@ -412,23 +447,201 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+				Subject: c.ID,
+				Message: fmt.Sprintf("billing account[%s] owes %d tokens: contact support to settle the balance, then retry the delete", c.ID, -balance),
+			})
+		}
+	}
+	if len(blockers) > 0 {
+		return &BlockedError{OrgID: id, Blockers: blockers}
+	}
+
+	// no blockers were found, so the delete will happen. Only free-plan
```
rohilsurana, line 526: This second pass cancels each free-plan subscription one account at a time. If a paid subscription has shown up on a later account since the first pass, this returns a BlockedError after an earlier account's free subscription was already canceled. That cancel is immediate and bills usage, so the caller ends up with a canceled subscription and a final invoice even though the delete did not go through. It is a narrow race, but worth a note or a re-check before the cancel.

core/deleter/service_test.go
```diff
@@ -63,13 +68,26 @@ func newMocks(t *testing.T) deleterMocks {
 		checkoutSvc: mocks.NewCheckoutService(t),
 		creditSvc:   mocks.NewCreditService(t),
 		kycSvc:      mocks.NewKycService(t),
+		planSvc:     mocks.NewPlanService(t),
 	}
+	// the standard plan resolves on any org with provider-backed billing;
```
rohilsurana: These first two comment lines describe an older "stub once for every test" scheme and contradict the next comment, which says plans resolve lazily by id. Drop the stale lines so the next reader is not misled.

whoAbhishekSah:
Round 3 sandbox results — retesting the behavior changes from the review (stack: #1857 → #1880 → #1894, tested at the tip `org-delete-check-api` @ 6b9468e2). Each case used a fresh org, a real Stripe test customer, and a normal org-owner session.

**1. Credit-note invoice no longer blocks.** Draft invoice with total −500 (amount_remaining 0). `CheckOrganizationDelete` returned `can_delete: true` and the delete succeeded. The old total-based check would have blocked it.

**2. Credit-balance-covered invoice no longer blocks.** Customer with a 1000 credit balance and a draft invoice of total 800 (amount_due 0 on Stripe). Check passed and the delete succeeded.

**3. Dangling plan yields a blocker, not an error.** Subscription row pointing at a plan id that no longer exists (a real Stripe sub behind it). Check and delete both returned the `failed_precondition` blocker: "subscription[…] is active on a plan that no longer exists: cancel the subscription, then retry the delete". The advice works: `CancelSubscription` succeeded despite the dangling plan (canceled on Stripe too), and the retry deleted the org.

**4. Cancel pass is two-phase.** One org, two billing accounts: A with an active free-plan subscription, B with an open unpaid invoice. The delete blocked on B's invoice and A's subscription stayed active on Stripe and in the local rows. After voiding the invoice, the retry deleted the org and canceled A's subscription itself. Both Stripe customers were removed.

**5. Retry recovers the forfeit notice from audit.** Org with zero live balance plus an `app.billing.tokens.forfeited` audit row (amount 750, purchased 200), the state a failed first attempt leaves behind. The retry delete succeeded and the owner received the notice rebuilt from the audit row: "Your organization Round3 r3-retry was deleted by delowner3@raystack.org with 750 unused tokens remaining, of which 200 came from purchases."

All test orgs, billing accounts, and Stripe customers were cleaned up afterwards.

billing/invoice/service.go
```diff
@@ -244,6 +244,57 @@ func (s *Service) isCreditOverdraftDayOfInvoice() bool {
...
+				Context: ctx,
+			},
+		})
+		for stripeInvoices.Next() {
+			stripeInvoice := stripeInvoices.Invoice()
+			// AmountRemaining is what the customer still owes; Total is not:
+			// a negative total is a credit owed to the customer, and an open
+			// invoice fully covered by their credit balance has a positive
+			// total with nothing due. Neither asks for money.
+			if stripeInvoice.AmountRemaining <= 0 {
```
rohilsurana, line 284: The filter now skips any invoice with AmountRemaining <= 0 for all three statuses, including draft. That is right for open and uncollectible, but a draft can have amount_remaining = 0 while it is still being finalized, with the real charge sitting in its total or line items. The invoiceBlockers docstring says drafts must block precisely so a pending charge is not silently lost, so this may now drop the very drafts it means to keep. Worth confirming what Stripe returns for amount_remaining on an unfinalized draft. If it is not reliable there, keep drafts blocking on total > 0 and use AmountRemaining only for open and uncollectible.

rohilsurana (commented):
Re-review after the fix commits. The earlier points are handled well: AmountRemaining for the filter, the unresolvablePlanBlocker for dangling plans, the judge-before-cancel restructure, and the test comment. A few new things the fixes surfaced, plus two small doc nits.

core/deleter/service.go
```diff
@@ -412,23 +460,222 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
...
+		Subject: sub.ID,
+		Message: fmt.Sprintf("subscription[%s] is %s on a paid plan: downgrade to the standard plan, then retry the delete", sub.ID, sub.State),
+	}
+}
+
+// isPaidPlan reports whether the plan charges money, caching each plan for
+// the length of one delete. A subscription without a plan counts as paid:
+// blocking it is recoverable, canceling it is not.
+func (d Service) isPaidPlan(ctx context.Context, planID string, cache map[string]bool) (bool, error) {
+	if planID == "" {
```
rohilsurana: An empty PlanID returns paid = true here, so subscriptionBlockers emits a paidSubscriptionBlocker whose message tells the user to downgrade to the standard plan. There is no plan to downgrade. Now that unresolvablePlanBlocker exists with a clearer message (cancel the subscription), an empty PlanID reads better sent down that path so the guidance is actually doable.

core/deleter/service.go
```diff
@@ -217,13 +229,27 @@ func (d Service) DeleteGroup(ctx context.Context, id string) error {
 // leaves the org owned and the delete can simply be run again. Every step
 // treats already-deleted data as success for the same reason.
 func (d Service) DeleteOrganization(ctx context.Context, id string) error {
-	// check if delete is allowed
-	if err := d.canDelete(ctx, id); err != nil {
-		return fmt.Errorf("%s: %w", err.Error(), ErrDeleteNotAllowed)
+	// an org that is already gone has nothing left to check or tear down;
+	// disabled orgs stay deletable
+	if _, err := d.orgService.Get(ctx, id); err != nil && !errors.Is(err, organization.ErrDisabled) {
```
rohilsurana, line 237: This preflight returns the error for any Get failure other than ErrDisabled. For an org whose row was already fully deleted (a retry after a delete that succeeded but whose response was lost), Get returns ErrNotExist and the whole call now fails with NotFound. The docstring just above says every step treats already-deleted data as success and the delete can simply be run again, so this quietly breaks that idempotency for the already-gone case. Treating ErrNotExist here as success would keep the retry idempotent.

billing/invoice/service.go
```diff
@@ -244,6 +244,57 @@ func (s *Service) isCreditOverdraftDayOfInvoice() bool {
 	return time.Now().UTC().Day() == s.creditOverdraftInvoiceDay
 }
 
+// ListPayableOnProvider returns the customer's invoices that still ask for
+// money — open, uncollectible, or being prepared as drafts — read straight
+// from the billing provider so the answer is current. Zero-amount invoices
+// are skipped. Where a local row exists for the invoice it keeps its local
+// id; an invoice the sync has not seen yet is returned with an empty id and
+// only its provider reference. Unlike SyncWithProvider this touches no local
+// rows and reads only three small filtered pages, so it is cheap enough for
```
rohilsurana: Small doc point: this says it reads only three small filtered pages and is cheap enough for a request path, but the Next() loop below auto-paginates through every page for each of the three statuses. For a customer with many open invoices that is more than three requests. Worth softening the comment so a later caller does not assume a bounded cost.

core/deleter/service.go
rohilsurana: This preflight returns the error for any Get failure other than ErrDisabled. For an org whose row was already fully deleted (a retry after a delete that succeeded but whose response was lost), Get returns ErrNotExist and the whole call now fails with NotFound. The docstring just above says every step treats already-deleted data as success and the delete can simply be run again, so this quietly breaks that idempotency for the already-gone case. Treating ErrNotExist here as success would keep the retry idempotent.

billing/invoice/service.go
rohilsurana: Small doc point: this says it reads only three small filtered pages and is cheap enough for a request path, but the Next() loop below auto-paginates through every page for each of the three statuses. For a customer with many open invoices that is more than three requests. Worth softening the comment so a later caller does not assume a bounded cost.

billing/invoice/service.go
whoAbhishekSah: Kept AmountRemaining for drafts, with live evidence: round 3 case 2 (comment on this PR) was exactly a draft — total 800, customer credit balance 1000 — where Stripe computed amount_due 0 and amount_remaining 0 on the unfinalized draft. Stripe maintains the due amounts live on drafts as line items change, so a draft with amount_remaining 0 means nothing will be owed at finalization. Switching drafts to total > 0 would block that credit-covered case again — the exact fake blocker your first comment removed. A draft whose charge is real shows it in amount_remaining, and round 3 case 1 (negative-total draft) plus the round 2 draft-blocks cases behaved accordingly.

core/deleter/service.go
whoAbhishekSah: Behavior kept, doc fixed. The not-found answer for a fully gone org is deliberate: it lets a caller tell a finished delete from a repeatable one, and returning success for a missing id would make a mistyped org id report a successful delete. The idempotency promise in the docstring is about data inside a half-deleted org — every teardown step tolerates already-deleted rows so a retry completes. The docstring now says both halves explicitly so they no longer read as contradicting.

billing/invoice/service.go
rohilsurana: Withdrawing this one. On an independent second look you are right: Stripe computes amount_due and amount_remaining on drafts too, so a draft that will actually charge keeps amount_remaining > 0 and still blocks here. The only drafts this skips are credit-covered or sub-minimum ones, where the real charge is 0, so nothing is lost. Thanks for the draft example.

rohilsurana (approved):

PR #1869, opened 2026-08-10
fix(admin): stop duplicate API calls across admin tables

Shreyag02:
## Summary

Every server-mode table fetched its first page twice, and the org detail page could fetch its active tab up to four times — visible on staging as `(canceled)` requests.

Our tables pass `defaultSort` to `DataTable` but leave `sort` out of the initial query. `DataTable` merges `defaultSort` in and emits it on mount, changing the connect-query cache key and triggering a second request; the first is aborted mid-flight, after the server has already done the work.

Scoped to duplicate and redundant requests only.

## Changes

| Change | What was happening | Scope |
|---|---|---|
| Seed the initial query with `sort` | The mount emit changed the request, so every table fetched page 1 twice | All 11 server tables |
| Drop the empty `defaultSort` on project members | Sent a sort with an empty field name; that endpoint ignores sort | Project members dialog |
| Keep the org detail tab mounted while billing loads | The tab unmounted and remounted mid-load, replaying every request in it | `details/index.tsx` |
| Cover both billing legs in `isBillingAccountLoading` | The flag missed the `listBillingAccounts` leg, so the side panel showed empty fallbacks as settled | Org context, side panel, edit billing, add tokens |
| Latch "load more" against scroll bursts | Repeat calls cancel the in-flight page and gain nothing | All 11 server tables |
| Scope the members invalidation to its own org | An empty input matched partially, invalidating every org's member list | Members tab |
| Give the shared reads a `staleTime` | `staleTime: 0` + `refetchOnMount` refetched reference data on every navigation | `getOrganization`, roles, org member map |
| Reuse the org resolved from a slug URL | The same org was fetched again by id under a different cache key | Org detail page |
| Share the query state and load-more latch | The latch was copied into all 11 tables and had drifted; query state came in three shapes | `useServerTableQuery`, `useLoadMore` |
| Fetch the org member map only in the projects tab | The full member list was fetched on every org page, read by one tab | Org context → projects tab |
| Gate the invite dialog's queries on `open` | Its trigger sits in the navbar, so two queries ran on every visit | Users list |
| Disable **Add tokens** and **Save** without a billing account | Both submit guards return silently, so the buttons looked live and did nothing | Add tokens, Edit billing |
| Log member-map fetch failures again | Moving the query into a hook dropped the provider's `console.error` | Projects tab |

## Why

| Topic | Mechanism |
|---|---|
| **Why the key changes** | `createMessageKey` omits unset fields, so `sort: []` and `sort: [{name: "created_at", …}]` hash differently — one query in our heads, two cache entries. Seeding the initial sort makes the mount emit structurally identical to state, so the key never changes. |
| **Why the tab remounted** | The gate included `isBillingAccountLoading` but not the `listBillingAccounts` call that *enables* it, and a disabled query reports `isLoading: false`. So it ran `true` → `false` (tab mounts, tables fetch) → `true` (unmounts) → `false` (remounts, tables refetch). It now only includes queries enabled from the first render, so it flips once. |
| **What that exposed** | The same flag was wrong for its own consumers: the side panel rendered **N/A** and **0 / Prepaid** as settled, and both billing submits became reachable before the ids existed. Guards alone weren't enough — with billing settled and no account the flag is `false`, so the buttons enabled and the guards swallowed the click. Both now disable on the condition their guard rejects. |
| **Why a guard wasn't enough for load-more** | `hasNextPage` / `isFetchingNextPage` are last-render values and react-query notifies on a macrotask, but `VirtualizedContent` calls `onLoadMore` straight from `onScroll` — a burst clears the guard before React re-renders, and `fetchNextPage` defaults to `cancelRefetch: true`, aborting the in-flight page. Only a synchronously-flipped ref closes it. |
| **Seed + `staleTime` only work together** | Not seeded + 30s → 1 request. Seeded + `staleTime: 0` → 1. Seeded + 30s → **0**. The seed writes only into an empty key: slug- and id-keyed entries are separate and an edit invalidates only the id-keyed one, so an unconditional seed could overwrite fresher data. |
| **`staleTime` is opt-in** | A client-wide default would make correctness ride on every mutation invalidating what it touches. `SHARED_QUERY_STALE_TIME` is applied to the four queries that need it; everything else keeps `refetchOnMount`. The member map had no writer, so the members tab now invalidates `listOrganizationUsers` too. |
| **Seeding runs in a layout effect** | The view mounts in the same commit and subscribes from a passive effect, which runs after every layout effect — so the seed lands in time while staying out of the render path. A plain `useEffect` costs the extra `GetOrganization` back. |
| **What sharing the hooks changed** | The org and user lists debounced the *table state*, so the grid lagged behind typing; only the request is debounced now, and top-level invoices gains the debounce it never had. The apis tab now resets `offset` on query change, as the other ten did. Audit logs still publishes its request for the CSV export, now from an effect. `useLoadMore` takes the union of the two drifted latches, so neither half regresses. |

## Merged `main`

One conflict in `details/index.tsx`: `main` replaced the inline role queries with `useOrganizationRoles`, this branch removed the member map. Both kept, `isLoading` combined as `isOrganizationLoading || isRolesLoading`. `main`'s new invites table needs nothing here — it's `mode="client"`, so Apsara skips the mount emit, and it has no load-more path.

## Test Plan

- [x] Build, type check and lint pass
- [x] Request counts measured on the production bundle
- [ ] A pass against a real backend — see below

| Check | Result |
|---|---|
| `pnpm build` in `web/sdk` and `web/apps/admin` | Succeed |
| `tsc --noEmit` | Same 21 pre-existing errors before and after; none in touched files. `web/apps/admin` clean |
| `go build ./...`, `go vet` after the merge | Clean |
| `eslint` on all changed files | 0 errors; only pre-existing warnings |
| Every `mode="server"` table audited | 11 in `web/sdk/admin`, all covered; each pairs seeded `sort` with `defaultSort`, project members correctly has neither |
| Cache key, real Apsara + connect-query | Keys differ before the change, match after |
| Seed timing, React 19.2.4 + react-query 5.90.21 | 0 fetches for a layout seed, 1 for a passive one — same under `StrictMode` |
| `useLoadMore`, real hook in jsdom | Burst of 10 → 1 fetch; latch releases; all three early-outs hold; a rejected fetch doesn't wedge it |

**Production bundle in headless Chrome** — every page issued each RPC exactly once, none cancelled:

| Page | RPCs |
|---|---|
| `/organizations` cold | 4 + `configs` = **5, none cancelled** (was 6, with a cancelled `SearchOrganizations`) |
| `/users`, `/audit-logs`, `/invoices` | 3 each |
| Org detail cold via **slug** | 10 — exactly one `GetOrganization`, confirming the seed |
| Org detail → members / projects / pat | 10 each |
| Org detail → apis / invoices / security / tokens | 9 each |
| Org list search, 4 keystrokes | 1 `SearchOrganizations`, none cancelled |

Ran against a stub backend, so tables were empty: real data, latencies and error responses aren't covered, and browser-level load-more wasn't exercised — the hook test above covers the latch instead.

**Still worth a click-through:**

| # | Check | Expected |
|---|---|---|
| 1 | Fast-scroll load-more on a table with real rows | One request per page, none cancelled |
| 2 | Sort, filter and infinite scroll on each table | One request per change |
| 3 | Projects tab | Member avatars render; add-members dropdown still filters |
| 4 | Org side panel during load | Skeletons, not N/A / 0 / Prepaid |
| 5 | Audit logs → export CSV without touching a filter | Downloads, and now respects on-screen filters; previously sent an empty request |

## Out of scope

| Issue | Detail |
|---|---|
| `GetOrganization` **403** on the admins list | Not a duplicate-call artifact. `app/organization#get` grants superusers access only via `platform->superuser`, which needs the org's `platform` relation tuple — written once at creation, no backfill. `OrgCell` issues one per service-user row (`views/admins/columns.tsx:74-78`); the row falls back to the raw id with its button disabled, so nothing breaks. |
| Edit KYC shows a verified org as unverified | **Pre-existing**, but narrowing the loading gate makes it easier to hit — and saving writes the verification away. Fixed separately by the one-hop KYC query. |


web/apps/admin/src/contexts/ConnectProvider.tsx
```diff
@@ -4,12 +4,14 @@ import type { ReactNode } from "react";
...
 
-// Create a QueryClient instance
+/* Otherwise every mount refetches. Mutations invalidate their own keys;
+ * the search tables opt out with staleTime: 0. */
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       retry: false,
       refetchOnWindowFocus: false,
+      staleTime: 30 * 1000,
```
rohilsurana: A 30s app-wide staleTime means every read that does not opt out serves cached data for up to 30s after navigation, so correctness now rides on every mutation invalidating or setting the keys it touches. That is easy to miss. For example, creating an invite from the users list does not invalidate an org's listOrganizationInvitations, so it can show stale for 30s. Worth stating that convention clearly, or scoping the longer staleTime to just the queries that want it.

web/apps/admin/src/pages/organizations/details/index.tsx
```diff
@@ -77,6 +85,26 @@ export default function OrganizationDetailsPage() {
...
+  if (org?.id && org.id !== urlParam && primedOrgId.current !== org.id) {
+    primedOrgId.current = org.id;
+    const orgKey = createConnectQueryKey({
+      schema: FrontierServiceQueries.getOrganization,
+      transport,
+      input: { id: org.id },
+      cardinality: 'finite',
+    });
+    if (queryClient.getQueryData(orgKey) === undefined) {
+      queryClient.setQueryData(
```
rohilsurana, line 100: This seeds the cache with setQueryData in the render body rather than in an effect. A write during render can trigger React's "Cannot update a component while rendering a different component" warning if another mounted component already observes that query, and a discarded concurrent render still mutates the global cache. Moving it into a useEffect keeps it out of the render path.

web/sdk/admin/views/organizations/details/members/index.tsx
```diff
@@ -241,11 +249,12 @@ export function OrganizationMembersView() {
   });
 
   async function invalidateMembersQuery() {
+    // Keys match partially: {} would hit every org; omitting query is deliberate.
     await queryClient.invalidateQueries({
       queryKey: createConnectQueryKey({
         schema: AdminServiceQueries.searchOrganizationUsers,
         transport,
-        input: {},
+        input: { id: organizationId },
```
rohilsurana: If organizationId is ever an empty string, createConnectQueryKey drops the empty id field and the input collapses back to {}, which partial-matches and re-invalidates every org's member list, the exact bug this line fixes. It is guarded by the member-action preconditions today, but a non-empty-id guard here would make it safe by construction.

web/sdk/admin/views/organizations/details/layout/add-tokens-dialog.tsx
```diff
@@ -153,6 +164,7 @@ export const AddTokensDialog = ({ onOpenChange }: InviteUsersDialogProps) => {
                 data-test-id="add-tokens-invite-button"
                 type="submit"
                 loaderText="Adding..."
+                disabled={isBillingAccountLoading || !canCheckout || isSubmitting}
```
rohilsurana: isBillingAccountLoading is redundant here. While billing loads, billingAccountId is empty so canCheckout is already false and the button is disabled. The extra term does not change the outcome.

web/sdk/admin/views/audit-logs/index.tsx
```diff
@@ -140,12 +142,17 @@ export default function AuditLogsView({ appName, onExportCsv, onNavigate }: Audi
     [queryClient],
   );
 
+  // isFetchingNextPage lags a render; the ref doesn't.
+  const isLoadingMoreRef = useRef(false);
```
rohilsurana: This load-more latch (ref, guard, set, await, reset) is copy-pasted across about 11 table files, and the copies already vary slightly. The PR notes a shared useServerTableQuery is a follow-up. This is the concrete duplication it would remove.

rohilsurana (commented):
Review against main. Solid fix and well tested. Two broader risks worth a look (staleTime, render-phase cache write) and three smaller notes.

web/apps/admin/src/contexts/ConnectProvider.tsx
rohilsurana: Are we ok with this in long term?

web/sdk/admin/views/organizations/details/projects/use-add-project-members.tsx
```diff
@@ -15,7 +16,8 @@ interface useAddProjectMembersProps {
 export function useAddProjectMembers({ projectId }: useAddProjectMembersProps) {
   const t = useTerminology();
   const memberLabel = t.member({ case: "capital" });
-  const { orgMembersMap } = useContext(OrganizationContext);
+  const { organization } = useContext(OrganizationContext);
+  const { data: orgMembersMap = {} } = useOrgMembersMap(organization?.id);
```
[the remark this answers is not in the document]
Shreyag02: Real, but not blocking — leaving it for follow-up rather than fixing here.

The window is narrower than it looks. The projects tab calls `useOrgMembersMap` itself on mount (`projects/index.tsx:94`), and `useAddProjectMembers` runs per row from `columns.tsx:118`, so those observers mount together and share one in-flight fetch. Hitting the wrong empty state needs the dropdown *opened* inside the gap where `listProjectUsers` has resolved but `listOrganizationUsers` has not — so a click, on a cold cache, on a slow connection.

`54759942` also gave `useOrgMembersMap` a `staleTime`, so anything that has already loaded the projects tab serves from cache and the window closes entirely.

Queuing it on `fix/admin-ui-followups` instead of growing a PR that is already carrying a refactor.

web/sdk/admin/views/organizations/details/edit/billing.tsx
```diff
@@ -126,6 +126,7 @@ export function EditBillingPanel({ open = false, onClose }: EditBillingPanelProp
   );
 
   const onSubmit = async (data: BillingDetailsForm) => {
+    if (!organizationId || !billingId) return;
```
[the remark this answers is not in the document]
Shreyag02: Fixed in 28da34a1 — and this was worth catching, because the reason it had been left alone was wrong.

The PR description claimed the Edit billing panel was unreachable without a billing account. It isn't: `Edit… → Billing…` in the navbar action menu (`layout/navbar.tsx:157`) is ungated, so it opens for any org. Both ids come from `billingAccount`, so Save sat live over a guard that returns silently — the same dead click already fixed on Add tokens. The description has been corrected.

Save is now `disabled={isLoading || !canSave || isSubmitting}` with `canSave = !!organizationId && !!billingId`, and `onSubmit` reuses that same constant so the two cannot drift apart.

On the other two sites in this comment: `add-tokens-dialog.tsx#L88` still has its guard, and `#L161` was already covered by `canCheckout`. `isBillingAccountLoading` came off that button separately in 5d85db07 — while billing loads there is no account, so `!canCheckout` already disabled it and the extra term never changed the outcome.

web/apps/admin/src/contexts/ConnectProvider.tsx
Shreyag02: No, we weren't — you were right, and it's gone. `ConnectProvider.tsx` is back to exactly what's on `main`, no client-wide `staleTime`.

It's opt-in per query now, via `SHARED_QUERY_STALE_TIME`, applied to the four reads that were actually relying on the default:

- `getOrganization` by id — the slug-resolve seed is worthless without it (seeded + `staleTime: 0` is still 1 request); edit and block/unblock both invalidate that key
- both legs of `useOrganizationRoles` — reference data, and nothing in the console writes roles
- `useOrgMembersMap` — one observer per project row plus the project members page

Everything else keeps `refetchOnMount`, so your `listOrganizationInvitations` example can't go stale.

The convention is stated where you'd read it before using it, in `constants.ts`: only safe where every writer invalidates the key, so each query opts in and says why. That precondition wasn't holding for the member map — nothing invalidated it — so the members tab now invalidates `listOrganizationUsers` alongside its own table. Without that, a member removed there would have shown for 30s in the projects tab.

web/apps/admin/src/pages/organizations/details/index.tsx
Shreyag02: Moved out of render — but to a **layout** effect rather than a passive one, because a plain `useEffect` lands too late to be worth anything.

The view mounts in this same commit, and react-query subscribes from `useSyncExternalStore` (`useBaseQuery.js:56`), which React runs as a passive effect. Passive effects for a commit run after *all* layout effects, so:

| Seed runs in | Child fetches |
|---|---|
| render (what you flagged) | 0 |
| **layout effect** | 0 |
| passive effect | 1 |

Measured against real React 19.2.4 + react-query 5.90.21, identical under `StrictMode`. So a `useEffect` here would have quietly given back the `GetOrganization` this was saving.

The layout effect answers both halves of your comment: it's out of the render path, and layout effects only run on committed renders, so a discarded concurrent render no longer mutates the cache. The `useRef` latch went with it — the dep array plus the existing empty-key check cover it.

Confirmed end to end since: a cold load on a slug URL now issues exactly one `GetOrganization`.

web/sdk/admin/views/organizations/details/members/index.tsx
Shreyag02: Guarded — `invalidateMembersQuery` now returns early on an empty id.

It stopped being optional: scoping the `staleTime` meant `useOrgMembersMap` needed a writer, so this function gained a second `invalidateQueries` for `listOrganizationUsers`. Adding a second partial-match invalidation while relying on "the member-action preconditions hold today" would have doubled exactly the hazard you named.

Confirmed the mechanism rather than assuming it — `message-key.js` says it outright: *"Default values are omitted (both implicit and explicit field presence)"*. `""` is the proto3 default for a string, so `{ id: "" }` really does collapse to `input: {}` and prefix-matches every org.

Both invalidations sit behind the one guard, and `updateMember` / `removeMember` are the only callers, so it's safe by construction now.

Four other sites build org-scoped keys the same way without a guard — `remove-invite-dialog.tsx`, `invite-users-dialog.tsx`, `domains-list.tsx`, `sessions/index.tsx`. The last two are worse in principle, since they write `|| ""` explicitly. None is touched by this PR, so I've left them rather than fix in passing.

web/sdk/admin/views/organizations/details/layout/add-tokens-dialog.tsx
Shreyag02: Agreed, dropped. `disabled={!canCheckout || isSubmitting}`.

`isLoading` is `isPending && isFetching`, so it's only true when there's no data at all — which means `billingAccount` is undefined, `billingAccountId` is `""`, and `!canCheckout` has already disabled the button. There's no state where the term flipped the result: a cached-data refetch surfaces as `isFetching`, not `isLoading`, so it can't produce a truthy flag alongside a real account either.

It was this file's only reader of the flag, so it came off the context destructure too. The flag itself stays — `edit/billing.tsx` and both side-panel sections still use it.

Related, from the same review pass: **Edit billing** had the mirror-image bug and is now fixed the same way. The description had claimed that panel was unreachable without a billing account; it isn't — `Edit… → Billing…` (`layout/navbar.tsx:157`) is ungated, so Save sat live over a guard that returns silently. Both panels now disable on the condition their guard rejects.

rohilsurana (approved):

PR #1870, opened 2026-08-11
chore(server): remove the billing plans boot loader flow

rohilsurana:
## What

Removes the `billing.plans_path` boot loader end to end. Billing products, plans, prices, and features are managed through the reconcile flow now (the `BillingProduct` and `BillingPlan` kinds), so the server no longer reads a plans file at startup.

- `cmd/serve.go`: drop the billing blob store and the `MigrateBillingPlans` call.
- `internal/bootstrap`: drop the `BillingPlanRepository` and `PlanService` dependencies and the `MigrateBillingPlans` method.
- `billing/config.go`: remove the `plans_path` field.
- `internal/store/blob/plan_repository.go`: removed, no longer used.
- `config/sample.config.yaml`: drop the sample entry.
- `test/e2e`: the billing regression suite seeds the fixtures it needs (`support_credits` and `enterprise_yearly`) through the admin API in `SetupSuite` instead of the boot loader.

`UpsertPlans` and `plan.File` stay: the admin `CreatePlan` handler still uses them to create a plan with its products.

## Why it is safe

Existing plans and products live in the database, so this is a no-op for a server that has already migrated. The reconcile flow (`frontier reconcile` with the `BillingProduct` / `BillingPlan` kinds) owns this state now.

## Gating

This is the billing counterpart to #1767 (the resources_config boot loader removal). Same rule: a deployment must run the reconcile flow so plans and products are owned there, and drop `plans_path` from its own config, before taking this build.

One boot-time dependency to call out: if a deployment sets `billing.customer.credit_overdraft_product` (or `billing.customer.default_plan`), that product or plan must already exist in the database before the server starts. `InvoiceService.Init` resolves the overdraft product at boot and fails startup if it is missing, and `EnsureDefaultPlan` needs the default plan to exist. The boot loader used to seed these; now they must come from the reconcile flow or already be present. Deployments that already ran the loader in a past release have this data, so this is a no-op for them. A fresh install must seed the catalog through reconcile first.

## Verification

`go build ./...`, `go vet ./...`, and unit tests for `internal/bootstrap`, `billing/plan`, `core/event`, and `internal/api/v1beta1connect` pass. gofmt and lint clean.


cmd/serve.go
```diff
@@ -200,14 +193,6 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
...
 
-	// apply billing plans
-	if cfg.Billing.PlansPath != "" {
-		if err = deps.BootstrapService.MigrateBillingPlans(ctx); err != nil {
-			return err
-		}
-		logger.Info("migrated billing plans")
-	}
-
 	// apply roles over nil org id
```
rohilsurana, line 179: **Likely hard boot failure when `credit_overdraft_product` is set.**

Removing the boot plan loader drops the only thing that seeded `credit_overdraft_product` before `InvoiceService.Init`. On a fresh database with `billing.customer.credit_overdraft_product` set, `Init` calls `productService.GetByID(...)`, gets not-found, and returns `failed to get credit overdraft product`, so `StartServer` fails and the server never comes up. The e2e suite hides this with a new pre-start DB seeder, but production has no equivalent replacement.

billing/config.go
```diff
@@ -3,13 +3,11 @@ package billing
...
-	StripeAutoTax        bool     `yaml:"stripe_auto_tax" mapstructure:"stripe_auto_tax"`
-	StripeWebhookSecrets []string `yaml:"stripe_webhook_secrets" mapstructure:"stripe_webhook_secrets"`
-	// PlansPath is a directory path where plans are defined
-	PlansPath           string                `yaml:"plans_path" mapstructure:"plans_path"`
-	DefaultCurrency     string                `yaml:"default_currency" mapstructure:"default_currency"`
-	PaymentMethodConfig []PaymentMethodConfig `yaml:"payment_method_config" mapstructure:"payment_method_config"`
+	StripeKey            string                `yaml:"stripe_key" mapstructure:"stripe_key"`
+	StripeAutoTax        bool                  `yaml:"stripe_auto_tax" mapstructure:"stripe_auto_tax"`
+	StripeWebhookSecrets []string              `yaml:"stripe_webhook_secrets" mapstructure:"stripe_webhook_secrets"`
+	DefaultCurrency      string                `yaml:"default_currency" mapstructure:"default_currency"`
```
rohilsurana, line 9: **Plan/product seeding is removed with no replacement, and `plans_path` is now silently ignored.**

For a deployment that defined its catalog via `plans_path` and set `default_plan`, the key is now an unknown field that config loading ignores. Nothing seeds the plans on boot, so `EnsureDefaultPlan` fails for every new org (only logged in `listener.go`, so the org comes up with no subscription or entitlements), and `CreateCheckout` fails plan-not-found for those plan names. Please confirm the reconcile/admin path fully covers this before merge.

rohilsurana (commented):
Automated code review (xhigh). Findings below, most important first. The first two look like real boot/runtime regressions worth confirming before merge.

test/e2e/testbench/testbench.go
```diff
@@ -114,6 +121,23 @@ func Init(appConfig *config.Frontier) (*TestBench, error) {
...
+	// seed rows the server needs at boot before it starts. The database is
+	// migrated but the server is not up yet, so a seeder can write directly.
+	if len(preStartSeeders) > 0 {
+		seedClient, err := db.New(appConfig.DB)
+		if err != nil {
+			return nil, err
+		}
+		for _, seed := range preStartSeeders {
+			if err := seed(context.Background(), seedClient); err != nil {
+				return nil, errors.Join(err, seedClient.Close())
```
rohilsurana: **Error path here leaks already-started containers.**

If `db.New` fails or a `PreStartSeeder` returns an error, `Init` returns `(nil, err)` before `stripeClose` and `te.close` are wired together (that happens later, around line 157). Postgres, SpiceDB, and (for the mock key) stripe-mock are already running, so they leak. `SetupSuite` gets a nil testBench and never calls `Close`, so repeated failing runs pile up leaked containers and networks. Consider tearing down started resources on this path.

rohilsurana: Fixed in 26001fb1: the containers are torn down on every seeder error path. Used `te.close()` + `stripeClose()` rather than `te.Close()`, since the server is not started yet and `te.Close()` would send SIGINT to the test process itself.

cmd/serve.go
rohilsurana: Covered for our rollout: prod and staging set `credit_overdraft_product` by UUID, the product already exists in the DB and is in the reconcile billing-products export (`behavior: credits`), so `Init` resolves it at boot. Fresh installs must seed the catalog via reconcile first, which is now noted in the PR description. Leaving `Init` fail-fast rather than changing that core behavior in this PR.

billing/config.go
rohilsurana: Covered: pixxel sets no `default_plan`, and plans are owned by reconcile now. Removed the stale `plans_path` docs. `plans_path` being ignored is the intended removal; fresh-install ordering is noted in the PR description.

whoAbhishekSah (approved):

PR #1871, opened 2026-08-11
fix(bootstrap): harden AppendSchema against list errors and missing metadata description

rohilsurana:
## What

Fixes two long-standing bugs in `bootstrap.AppendSchema`. Both are already on `main` and are independent of the boot loader removal work.

### 1. A failed permission list was swallowed

```go
existingPermissions, err := s.permissionService.List(ctx, permission.Filter{})
if err != nil {
    return nil // <- returns success without applying the schema
}
```

When the list call failed, `AppendSchema` returned `nil` before `applySchema` ran. So a transient list error at boot skipped the schema re-apply but still reported success. Nothing was applied and nothing was dropped, but boot looked healthy while silently doing no schema work. On a first boot, where the base schema has not been applied yet, that leaves the server with no schema at all. The fix returns the wrapped error so boot stops instead of reporting a false success.

### 2. The metadata description read had inverted logic and could panic

```go
if v, ok := existingPermission.Metadata["description"]; !ok {
    description = v.(string) // <- runs only when the key is absent, then asserts nil to string
}
```

The `!ok` branch runs only when the `description` key is missing. In that case `v` is a nil `any`, so `v.(string)` panics. When the key is present, the value was ignored and the description was dropped. The logic is now pulled into a small `permissionDescription` helper that reads the value when present, returns `""` when it is missing or not a string, and never panics.

## Why now

These are pre-existing bugs, but #1767 removes the boot-time config schema loader and makes `AppendSchema` the only path that applies the schema at boot. That raises how much these two edge cases matter, so it is worth fixing them on their own.

## Tests

- `Test_AppendSchema` proves a list error is returned, not swallowed.
- `Test_permissionDescription` covers nil metadata, a present string, a missing key, and a non-string value (the last two used to panic).
- `Test_existingPermissionsAsServiceDefinition` covers the mapping from existing permissions into the schema, so a wrong field or dropped description is caught.

`go build ./...`, `go vet ./...`, `go test ./internal/bootstrap/...` pass. gofmt and golangci-lint clean.

### DIFF
--- a/internal/bootstrap/service.go
+++ b/internal/bootstrap/service.go
@@ -188,20 +188,29 @@ func (s Service) AppendSchema(ctx context.Context, customServiceDefinition schem
 		return fmt.Errorf("AppendSchema: listing existing permissions: %w", err)
 	}
 	for _, existingPermission := range existingPermissions {
-		description := ""
-		if existingPermission.Metadata != nil {
-			if v, ok := existingPermission.Metadata["description"]; !ok {
-				description = v.(string)
-			}
-		}
 		existingServiceDefinition.Permissions = append(existingServiceDefinition.Permissions, schema.ResourcePermission{
 			Name:        existingPermission.Name,
 			Namespace:   existingPermission.NamespaceID,
-			Description: description,
+			Description: permissionDescription(existingPermission),
 		})
 	}
 
 	return s.applySchema(ctx, schema.MergeServiceDefinitions(customServiceDefinition, existingServiceDefinition))
+}
+
+// permissionDescription reads the human description out of a permission's
+// metadata. It returns "" when the metadata is missing, has no description, or
+// stores a non-string value, and never panics on a missing key.
+func permissionDescription(p permission.Permission) string {
+	if p.Metadata == nil {
+		return ""
+	}
+	v, ok := p.Metadata["description"]
+	if !ok {
+		return ""
+	}
+	desc, _ := v.(string)
+	return desc
 }
 
 // applySchema builds and apply schema over az engine and db
--- a/internal/bootstrap/service.go
+++ b/internal/bootstrap/service.go
@@ -16,6 +16,7 @@ import (
 	"github.com/raystack/frontier/core/relation"
 	"github.com/raystack/frontier/core/role"
 	"github.com/raystack/frontier/internal/bootstrap/schema"
+	"github.com/raystack/frontier/pkg/metadata"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


internal/bootstrap/service.go
```diff
@@ -185,25 +185,34 @@ func (s Service) AppendSchema(ctx context.Context, customServiceDefinition schem
 
 	existingPermissions, err := s.permissionService.List(ctx, permission.Filter{})
 	if err != nil {
-		return nil
+		return fmt.Errorf("AppendSchema: listing existing permissions: %w", err)
```
rohilsurana, line 187: **Boot now aborts on a transient list error.**

This turns a previously swallowed `permissionService.List` error into a propagated one, so a transient list failure during startup now fails boot instead of being tolerated. That looks like the intent of the fix. Flagging only so you can confirm boot should not stay tolerant of a temporary list error.

internal/bootstrap/service_test.go
```diff
@@ -290,3 +310,41 @@ func Test_migrateRole(t *testing.T) {
 		roleSvc.AssertNotCalled(t, "Update")
 	})
 }
+
+func Test_AppendSchema(t *testing.T) {
```
rohilsurana, line 314: **Success path of `AppendSchema` is untested.**

This test only covers the list-error branch. The success path (the loop over existing perms, description wiring, the merge, and `applySchema`) has no coverage. A regression there (wrong field wired, description dropped from the `ResourcePermission`, or merge args swapped) would ship green. Consider a case with a non-empty permission list that asserts the merged schema.

internal/bootstrap/service_test.go
```diff
@@ -290,3 +310,41 @@ func Test_migrateRole(t *testing.T) {
...
+	t.Run("returns the error when listing existing permissions fails", func(t *testing.T) {
+		// A failed list must not be swallowed: boot would then apply an empty
```
rohilsurana: **This comment misstates the old behavior.**

The old code returned before `applySchema` ran, so a list error was a silent no-op success. No schema was written and nothing was dropped. It did not apply an empty schema or drop existing permissions. Worth fixing the comment so it does not read as if this change fixes data loss rather than a silent false success.

internal/bootstrap/service.go
```diff
@@ -185,25 +185,34 @@ func (s Service) AppendSchema(ctx context.Context, customServiceDefinition schem
...
 		})
 	}
 
 	return s.applySchema(ctx, schema.MergeServiceDefinitions(customServiceDefinition, existingServiceDefinition))
 }
 
+// permissionDescription reads the human description out of a permission's
+// metadata. It returns "" when the metadata is missing, has no description, or
+// stores a non-string value, and never panics on a missing key.
+func permissionDescription(p permission.Permission) string {
```
rohilsurana: **`permissionDescription` can be simpler and cheaper.**

Two small things on this helper:

1. The three guard branches reduce to one line: `desc, _ := p.Metadata["description"].(string); return desc`. Indexing a nil map is safe, and a two-value assertion on a nil or non-string value returns `""` without panicking, so all four cases collapse.
2. It takes the whole `permission.Permission` by value but only reads `Metadata`. Since it runs once per permission in the `AppendSchema` loop, passing just the metadata map avoids copying the full struct each time.

rohilsurana (commented):
Automated code review (xhigh). Findings below. One behavior question, one test-comment fix, a coverage gap, and a small simplification.

internal/bootstrap/service_test.go
rohilsurana: Fixed in 9fa54032: extracted `existingPermissionsAsServiceDefinition` and added `Test_existingPermissionsAsServiceDefinition`, which asserts the Name/Namespace/Description wiring for a non-empty permission list.

internal/bootstrap/service_test.go
rohilsurana: Fixed in 9fa54032: corrected the comment. The old code returned before `applySchema` ran, so a list error was a silent no-op success, not data loss. The PR description is corrected too.

internal/bootstrap/service.go
rohilsurana: Fixed in 9fa54032: `permissionDescription` now takes the metadata map and collapses to `desc, _ := m[\"description\"].(string); return desc`.

internal/bootstrap/service.go
rohilsurana: Intended: a transient list error at boot should fail boot rather than silently skip the schema re-apply. The previous `return nil` reported a false success.

whoAbhishekSah (approved):

PR #1872, opened 2026-08-11
docs: rewrite the custom resources guide for the reconcile flow

rohilsurana:
## What

Rewrites the "Custom Resources and Permissions" guide to match the new model. The boot-time `resources_config_path` loader is gone (removed in #1767), so custom permissions and roles are registered through the `frontier reconcile` flow or the admin API now, not a config file read at startup.

Changes:

- Rewrites "How custom resources are registered": the reconcile `Permission` / `Role` document format, the `frontier reconcile` and `frontier export` commands, and the reconcile rules (identity-only permissions, delete needs `delete: true`, `app/...` namespaces rejected).
- Explains what happens when a permission is registered: `AppendSchema` regenerates the schema (base schema plus every permission in the database) and writes to Postgres and SpiceDB, and the same step runs at boot.
- Keeps the accurate schema-generation and role-mapping sections, updated for the new flow.
- Updates the `user/project` proxy section: `app/...` namespaces are now rejected with an error by both reconcile and the admin API, instead of being silently ignored from a config file.
- Updates the role examples to `kind: Role` documents and the reconcile role semantics: a listed field is the whole desired value (so listing `permissions` replaces the set), while an omitted field keeps the role's default.
- Removes emdashes to match our writing style.

## Why

The old guide described the removed boot-time config-file loader end to end. #1767 added a short banner as a stopgap; this replaces the banner with a full rewrite.

## Stacked on #1767

This is stacked on `chore/remove-resources-config-path-flow` (#1767), because the guide describes the world after that PR removes the loader. Retarget to `main` once #1767 merges.

### DIFF
--- a/docs/content/docs/authz/custom-resources.mdx
+++ b/docs/content/docs/authz/custom-resources.mdx
@@ -67,6 +67,9 @@ A few rules the reconcile flow enforces for permissions:
   updated.
 - Nothing is deleted by leaving it out. A permission that is in the server but missing from
   the file fails the plan. To remove one, mark its entry with `delete: true`.
+- A namespace must be in `service/resource` form: two non-empty parts, each lowercase
+  alphanumeric. So `compute/machine` works, but `compute/machine-v2` does not (the hyphen is
+  not alphanumeric). The action name must be alphanumeric too.
 - A namespace under `app` or `app/...` is rejected, because the base schema owns those types
   (see [Why a separate `user/project` namespace](#why-a-separate-userproject-namespace)).
 
@@ -245,7 +248,8 @@ there, and the generator mirrors it onto the real project as
 effect, `user/project` is the allowed way to hang project-level capabilities off `app/project`.
 
 The name `user/project` is not special, and there is no naming convention to follow. You can
-pick any `service/resource` namespace, as long as it is not under `app/...`. The generator mirrors
+pick any `service/resource` namespace whose two parts are lowercase alphanumeric, as long as it
+is not under `app/...`. The generator mirrors
 every custom permission onto `app/project` (and `app/organization`) no matter which namespace you
 chose, so the name does not change where the check runs. `user/project` is simply the name
 Frontier uses here, because the slug it produces, `user_project_createcomputemachine`, reads as
@@ -314,8 +318,8 @@ This does three things:
 - Project-level capabilities (`create`, project-wide `list`) go on `user/project`. Checked
   against `app/project:<project_id>`.
 - Treat `user/project` as a stand-in for `app/project` that you are allowed to write to. The
-  name is your choice: any non-`app/...` namespace works. `user/project` is just the example
-  used here.
+  name is your choice: any non-`app/...` namespace works, as long as its two parts are lowercase
+  alphanumeric. `user/project` is just the example used here.
 
 This keeps create and list anchored on the project and avoids the dead resource-level `create`
 rule the generator would otherwise leave unused.
@@ -432,7 +436,8 @@ custom role later, mark its entry with `delete: true`. A predefined role cannot 
 - Registering a permission merges generated rules into the base schema and writes them to
   Postgres and SpiceDB. Boot re-applies the base schema plus the permissions already in the
   database.
-- The **resource owner**, **platform admin**, and **Owner** (`app_organization_owner`) can always
-  perform every action on a resource. Project and direct grants depend on the roles in use.
+- The **resource owner**, **platform admin**, **Owner** (`app_organization_owner`, org-wide), and
+  **Project Owner** (`app_project_owner`, within its project) can always perform every action on a
+  resource. Other project and direct grants depend on the roles in use.
 - Per-item actions (`get`, `update`, `delete`) live on the resource namespace; project-level
   actions (`create`, `list`) live on `user/project` and are checked against the project.
--- a/docs/content/docs/authz/custom-resources.mdx
+++ b/docs/content/docs/authz/custom-resources.mdx
@@ -5,63 +5,91 @@ order: 7
 
 # Custom Resources and Permissions
 
-> **Note:** The boot-time `resources_config_path` loader described below has been
-> removed. Frontier no longer reads a resource config file at startup. Create and
-> update custom permissions and roles through the admin API (`CreatePermission`,
-> `CreateRole`) or the `frontier reconcile` flow instead. See
-> [the declarative reconcile RFC](https://github.com/raystack/frontier/blob/main/docs/rfcs/0001-declarative-reconcile.md).
-> The mechanics below still explain the schema that bootstrap applies from the
-> base schema plus the permissions already in the database.
-
 Frontier lets services register their own resource types (for example `compute/machine`).
 Once registered, Frontier can answer permission checks on those resources the same way it
 does for built-in types like projects and organizations.
 
-This page explains two things:
+This page explains three things:
 
-1. How a custom resource type is loaded into Frontier.
-2. What permission rules Frontier generates for it, and which role each action ends up with.
+1. How you register a custom resource type and its actions.
+2. What permission rules Frontier generates for it.
+3. Which role each action ends up with.
 
 ---
 
-## How custom resources are loaded
+## How custom resources are registered
 
-A custom resource type is described in a small config file. Each file lists a namespace and
-the actions (permissions) that namespace supports. Here is the built-in `compute/machine`
-example from `resources_config/compute_machine.yml`:
-
-```yaml
-permissions:
-  - name: get
-    namespace: compute/machine
-  - name: create
-    namespace: compute/machine
-  - name: update
-    namespace: compute/machine
-  - name: delete
-    namespace: compute/machine
-```
-
-A namespace has two parts joined by a slash: `service/resource`. So `compute/machine` is the
+A custom resource type is a **namespace** plus the **actions** (permissions) it supports. A
+namespace has two parts joined by a slash, `service/resource`. So `compute/machine` is the
 `machine` resource in the `compute` service.
 
-At startup Frontier runs a bootstrap step (`MigrateSchema`) that does the following:
+You register the actions in one of two ways:
 
-1. Reads every resource config file into a `ServiceDefinition` (the list of namespaces and
-   their actions).
-2. Loads the permissions already in Postgres — including any added later through the
-   `CreatePermission` API — and merges them in, so a restart does not drop them.
-3. Loads the base SpiceDB schema (`base_schema.zed`), which defines users, organizations,
+- **The reconcile flow** (`frontier reconcile`). You write the actions you want in a
+  desired-state YAML file and apply it. This is the recommended path: the file is the full
+  desired state, so it reads back the same way it was written and fits GitOps.
+- **The admin API** (`CreatePermission`, `CreateRole`). The imperative version of the same
+  thing, useful for one-off changes or scripting.
+
+Here is the `compute/machine` example as a reconcile file. Each entry is a namespace and an
+action name:
+
+```yaml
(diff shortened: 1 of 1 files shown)


rohilsurana:
@coderabbitai review

rohilsurana (commented):
Automated code review (xhigh) of the docs rewrite. Findings below, most important first. The first two are worth fixing before merge: the main example and the "omitting permissions" note both describe reconcile as adding/keeping permissions when it actually replaces the whole set, so a reader following them would strip permissions from built-in roles.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -204,58 +232,64 @@ because at check time you have no machine id to check against.
...
+  # app namespaces are reserved for the base schema and are rejected.
+  - namespace: user/project
+    name: createcomputemachine
+  - namespace: user/project
+    name: listcomputemachines
+---
+apiVersion: v1
+kind: Role
+spec:
   - name: app_project_owner          # extend the built-in Project Owner role
```
rohilsurana: **The main example strips `app_project_administer` from Project Owner.**

This Role document lists the predefined `app_project_owner` with `permissions:` set to only the two custom slugs. Reconcile replaces a listed role's whole permission set (this page's own Choice 1 rule, and `RoleSpec.resolve` replaces when `permissions` is set). The predefined role ships with exactly `[app_project_administer]`, so running this example strips `app_project_administer` from Project Owner org-wide. Because the generated custom-resource rules reach Project Owners only through `granted->app_project_administer`, they also lose the automatic access the "Works by default" section promises. The `# extend the built-in Project Owner role` comment makes it worse by implying the permissions are added, not replaced. The example should re-list `app_project_administer` alongside the new slugs (or use a custom role).

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -368,26 +408,31 @@ roles:
...
 
 ### In short
 
 - The **Owner** (`app_organization_owner`), **Project Owner** (`app_project_owner`), and the
   **platform admin** get every custom action for free. The org **Admin** role
   (`app_organization_manager`) does not.
 - Every other role gets an action only when you grant it.
-- Re-using a built-in role name replaces its permissions, so list everything you want it to keep.
+- Listing `permissions` on a role replaces the set, so list everything you want it to keep.
+  Omitting `permissions` keeps the role's current values.
```
rohilsurana: **"Omitting `permissions` keeps the role's current values" is not true for reconcile.**

Reconcile resolves an omitted `permissions` field to the shipped default, not to whatever is on the server now. So if someone earlier added a permission to a built-in role and later re-applies a file that omits `permissions`, `diffRoles` plans an update that resets the role to its shipped default and silently drops the added permission. For a custom role the omission resolves to empty and fails validation ("must list at least one permission"), so the apply errors out. This also contradicts line 388. Suggest rewording to: omitting `permissions` resets the role to its shipped default (and is invalid for a custom role).

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -368,26 +408,31 @@ roles:
...
+A new role starts empty, so you only list what you want it to have. Once it exists, you assign
+this role to a user or group on a project, the same way you assign any other role. To remove a
+custom role later, mark its entry with `delete: true`. A predefined role cannot be deleted.
 
 ### In short
 
 - The **Owner** (`app_organization_owner`), **Project Owner** (`app_project_owner`), and the
   **platform admin** get every custom action for free. The org **Admin** role
   (`app_organization_manager`) does not.
 - Every other role gets an action only when you grant it.
```
rohilsurana, line 440: **The role section never states the whole-desired-state rule.**

The permission section documents this at line 67, but the role section does not. Two behaviors bite operators: a custom platform role that exists on the server but is missing from the file aborts the whole run ("roles exist on the server but are not in the file ... keep them or mark them 'delete: true'"), and a customized predefined role missing from the file is reset to its shipped default. Worth a short note here mirroring line 67.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -204,58 +232,64 @@ because at check time you have no machine id to check against.
...
 > `app_project_createcomputemachine`.
 
-### Config
+### Declaring it
 
 Put the per-item actions (`get`, `update`, `delete`) on the resource namespace, and the
 project-level capabilities (`create`, project-wide `list`) on `user/project`. Then grant the
-project-level ones to a project-scoped role such as the built-in Project Owner:
+project-level ones to a project-scoped role such as the built-in Project Owner. A single
+reconcile file can hold more than one document, separated by `---`, so the permissions and the
```
rohilsurana, line 270: **Say that `Permission` must come before `Role` in the file.**

The text tells readers to combine both kinds in one file and shows `Permission` first, but never says the order is required. Reconcile enforces `Role` depends on `Permission` and aborts with "kind Permission must come before kind Role in the file". A reader who lists the Role first (natural, since the role is the goal) hits that error with no guidance from this page. One sentence stating the required order fixes it.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -297,22 +331,22 @@ nothing until you grant them.
 
 You do not have to set up any roles for these. They work as soon as the resource is registered:
 
-- **Owner** (`app_organization_owner`) — can do every action on every custom resource in the
+- **Owner** (`app_organization_owner`), can do every action on every custom resource in the
```
rohilsurana: **Comma between subject and verb across all four bullets.**

The emdash was replaced with a plain comma, so these read as "**Owner** (`app_organization_owner`), can do every action ..." (lines 334-338). The comma splits the subject from its verb. Reword instead, for example "**Owner** (`app_organization_owner`) can do every action ..." or use a colon.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -174,11 +202,11 @@ check them against the project id with the caller as the subject:
 Check(
   subject    = app/user:<userid>,                  # the authenticated caller
   permission = user_project_createcomputemachine,
-  resource   = app/project:<project_id>,           # the container — it already exists
+  resource   = app/project:<project_id>,           # the container, it already exists
 )
 ```
 
-### These actions are not special — it is a modeling choice
+### These actions are not special, it is a modeling choice
```
rohilsurana: **This heading is a comma splice.**

"These actions are not special, it is a modeling choice" joins two full sentences with a comma. It came from removing the emdash. Use a colon or reword, for example "These actions are not special: it is a modeling choice".

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -5,63 +5,91 @@ order: 7
...
+`machine` resource in the `compute` service.
+
+You register the actions in one of two ways:
 
-A custom resource type is described in a small config file. Each file lists a namespace and
-the actions (permissions) that namespace supports. Here is the built-in `compute/machine`
-example from `resources_config/compute_machine.yml`:
+- **The reconcile flow** (`frontier reconcile`). You write the actions you want in a
+  desired-state YAML file and apply it. This is the recommended path: the file is the full
+  desired state, so it reads back the same way it was written and fits GitOps.
```
rohilsurana: Minor style: "fits GitOps" is a buzzword the writing style avoids, and it does not add anything the sentence has not already said. Same kind of jargon shows up at line 31 ("imperative") and line 86 ("idempotent"); plain wording would read better for a non-expert.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -204,58 +232,64 @@ because at check time you have no machine id to check against.
...
 `app/project#user_project_createcomputemachine`. That mirrored permission is what you check. In
-effect, `user/project` is the config-legal way to hang project-level capabilities off
-`app/project`.
+effect, `user/project` is the allowed way to hang project-level capabilities off `app/project`.
 
 The name `user/project` is not special, and there is no naming convention to follow. You can
-pick any `service/resource` namespace, as long as it is not under `app/*`. The generator mirrors
+pick any `service/resource` namespace, as long as it is not under `app/...`. The generator mirrors
 every custom permission onto `app/project` (and `app/organization`) no matter which namespace you
 chose, so the name does not change where the check runs. `user/project` is simply the name
```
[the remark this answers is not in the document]
rohilsurana: Fixed in e848473b: the permission rules now state a namespace must be `service/resource` with two non-empty lowercase alphanumeric parts (so `compute/machine-v2` is rejected), and the same restriction is added to the rule of thumb and the user/project section.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -368,26 +408,31 @@ roles:
...
-- The **resource owner**, **platform admin**, and **Owner** (`app_organization_owner`) can always perform every action
-  on a resource. Project and direct grants depend on the roles in use.
+- A custom resource is registered by declaring its `service/resource` namespace and actions
+  through `frontier reconcile` (a desired-state file) or the admin API. There is no boot-time
+  config file.
+- Registering a permission merges generated rules into the base schema and writes them to
+  Postgres and SpiceDB. Boot re-applies the base schema plus the permissions already in the
+  database.
+- The **resource owner**, **platform admin**, and **Owner** (`app_organization_owner`) can always
+  perform every action on a resource. Project and direct grants depend on the roles in use.
```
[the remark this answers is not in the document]
rohilsurana: Fixed in e848473b: added Project Owner (`app_project_owner`, within its project) to the quick reference.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Good catch, this was a real bug. Fixed in ed41255e: the example re-lists `app/project:administer` alongside the two custom slugs, the comment no longer says extend, and a note explains that listing permissions replaces the whole set, so the shipped permission must be kept or Project Owner loses its automatic access.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: reworded to say omitting `permissions` resets the role to its shipped default (a no-op only for an unchanged predefined role, and an error for a custom role, which must list at least one). The Choice 1 note is aligned to match.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: added a note at the top of the role section mirroring the permission rule. A custom role on the server but missing from the file stops the run (keep it or mark `delete: true`), a customized predefined role left out resets to its shipped default, and predefined roles are never deleted.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: added a sentence stating a `Permission` document must come before a `Role` document that uses it, which is why the permissions come first in the example.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: dropped the comma between subject and verb in all four bullets.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: changed the heading to use a colon.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Fixed in ed41255e: reworded to drop fits GitOps, imperative, and idempotent in favor of plain wording.

whoAbhishekSah (approved):

PR #1873, opened 2026-08-12
fix(server): set ReadHeaderTimeout and IdleTimeout on all HTTP servers

### LINKED ISSUE #1819: Set ReadHeaderTimeout and IdleTimeout on the HTTP servers
All three `http.Server` instances in `pkg/server/server.go` — the connect server, the metrics server, and the UI server — are constructed with only `Addr` and `Handler`. No timeout fields have ever been set, so connections can sit idle or withhold request headers for as long as they like.

## Proposal

- Set `ReadHeaderTimeout` and `IdleTimeout` on all three servers.
- Pick an `IdleTimeout` larger than the keep-alive timeout of any fronting proxy, so the proxy never reuses a connection the server is about to close.
- Deliberately do **not** set a blanket `WriteTimeout`: the connect server has long streaming export responses (for example `ExportAuditRecords`), and a write deadline would cut them off.
- `ReadTimeout`, if added at all, needs a value that accounts for large request bodies.

Came out of review discussion on #1818, which touched the connect server construction; the gap itself predates that PR.

whoAbhishekSah:
Fixes #1819.

All three `http.Server` instances in `pkg/server/server.go` (connect, metrics, UI) were built with only `Addr` and `Handler`. Nothing limited how long a client could withhold request headers or keep an idle connection open.

- The `idle_timeout` default exceeds common proxy keep-alive timeouts (ALB 60s, nginx 75s). Deployments behind a proxy with a longer keep-alive should raise it, so the proxy never sends a request on a connection the server is already closing.
- No `WriteTimeout` on purpose: the connect server streams long export responses (e.g. `ExportAuditRecords`) that a write deadline would cut off.
- No `ReadTimeout` either; it would need a value that accounts for large request bodies, and the header timeout already covers the slow-client case this issue is about.

## Testing

- `go build ./pkg/server/...` and `go test ./pkg/server/` pass.
- Verified with a throwaway config test that `Load` fills both defaults (`10s` / `180s`) when the fields are absent from the config file.

### DIFF
--- a/config/sample.config.yaml
+++ b/config/sample.config.yaml
@@ -58,6 +58,12 @@ app:
   metrics_port: 9000
   # how long each server waits for in-flight requests to finish during shutdown
   shutdown_grace_period: 10s
+  # how long each server waits for a client to send its request headers
+  read_header_timeout: 10s
+  # how long each server keeps an idle keep-alive connection open.
+  # keep it larger than the keep-alive timeout of any fronting proxy,
+  # so the proxy never sends a request on a connection the server is closing
+  idle_timeout: 180s
   # enable pprof endpoints for cpu/mem/mutex profiling
   profiler: false
   host: "127.0.0.1"
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -60,6 +60,16 @@ type Config struct {
 	// requests to finish during shutdown before cutting them off
 	ShutdownGracePeriod time.Duration `yaml:"shutdown_grace_period" mapstructure:"shutdown_grace_period" default:"10s"`
 
+	// ReadHeaderTimeout is how long each server waits for a client to send
+	// its request headers before closing the connection
+	ReadHeaderTimeout time.Duration `yaml:"read_header_timeout" mapstructure:"read_header_timeout" default:"10s"`
+
+	// IdleTimeout is how long each server keeps an idle keep-alive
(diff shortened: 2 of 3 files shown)


whoAbhishekSah:
Tested this live. Built the branch and ran server (connect 18002, metrics 19000, UI 18100) against the same local DB. Probed with raw TCP clients that stall mid-request or sit idle on a keep-alive connection.

| Test | Config | Expected | Got |
|---|---|---|---|
| Slow client never finishes headers | defaults | dropped at 10s | closed at 10.0s on all 3 servers |
| Slow client never finishes headers | `read_header_timeout: 3s` | dropped at 3s | closed at 3.0s on all 3 servers |
| Idle keep-alive connection | `idle_timeout: 6s` | closed at 6s | closed at 6.0s |
| Reuse connection after 2s idle | `idle_timeout: 6s` | second request works | worked |
| Reuse connection after 12s idle | defaults (180s) | connection survives | worked — idle timeout is not tied to the 10s header timeout |
| Normal RPC through connect | both configs | server answers | answered fine (401 without a cookie, as expected) |

Defaults fill in when the config file has no timeout keys, and setting `read_header_timeout` / `idle_timeout` under `app:` changes the behavior on all three servers.

AmanGIT07 (approved):

PR #1877, opened 2026-08-13
fix(server): recover from panics in RPC and UI HTTP handlers

whoAbhishekSah:
## Problem

Frontier has no panic recovery in the request path. When a handler panics, the only net is Go's built-in per-connection recover in `net/http`:

- The client's connection is dropped with no response at all (HTTP/1.1: empty reply; HTTP/2: `INTERNAL_ERROR` stream reset), instead of an RPC error.
- The stack trace goes through net/http's error logger. With slog set as the default logger, that lands as a single INFO-level line with the whole stack stuffed into `msg` — invisible to anything watching for error-level logs, and with no request context.

Reproduced by planting a deliberate `panic()` in a handler:

```
$ curl -v .../FrontierService/ListAuthStrategies ...
* Empty reply from server            # HTTP/1.1
* HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```

## Fix

- Pass `connect.WithRecover(...)` when building the Frontier and Admin service handlers. The recover callback logs the procedure, panic value, and stack through slog at error level, and returns `CodeInternal` with a generic message. It is placed before the interceptor chain so it sits outermost and also catches panics thrown from interceptors, not just handlers.
- Wrap the UI server mux (`/configs`, `/frontier-connect/` proxy, SPA routes) with an equivalent recovery handler that returns a plain 500, since `WithRecover` only covers connect handlers. It re-panics on `http.ErrAbortHandler`, matching net/http's own contract.
- No panic details reach the caller in either path.

Out of scope: panics on internal background goroutines (audit log listener, shutdown watchers) — those are not HTTP calls and need their own recover treatment separately.

## After the fix

Same deliberately panicking handler:

```
$ curl -s -w "\nHTTP %{http_code}\n" .../FrontierService/ListAuthStrategies ...
{"code":"internal","message":"internal server error"}
HTTP 500
```

and one structured log entry:

```
{"level":"ERROR","msg":"rpc handler panic","procedure":"/raystack.frontier.v1beta1.FrontierService/ListAuthStrategies","panic":"...","stack":"goroutine ..."}
```

## Tests

- A panicking connect handler returns HTTP 500 with `"code":"internal"`, no panic detail in the body, and an error-level log entry carrying the procedure and stack.
- The UI recovery wrapper returns a generic 500 on panic, passes normal requests through untouched, and re-panics on `http.ErrAbortHandler`.
- Verified live against a running server over both HTTP/1.1 and HTTP/2 (output above).

### DIFF
--- a/pkg/server/recovery.go
+++ b/pkg/server/recovery.go
@@ -23,10 +23,39 @@ func connectPanicRecovery(logger *slog.Logger) connect.HandlerOption {
 	})
 }
 
+// committedWriter tracks whether the response status line has gone out, so
+// the recovery handler knows whether a 500 can still be sent.
+type committedWriter struct {
+	http.ResponseWriter
+	committed bool
+}
+
+func (c *committedWriter) WriteHeader(statusCode int) {
+	c.committed = true
+	c.ResponseWriter.WriteHeader(statusCode)
+}
+
+func (c *committedWriter) Write(b []byte) (int, error) {
+	c.committed = true
+	return c.ResponseWriter.Write(b)
+}
+
+// Flush commits the response too. The connect reverse proxy on the UI mux
+// needs the Flusher interface for streaming.
+func (c *committedWriter) Flush() {
+	c.committed = true
+	if f, ok := c.ResponseWriter.(http.Flusher); ok {
+		f.Flush()
+	}
+}
+
 // uiPanicRecovery responds with a plain 500 instead of dropping the
-// connection when a handler behind the UI mux panics.
+// connection when a handler behind the UI mux panics. If the handler already
+// wrote part of a response, the status can no longer be changed, so it aborts
+// the connection instead of appending an error to the partial body.
 func uiPanicRecovery(logger *slog.Logger, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		cw := &committedWriter{ResponseWriter: w}
 		defer func() {
 			if panicValue := recover(); panicValue != nil {
 				// net/http checks for ErrAbortHandler with ==, so we should too.
@@ -37,9 +66,12 @@ func uiPanicRecovery(logger *slog.Logger, next http.Handler) http.Handler {
 					"path", r.URL.Path,
 					"panic", panicValue,
 					"stack", string(debug.Stack()))
-				http.Error(w, errors.ErrInternalServerError.Error(), http.StatusInternalServerError)
+				if cw.committed {
+					panic(http.ErrAbortHandler)
+				}
+				http.Error(cw, errors.ErrInternalServerError.Error(), http.StatusInternalServerError)
 			}
 		}()
-		next.ServeHTTP(w, r)
+		next.ServeHTTP(cw, r)
 	})
 }
--- a/pkg/server/recovery.go
+++ b/pkg/server/recovery.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"io"
 	"log/slog"
 	"net/http"
 	"runtime/debug"
@@ -40,7 +41,7 @@ func (c *committedWriter) Write(b []byte) (int, error) {
 	return c.ResponseWriter.Write(b)
 }
 
-// Flush commits the response too. The connect reverse proxy on the UI mux
+// Flush commits the response too. The connect reverse proxy on the UI server
 // needs the Flusher interface for streaming.
 func (c *committedWriter) Flush() {
 	c.committed = true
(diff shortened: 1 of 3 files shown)


pkg/server/recovery.go
```diff
@@ -0,0 +1,47 @@
...
+			if panicValue := recover(); panicValue != nil {
+				// net/http checks for ErrAbortHandler with ==, so we should too.
+				if panicValue == http.ErrAbortHandler {
+					panic(panicValue)
+				}
+				logger.ErrorContext(r.Context(), "http handler panic",
+					"path", r.URL.Path,
+					"panic", panicValue,
+					"stack", string(debug.Stack()))
+				http.Error(w, errInternalServer.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
whoAbhishekSah: Fixed in 094eb1f9. The recovery handler now wraps the ResponseWriter to track whether the response is committed. Before the first write it sends the 500 as before. After a write or flush the status can't be changed, so it logs the panic and re-panics with `http.ErrAbortHandler` — the connection is dropped and the client sees a truncated response instead of a 200 with error text appended. The wrapper forwards `Flush` so the reverse proxy's streaming keeps working; `Hijacker`/`Pusher` aren't needed on this mux (plain HTTP/1.1, no websocket routes). Added tests for write-then-panic and flush-then-panic asserting nothing is appended.

pkg/server/recovery.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (c *committedWriter) Write(b []byte) (int, error) {
+	c.committed = true
+	return c.ResponseWriter.Write(b)
+}
+
+// Flush commits the response too. The connect reverse proxy on the UI mux
+// needs the Flusher interface for streaming.
+func (c *committedWriter) Flush() {
```
rohilsurana, line 46: committedWriter only re-exposes Flush. It does not implement http.Hijacker and has no Unwrap method, so http.ResponseController on this writer cannot reach the underlying connection.

The UI mux hands committedWriter to the /frontier-connect/ ReverseProxy. If a request ever needs a protocol upgrade, ReverseProxy calls http.NewResponseController(rw).Hijack(). Since committedWriter is not a Hijacker and has no Unwrap, that returns ErrNotSupported and the proxy fails with `can't switch protocols using non-Hijacker ResponseWriter type`. The same missing Unwrap also blocks SetReadDeadline, SetWriteDeadline, and EnableFullDuplex, and turns FlushError into the error-less Flush, so flush and write errors on a broken client are dropped without notice.

One method fixes all of it, because http.ResponseController walks Unwrap to find Hijacker, the deadline setters, and FlushError:

```go
func (c *committedWriter) Unwrap() http.ResponseWriter { return c.ResponseWriter }
```

pkg/server/recovery.go
```diff
@@ -0,0 +1,77 @@
...
+	http.ResponseWriter
+	committed bool
+}
+
+func (c *committedWriter) WriteHeader(statusCode int) {
+	c.committed = true
+	c.ResponseWriter.WriteHeader(statusCode)
+}
+
+func (c *committedWriter) Write(b []byte) (int, error) {
```
rohilsurana, line 39: committedWriter does not forward io.ReaderFrom, so the SPA static file server behind the UI mux loses its sendfile fast path.

Before this change the SPA handler got the raw *http.response, which implements io.ReaderFrom, so ServeContent used sendfile for static assets. Now it gets committedWriter, which has no ReadFrom, so every admin UI asset is copied through a 32KB userspace buffer loop instead. Small but real cost on the busiest UI path.

Unwrap does not fix this one. io.Copy checks the destination for io.ReaderFrom directly and does not walk Unwrap, so it needs its own method:

```go
func (c *committedWriter) ReadFrom(r io.Reader) (int64, error) {
	c.committed = true
	if rf, ok := c.ResponseWriter.(io.ReaderFrom); ok {
		return rf.ReadFrom(r)
	}
	return io.Copy(c.ResponseWriter, r)
}
```

pkg/server/recovery.go
AmanGIT07: Agree on Unwrap — worth adding. One correction: Unwrap restores Hijack, the deadline setters, and EnableFullDuplex, but not FlushError. ResponseController.Flush checks FlushError, then Flusher, then unwraps; committedWriter implements Flush, so the lookup stops at the wrapper and returns nil either way. Getting flush errors back needs a FlushError method on the wrapper itself. The proxy discards flush errors anyway, so Unwrap alone seems enough.

pkg/server/recovery.go
AmanGIT07: The sendfile premise doesn't hold here. The admin SPA is served from embed.FS (web/apps/admin/embed.go, go:embed all:dist), and sendfile only applies when the copy source is a real OS file. Embedded assets were already copied through a userspace buffer before this change. The only delta is io.Copy allocating its 32KB buffer instead of using net/http's pooled one. The ReadFrom passthrough is still fine to add, but it's tidiness, not a lost fast path.

pkg/server/server.go
AmanGIT07: `WithRecover` only covers panics inside the two connect service handlers. The other routes on this mux still have none: the webhook bridge parses the path, reads the body, and marshals JSON before it reaches frontierHandler  (webhook_bridge.go), and /ping, health, reflection, and the CORS wrapper are bare too. A panic in any of those still drops the connection with the old logging — the behavior this PR removes everywhere else.                       

The UI server wraps its whole mux. Doing the same here makes the two servers consistent and also covers any future service handler someone registers without the recovery option:                                                    

suggestion                                                                                                                                                                                                                           
       handler := uiPanicRecovery(logger, connectinterceptors.WithConnectCORS(mux, cfg.ConnectCors))                                                                                                                                    
                                                                                                                                                                                                                                     
RPC panics are still handled by WithRecover first, so clients keep getting proper connect error codes; the outer wrapper only sees panics that escape it, and it already re-panics on http.ErrAbortHandler so the two nets don't fight. If it's reused here, uiPanicRecovery probably wants a generic name like httpPanicRecovery.

rohilsurana: should not be using `uiPanicRecovery`  name here.

pkg/server/recovery.go
whoAbhishekSah: Added Unwrap in 838970f1. ResponseController can now reach the underlying writer for Hijack, the deadline setters, and full duplex. Left FlushError alone per the thread below — the proxy discards flush errors anyway.

pkg/server/recovery.go
whoAbhishekSah: Added ReadFrom in 838970f1. As noted below, the admin assets come from embed.FS so there was no sendfile path to lose, but the passthrough keeps the underlying writer's copy path and it costs one small method. It also marks the response as committed, since bytes reached the wire. Added a test that io.Copy into the wrapper dispatches to it.

pkg/server/server.go
whoAbhishekSah: Done in 838970f1. The connect server mux is now wrapped the same way as the UI server, outside CORS, so the webhook bridge, ping, health, and reflection are covered too. WithRecover still turns RPC panics into connect error codes first; the outer wrapper only sees what escapes. Also renamed it to httpPanicRecovery since it fronts both servers now. Verified live that ping, a normal RPC, and the health check still respond through the wrapped mux.

AmanGIT07 (approved):

pkg/server/recovery_test.go
```diff
@@ -0,0 +1,203 @@
...
+	if cw.Unwrap() != http.ResponseWriter(rec) {
+		t.Errorf("Unwrap() = %v, want the underlying writer", cw.Unwrap())
+	}
+}
+
+func TestCommittedWriterReadFromCommitsAndCopies(t *testing.T) {
+	rec := httptest.NewRecorder()
+	cw := &committedWriter{ResponseWriter: rec}
+
+	n, err := io.Copy(cw, strings.NewReader("data"))
```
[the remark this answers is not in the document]
whoAbhishekSah: Fixed in f8913721. The test now calls cw.ReadFrom directly instead of going through io.Copy, so the method under test actually runs.

PR #1878, opened 2026-08-13
feat(metaschema): refresh the schema cache on a timer and make it concurrency-safe

rohilsurana:
## Summary

Metaschemas hold the JSON schemas that validate entity metadata for users, organizations, groups, and roles. The metaschema service keeps every schema in an in-memory map on the service. Today that map is primed once at boot and after that it only changes on the pod that handled a write. In a multi-pod deployment a schema change made through the API reaches one pod but not the others, so those pods keep validating against the old schema until they restart. The map also has no lock, so concurrent request goroutines on one pod can race on it, which Go turns into a "concurrent map read and map write" panic.

This PR makes the cache safe for concurrent access and reloads it on a timer, so every pod picks up a schema change within a small bounded window.

## Changes

- Guard the schema cache with a `sync.RWMutex` and switch every service method to a pointer receiver.
- Add a periodic reload that lists all schemas and swaps the cache, reusing the same `robfig/cron` pattern as the billing sync jobs.
- Add `Init` to prime the cache and start the job, and `Close` to stop it.
- Add the `app.metaschema.refresh_interval` config. It defaults to `1m`, and `0` disables the job for single-pod, local, and test runs.
- Wire `Init` and `Close` into server start and shutdown, and document the new setting.

## Technical Details

- Reads (`Get`, `List`, `Validate`) take a read lock. Writes (`Create`, `Update`, `Delete`, and the reload) take a write lock.
- The writing pod still updates its own cache right away, so its change is visible at once. The timer carries the change to the other pods within one interval.
- On a reload error the current cache is kept, so a database blip never blanks the schemas.
- `@every` schedules from pod start, so pods that start together reload in lockstep. For this tiny table that query cost is small. Jitter or Postgres `LISTEN/NOTIFY` are noted as future options if it ever matters.
- The receiver change is internal only. Method signatures do not change, so the `MetaSchemaService` interface and its generated mock stay valid.

## Test Plan

- [x] Build and type checking passes (`go build ./...`, `go vet`)
- [x] New unit tests in `core/metaschema` pass under `go test -race`, covering concurrent access alongside a reload, reload picking up a new schema, cache kept on a list error, and the refresh-disabled path

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -163,12 +163,16 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
 		return err
 	}
 
-	// load metadata schema in memory from db
-	if schemas, err := deps.MetaSchemaService.List(context.Background()); err != nil {
+	// prime the metaschema cache and start its periodic refresh
+	if err := deps.MetaSchemaService.Init(ctx); err != nil {
 		logger.Warn("metaschemas initialization failed", "err", err)
-	} else {
-		logger.Info("metaschemas loaded", "count", len(schemas))
 	}
+	defer func() {
+		logger.Debug("cleaning up metaschemas")
+		if err := deps.MetaSchemaService.Close(); err != nil {
+			logger.Warn("metaschema service cleanup failed", "err", err)
+		}
+	}()
 
 	// apply schema
 	if err = deps.BootstrapService.MigrateSchema(ctx); err != nil {
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -489,7 +489,7 @@ func buildAPIDependencies(
 	domainService := domain.NewService(logger, domainRepository, userService, organizationService, membershipService)
 
 	metaschemaRepository := postgres.NewMetaSchemaRepository(logger, dbc)
-	metaschemaService := metaschema.NewService(metaschemaRepository)
+	metaschemaService := metaschema.NewService(metaschemaRepository, logger, cfg.App.Metaschema.RefreshInterval)
 
 	userPATService := userpat.NewService(logger, userPATRepo, cfg.App.PAT, organizationService, roleService, membershipService, projectService, auditRecordRepository)
 	membershipService.SetUserPATService(userPATService)
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -163,9 +163,11 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
 		return err
 	}
 
-	// prime the metaschema cache and start its periodic refresh
+	// prime the metaschema cache and start its periodic refresh. A failed prime
+	// means validation would be silently off, so stop startup like a failed
+	// migration rather than come up with an empty cache.
 	if err := deps.MetaSchemaService.Init(ctx); err != nil {
-		logger.Warn("metaschemas initialization failed", "err", err)
+		return fmt.Errorf("metaschemas initialization: %w", err)
 	}
 	defer func() {
 		logger.Debug("cleaning up metaschemas")
--- a/core/metaschema/service.go
+++ b/core/metaschema/service.go
@@ -2,6 +2,7 @@ package metaschema
 
 import (
 	"context"
+	"sync"
 
 	"github.com/raystack/frontier/pkg/utils"
 
@@ -11,7 +12,9 @@ import (
 )
 
 type Service struct {
-	repository      Repository
+	repository Repository
+
+	mu              sync.RWMutex
 	metaSchemaCache map[string]MetaSchema
 }
 
@@ -22,17 +25,22 @@ func NewService(repository Repository) *Service {
 	}
 }
 
-func (s Service) Create(ctx context.Context, toCreate MetaSchema) (MetaSchema, error) {
+func (s *Service) Create(ctx context.Context, toCreate MetaSchema) (MetaSchema, error) {
 	mschema, err := s.repository.Create(ctx, toCreate)
 	if err != nil {
 		return MetaSchema{}, err
 	}
+	s.mu.Lock()
 	s.metaSchemaCache[mschema.Name] = mschema
+	s.mu.Unlock()
 	return mschema, nil
 }
 
-func (s Service) Get(ctx context.Context, idOrName string) (MetaSchema, error) {
-	if schema, ok := s.metaSchemaCache[idOrName]; ok {
+func (s *Service) Get(ctx context.Context, idOrName string) (MetaSchema, error) {
+	s.mu.RLock()
+	schema, ok := s.metaSchemaCache[idOrName]
+	s.mu.RUnlock()
+	if ok {
 		return schema, nil
 	}
 
@@ -41,24 +49,14 @@ func (s Service) Get(ctx context.Context, idOrName string) (MetaSchema, error) {
 		if err != nil {
 			return MetaSchema{}, err
 		}
-
 		return schema, nil
 	}
 	return MetaSchema{}, ErrInvalidID
 }
 
-func (s Service) List(ctx context.Context) ([]MetaSchema, error) {
-	if len(s.metaSchemaCache) == 0 {
-		schemas, err := s.repository.List(ctx)
-		if err != nil {
-			return nil, err
-		}
-		for _, schema := range schemas {
-			s.metaSchemaCache[schema.Name] = schema
-		}
-		return schemas, nil
-	}
-
+func (s *Service) List(ctx context.Context) ([]MetaSchema, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	schemas := make([]MetaSchema, 0, len(s.metaSchemaCache))
 	for _, schema := range s.metaSchemaCache {
 		schemas = append(schemas, schema)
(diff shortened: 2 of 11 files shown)


core/metaschema/service.go
```diff
@@ -115,3 +129,63 @@ func (s Service) Validate(mdata metadata.Metadata, name string) error {
...
+	if err != nil {
+		s.logger.WarnContext(ctx, "metaschema cache refresh failed", "err", err)
+		return
+	}
+	fresh := make(map[string]MetaSchema, len(schemas))
+	for _, schema := range schemas {
+		fresh[schema.Name] = schema
+	}
+	s.mu.Lock()
+	s.metaSchemaCache = fresh
```
rohilsurana, line 178: **Read-modify-write race can clobber a concurrent write.**

`reload()` reads the DB list at line 137 without the lock, then swaps in the whole map here under the lock. A write that lands in between gets wiped.

Timeline on the writing pod:
1. `reload` calls `repository.List` and gets a snapshot that does not include schema `C`.
2. A request runs `Create(C)`: the row commits and `s.metaSchemaCache[C] = C` is set under the lock.
3. `reload` takes the lock and overwrites the cache with its stale snapshot, so `C` disappears until the next refresh (up to `refresh_interval`, 1m by default).

During that window `Validate` for `C`'s entity type returns nil, so invalid metadata is accepted with no validation. The `Delete` case is the mirror image: it brings a just-deleted schema back and wrongly rejects valid metadata. This is the opposite of the PR's goal that the writing pod sees its own change at once.

One fix: merge the fresh set into the cache under a single lock (or re-read under the lock), instead of a blind full-map swap.

core/metaschema/service.go
```diff
@@ -115,3 +129,63 @@ func (s Service) Validate(mdata metadata.Metadata, name string) error {
 	}
 	return nil
 }
+
+// reload replaces the cache with the current set of metaschemas from the
+// database. On error it keeps the existing cache, so a database blip does not
+// blank validation.
+func (s *Service) reload(ctx context.Context) {
+	schemas, err := s.repository.List(ctx)
```
rohilsurana, line 167: **An empty but error-free List wipes the whole cache.**

If `repository.List` returns an empty slice with no error (read-replica lag, an isolation edge, or a repo bug), `reload` builds an empty `fresh` map and swaps it in at line 147. That drops every cached schema, and `Validate` then returns nil for all entity types until a later reload repopulates it. So validation is off platform-wide during that window.

The old code never blanked a populated cache from a `List` call. Consider a floor guard: skip the swap when the fresh set is empty but the current cache is not.

core/metaschema/service.go
```diff
@@ -41,65 +60,60 @@ func (s Service) Get(ctx context.Context, idOrName string) (MetaSchema, error) {
...
-		if err != nil {
-			return nil, err
-		}
-		for _, schema := range schemas {
-			s.metaSchemaCache[schema.Name] = schema
-		}
-		return schemas, nil
-	}
-
+func (s *Service) List(ctx context.Context) ([]MetaSchema, error) {
```
rohilsurana, line 88: **List no longer falls back to the DB, so a failed prime disables validation silently.**

The old `List` loaded from the DB whenever the cache was empty, so a transient DB failure at boot healed itself on the first request. Now `List` only reads the cache. If the priming `reload` in `Init` fails (a DB blip at startup), the cache stays empty, `List` returns an empty slice with no error, and `Validate` returns nil for every user, org, group, role and prospect create.

With `refresh_interval=0` (the documented single-pod, local and test setting) this never recovers until a `Create`/`Update`. With the default 1m it recovers only after the next reload, and every validation in that window is skipped with no error surfaced.

core/metaschema/service.go
```diff
@@ -115,3 +129,63 @@ func (s Service) Validate(mdata metadata.Metadata, name string) error {
...
+	s.mu.Lock()
+	s.metaSchemaCache = fresh
+	s.mu.Unlock()
+}
+
+// Init primes the cache from the database and, when refreshInterval is greater
+// than zero, starts a background job that reloads it on that interval so a
+// change made on one pod reaches the others.
+func (s *Service) Init(ctx context.Context) error {
+	s.reload(ctx)
```
rohilsurana: **Init swallows the priming error, so the boot-failure guard is dead.**

`reload` only logs its error and returns; it does not propagate it. So this priming call cannot fail `Init`. `Init` then logs `metaschemas loaded count=0` as if it succeeded and returns nil. The only error `Init` can return comes from `AddFunc`, which will not realistically fail for a valid duration.

In `cmd/serve.go` the guard `if err := deps.MetaSchemaService.Init(ctx); err != nil { ... }` therefore never fires for the common case, a DB that is unreachable at startup. Operators lose the clear boot-failure signal the old code produced from the `List` error, and boot proceeds with an empty cache. Consider having `reload` return its error and `Init` propagate it from the initial prime.

core/metaschema/service_test.go
```diff
@@ -0,0 +1,151 @@
...
+	name := f.byID[id].Name
+	delete(f.byID, id)
+	return name, nil
+}
+
+func (f *fakeRepo) MigrateDefaults(_ context.Context) error { return nil }
+
+func discardLogger() *slog.Logger { return slog.New(slog.NewTextHandler(io.Discard, nil)) }
+
+func TestService_ConcurrentAccess(t *testing.T) {
```
rohilsurana, line 140: **The concurrency fix the PR is built for is not actually tested.**

This test runs Create/Validate/List/Get concurrently, which catches map races under `-race`, but it never runs `reload()` alongside the writers. So the read-modify-write clobber flagged above is a logical race that `-race` cannot see, and no test covers it. A regression test would run `reload` in a loop while `Create`/`Delete` run, then assert the written schema is still present.

Also every test builds the service with interval 0, so the cron path (`AddFunc`/`Start`/`Stop`, `SkipIfStillRunning`, and `Close` waiting on `Stop().Done()`) never runs. The lifecycle the PR adds ships with no coverage.

rohilsurana (commented):
Automated review of the metaschema cache refresh. Findings 1 to 4 all trace back to one design choice: `reload()` does a blind full-cache swap. Left inline below, most serious first.

core/metaschema/service.go
```diff
@@ -115,3 +129,63 @@ func (s Service) Validate(mdata metadata.Metadata, name string) error {
...
+	s.mu.RLock()
+	count := len(s.metaSchemaCache)
+	s.mu.RUnlock()
+	s.logger.InfoContext(ctx, "metaschemas loaded", "count", count)
```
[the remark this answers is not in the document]
rohilsurana: Fixed in 7201eb20. `Init` now propagates the initial prime error, so startup stops instead of serving an empty cache with validation off. The scheduled refresh keeps the retain-and-log behavior for later failures, as you suggested.

core/metaschema/service.go
rohilsurana: Fixed in 7201eb20. `reload` now holds the write lock across the `List` and the swap, so a concurrent Create, Update, or Delete either shows up in the snapshot or re-applies to the cache after the swap. The writing pod never loses its own change. Added `TestService_reload_doesNotDropConcurrentWrite`, which orders a Create between reload's snapshot and its swap and asserts the schema survives. It fails on the old blind swap.

One tradeoff: the read now runs under the write lock, so Validate readers can wait on it. For this five-row table read once per interval, that wait is sub-millisecond and bounded by the query timeout. If you would rather readers never wait on a DB read, the alternative is a separate write-serialization mutex. Happy to switch if you prefer that.

core/metaschema/service.go
rohilsurana: Fixed in 7201eb20. `reload` now refuses to swap an empty set over a populated cache: when `List` returns no rows but the cache has entries, it returns an error and keeps the cache. Added `TestService_reload_keepsCacheOnEmptyList`.

core/metaschema/service.go
rohilsurana: Handled by the `Init` fix in 7201eb20 rather than by restoring the lazy read. The real risk here is a failed prime leaving validation off. `Init` now returns that error and boot fails, so the server never starts on an empty cache from a startup DB blip. Also note Validate reads the cache directly and never used List's old DB fallback, so bringing the fallback back would not have covered the Validate path. I kept List cache-only so the cache has a single owner. If you still want a lazy read for the ListMetaSchemas endpoint on its own, I can add it.

core/metaschema/service.go
rohilsurana: Fixed in 7201eb20. `reload` returns its error and `Init` propagates it from the initial prime, so the `cmd/serve.go` guard fires when the DB is unreachable at startup. Added `TestService_Init_returnsPrimeError`.

core/metaschema/service_test.go
rohilsurana: Fixed in 7201eb20. Two tests added. `TestService_reload_doesNotDropConcurrentWrite` reproduces the read-modify-write clobber with an orchestrated List hook, not just `-race`. `TestService_Init_runsScheduledRefresh` runs the service with a real interval and exercises the cron path: AddFunc and Start, a scheduled refresh firing, and Close stopping it.

cmd/serve.go
```diff
@@ -163,12 +163,16 @@ func StartServer(logger *slog.Logger, cfg *config.Frontier) error {
 		return err
 	}
 
-	// load metadata schema in memory from db
-	if schemas, err := deps.MetaSchemaService.List(context.Background()); err != nil {
+	// prime the metaschema cache and start its periodic refresh
+	if err := deps.MetaSchemaService.Init(ctx); err != nil {
 		logger.Warn("metaschemas initialization failed", "err", err)
```
whoAbhishekSah: `Init` returns an error now when the first cache load fails, but the caller only logs a warning and moves on. If the DB is unreachable for a moment at boot, the server comes up with an empty cache and `Validate` passes everything without checking. With `refresh_interval: 0` there is no later reload, so it stays that way until the process restarts. Nothing in the logs says validation is off — just one warn line at startup.

Two ways to close it:

1. Treat it as a startup failure:

```go
if err := deps.MetaSchemaService.Init(ctx); err != nil {
    return fmt.Errorf("metaschemas initialization: %w", err)
}
```

If the schemas can't be loaded, the server fails to start, same as a failed migration a few lines below. This also matches the comment inside `Init` that says "startup stops here" — right now the caller doesn't actually stop.

2. Fetch on demand — when a read finds the cache empty, fall back to `repository.List` and fill it. This is what the code did before this PR, so a failed load at boot heals on the first request that needs a schema. The cost is keeping the fallback path this PR deliberately removed, and until that first successful fetch, writes go through without validation.

I'd go with 1 — it's one line and there's no window where validation is silently off.

whoAbhishekSah (approved):

cmd/serve.go
rohilsurana: Fixed in 9b062b0f with option 1. `cmd/serve.go` now returns `fmt.Errorf("metaschemas initialization: %w", err)` when `Init` fails, so the server stops at startup instead of coming up with an empty cache and validation silently off, same as the failed-migration path just below. Thanks for catching that the caller was not actually stopping.

PR #1880, opened 2026-08-13
feat(deleter): email the org owners when a delete forfeits tokens

whoAbhishekSah:
Part of https://github.com/raystack/frontier/issues/1837. Stacked on #1857; the check RPC follows in #1893/#1894.

Deleting an organization forfeits any unused tokens (the client confirms that with the user before calling). This PR adds the second half of that flow: the org owners get an email saying how many tokens were left and that support can transfer the purchased amount to their bank account.

How it works:

- The email subject and body come from new config keys `billing.token_forfeit_notice.subject` and `billing.token_forfeit_notice.body`. Both are Go templates with `.Amount` (tokens forfeited), `.Purchased` (the share of it that was bought), `.Org` (the deleted organization), `.User` (the owner receiving the mail), and `.DeletedBy` (who ran the delete, when known). Empty config falls back to plain built-in text. This is the same pattern the PAT expiry alert emails use.
- **Only the purchased share is transferable.** Every token grant is recorded in `billing_transactions` with a source (`system.buy` for purchases, `system.starter` for plan starter grants, `system.awarded` for awards). Spending is not attributed to a source, so complimentary tokens are counted as spent first: the purchased share is the smaller of the balance and everything ever bought. Accounts holding starter or awarded grants get the split; when everything was purchased the share equals the whole amount and the email shows one number.
- The forfeit audit record (`app.billing.tokens.forfeited`) now carries `purchased` next to `amount` — the transaction rows are deleted with the org, so the audit row is what support settles from later.
- The owners are resolved through the membership service (the same path `ListOrganizationUsers` uses) and every owner gets the mail, through the shared `app.mailer` dialer.
- The owners and the amounts are read before teardown starts (both are gone after), but the mail is only sent after the delete fully succeeded. A failed delete never emails anyone; repeated calls on a deleted org return `not_found` first, so nobody gets the mail twice. A retry that lost the balances mid-teardown recovers the amounts from the forfeit audit records the earlier attempt wrote (needs the `db` audit sink). Send and render failures are logged per owner and never fail the API call.
- The owner lookup is best-effort: the delete never depends on the policy machinery, and a forfeit with no reachable owner logs a warning instead of silently mailing no one.
- The balances are read once and reused by the blocker check and the teardown audit, so the audit numbers always match what the pre-check saw.

Side change: the org lookup at the start of the delete switched from `Get` to `GetRaw`, so a disabled org keeps its title for the email and stays deletable without the `ErrDisabled` special case.


core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,164 @@
...
+	// User is the owner receiving this mail.
+	User user.User
+	// Org is the deleted organization.
+	Org organization.Organization
+}
+
+// forfeitNotice is everything sendForfeitNotices needs once the org is gone.
+// It has to be collected before teardown removes the owners and the token
+// balances.
+type forfeitNotice struct {
```
whoAbhishekSah, line 55: there are complimentary tokens could be given with each plan. the return must not include that. Have we ensured that ?

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if total == 0 {
+		return forfeitNotice{}, nil
+	}
+
+	ownerRole, err := d.roleService.Get(ctx, schema.RoleOrganizationOwner)
+	if err != nil {
+		return forfeitNotice{}, fmt.Errorf("failed to resolve the organization owner role: %w", err)
+	}
+	policies, err := d.policyService.List(ctx, policy.Filter{
```
whoAbhishekSah: can we use membership svc to list org owners ? 
all org owners should ge this email.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return forfeitNotice{}, fmt.Errorf("failed to fetch the organization owners: %w", err)
+	}
+	return forfeitNotice{Amount: total, Owners: owners}, nil
+}
+
+// sendForfeitNotices emails every org owner that the delete forfeited unused
+// tokens and that support can transfer the amount. The org is already gone at
+// this point, so failures are logged and never returned.
+func (d Service) sendForfeitNotices(ctx context.Context, org organization.Organization, notice forfeitNotice) {
```
whoAbhishekSah, line 228: are we recording who deleted the org in the email ?

core/deleter/forfeit_notice.go
whoAbhishekSah: We cannot split the remaining balance into purchased and complimentary here: the ledger records where tokens came from (system.buy, system.awarded, system.starter) but does not tie spending to a source, so there is no reliable way to say which part of what is left was paid for. What I changed instead: the email no longer promises the whole amount back — it now says the amount that was purchased can be transferred, and support works that share out from the payment records (Stripe keeps every purchase). Added a comment on collectForfeitNotice saying the same.

whoAbhishekSah: Correction to my earlier reply: this is computable after all. Every grant is recorded in billing_transactions with a source (system.buy, system.starter, system.awarded) — only spending is unattributed. So with complimentary tokens counted as spent first, the purchased share is min(balance, sum of system.buy credits). Implemented now: the email reports the purchased share and promises the transfer only for it, and the forfeit audit record carries purchased next to amount so support can settle from it after the transaction rows are gone. On deployments whose plans grant no free tokens the share equals the whole amount; accounts holding older starter or awarded grants get the split.

core/deleter/service.go
```diff
@@ -332,10 +357,24 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
 	if err := audit.NewLogger(ctx, id).Log(audit.OrgDeletedEvent, audit.OrgTarget(id)); err != nil {
 		slog.WarnContext(ctx, "failed to write audit log", "error", err, "event", audit.OrgDeletedEvent)
 	}
+
+	// the org is gone; tell the owners about any tokens the delete forfeited
+	if notice.Amount > 0 {
```
rohilsurana, line 377: The forfeit email is sent here at the very end, but notice.Amount came from a balance read taken before teardown, and the token rows are deleted mid-way. The doc says a failed delete can just be run again. If the delete fails after billing teardown but before this line, the retry runs collectForfeitNotice against the now-deleted balances, gets 0, so notice.Amount is 0 and no email goes out, even though the tokens were already forfeited and audited on the first attempt. So the email is silently lost on exactly the retry path the delete is built around. One fix is to drive the email off the audit record (which survives) rather than the live balance.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,256 @@
...
+			total += balance
+			purchased += bought
+			accounts[c.ID] = accountTokens{Balance: balance, Purchased: bought}
+		}
+	}
+	if total == 0 {
+		return forfeitNotice{Accounts: accounts}, nil
+	}
+
+	ownerRole, err := d.roleService.Get(ctx, schema.RoleOrganizationOwner)
```
rohilsurana, line 110: Resolving the owner role, listing owner principals, and fetching the owners are all fatal here, so a flaky SpiceDB or policy query, or a misconfigured owner role, now blocks the whole org delete. Before this change the delete never depended on owner lookup. Since a delete is often a compliance action, it may be better to treat a failed owner lookup as "skip the email and log it" than as "abort the delete".

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	ownerIDs := make([]string, 0, len(members))
+	for _, m := range members {
+		ownerIDs = append(ownerIDs, m.PrincipalID)
+	}
+	owners, err := d.userService.GetByIDs(ctx, ownerIDs)
+	if err != nil {
+		return forfeitNotice{}, fmt.Errorf("failed to fetch the organization owners: %w", err)
+	}
+	return forfeitNotice{Amount: total, Purchased: purchased, Accounts: accounts, Owners: owners}, nil
```
rohilsurana: If the org has unused tokens but no active owners (ownership held through a group, owners already removed, or every owner disabled, since GetByIDs filters disabled users out), notice.Owners is empty and sendForfeitNotices emails no one, while the tokens are still forfeited. Nothing warns that the notice went nowhere. A warn log when Amount > 0 but there are no owners would make this visible.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,256 @@
...
+	if subjectTpl == "" {
+		subjectTpl = defaultForfeitNoticeSubject
+	}
+	bodyTpl := d.forfeitNoticeConfig.Body
+	if bodyTpl == "" {
+		bodyTpl = defaultForfeitNoticeBody
+	}
+
+	deletedBy := deletedByFromContext(ctx)
+	for _, owner := range notice.Owners {
```
rohilsurana, line 258: Two small things in this loop. (1) A render error does return, so if rendering fails for the first owner the rest get no email, while a send error only does continue. The two error paths are inconsistent. (2) The subject and body templates are parsed on every iteration. Parse them once before the loop and execute per owner.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,256 @@
...
+		bought = 0
+	}
+	return min(bought, balance), nil
+}
+
+// sendForfeitNotices emails every org owner that the delete forfeited unused
+// tokens and that support can transfer the amount. The org is already gone at
+// this point, so failures are logged and never returned.
+func (d Service) sendForfeitNotices(ctx context.Context, org organization.Organization, notice forfeitNotice) {
+	if d.mailDialer == nil {
```
rohilsurana, line 229: cmd/serve.go always sets mailDialer (a mock dialer when SMTP is not configured), so this nil check never fires in practice. With the mock dialer, DialAndSend is a no-op that returns nil, so the code below logs "sent token forfeit notice" even though nothing was delivered. This matches the pattern used elsewhere so it is low priority, but the success log against a mock dialer can read as a false record.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,256 @@
...
+	owners, err := d.userService.GetByIDs(ctx, ownerIDs)
+	if err != nil {
+		return forfeitNotice{}, fmt.Errorf("failed to fetch the organization owners: %w", err)
+	}
+	return forfeitNotice{Amount: total, Purchased: purchased, Accounts: accounts, Owners: owners}, nil
+}
+
+// accountTokens returns one account's balance and purchased share, from the
+// caller's already-collected amounts when given, otherwise read fresh.
+func (d Service) accountTokens(ctx context.Context, accountID string, amounts map[string]accountTokens) (accountTokens, error) {
```
rohilsurana: The method accountTokens has the same name as the accountTokens type it returns. It compiles, but reading d.accountTokens(...) (accountTokens, error) makes a reader stop to tell the two apart. A small rename, e.g. resolveAccountTokens, clears it up.

rohilsurana (commented):
Stack-aware review (base org-delete-preflight-blockers). The retry issue in the first comment is the one worth fixing; the rest are smaller robustness and cleanup notes. One efficiency note that spans functions: customerService.List(OrgID) now runs three times per delete (ensureDeletable, collectForfeitNotice, deleteCustomers) and GetBalance runs twice per account, where one list and one balance read per account would do.

core/deleter/service.go
whoAbhishekSah: Fixed with the recovery you suggested: when the live balances show nothing, the delete rebuilds the amounts from the app.billing.tokens.forfeited audit records the earlier attempt wrote, and resolves the owners at that point (they still exist — identity teardown comes after billing). One limit worth knowing: this only works when the audit sink is a readable store (log.audit_events: db); with the stdout or noop sink there is nothing to read back and the retry stays email-less as before.

core/deleter/forfeit_notice.go
whoAbhishekSah: Left as is: cmd/serve.go always provides a dialer, and the mock-dialer no-op success is the same behavior the PAT expiry alerts and the OTP mail have today. Special-casing the log line only here would make this one mail path behave differently from the rest; if we want honest logs for unconfigured SMTP, that is a change for the mailer package across all senders.

whoAbhishekSah:
On the review-body efficiency note: fixed across the stack. DeleteOrganization now lists the billing accounts once and passes them down (blocker check, notice collection, teardown), the balances are read once in the notice collection and reused by the blocker check and the forfeit audit, and the plan lookups share one per-delete cache between the blocker pass and the cancel pass. The exported DeleteCustomers still reads for itself when called directly.

rohilsurana (commented):
Re-review after the fix commits. The retry recovery, the non-fatal owner lookup, the no-owner warn, parse-once, and the rename all look good. The new recoverForfeitFromAudit has a couple of reconciliation gaps worth a look, and one owner case is still uncovered.

core/deleter/service.go
```diff
@@ -242,14 +260,27 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
...
+	// the token forfeit notice reads owners and balances, so it has to be
+	// collected while they still exist; its balance reads are reused by the
+	// blocker check and the teardown audit below
+	notice, err := d.collectForfeitNotice(ctx, org, customers)
+	if err != nil {
+		return err
+	}
+	// a retry that lost the balances mid-teardown recovers the amounts from
+	// the audit records the earlier attempt wrote
+	if notice.Amount == 0 {
```
rohilsurana: Recovery only runs when the whole org's live balance is 0. On a multi-account org whose teardown was interrupted between accounts, the retry still sees a positive balance on the accounts not yet torn down, so recovery is skipped and the email reports only those, silently dropping the tokens already forfeited on the earlier accounts. This probably wants to be per-account (recover an account that reads 0 now but has a forfeit record) rather than all-or-nothing on the org total. Minor aside: for the common no-token org this still runs an extra audit query on every delete.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	return owners
+}
+
+// recoverForfeitFromAudit rebuilds the forfeit amounts from the audit
+// records a failed earlier attempt wrote before it deleted the balances, so
+// the retry that completes the delete still sends the owner notice. It runs
+// only when the live balances showed nothing and is best-effort: without a
+// readable audit store the notice simply stays empty.
+func (d Service) recoverForfeitFromAudit(ctx context.Context, orgID string, notice *forfeitNotice) {
```
rohilsurana, line 142: recoverForfeitFromAudit sums every BillingTokensForfeitedEvent for the org with no dedup and no time bound. Two things follow. (1) The forfeit audit is written before DeleteByAccountID, so a retry that re-forfeits the same still-funded account writes a second record, and this sum then double-counts. (2) A forfeit written by an earlier, separate teardown (for example the exported DeleteCustomers path) is picked up here and re-emailed when the org is later deleted. Reconciling against which accounts still hold a live balance, or marking records as consumed, would avoid both.

core/deleter/forfeit_notice.go
```diff
@@ -0,0 +1,288 @@
...
+// best-effort: the notice email must not make the delete depend on the
+// policy machinery, so a failed lookup logs and returns no owners.
+func (d Service) resolveOwners(ctx context.Context, orgID string) []user.User {
+	ownerRole, err := d.roleService.Get(ctx, schema.RoleOrganizationOwner)
+	if err != nil {
+		slog.WarnContext(ctx, "failed to resolve the organization owner role for the forfeit notice", "org_id", orgID, "error", err)
+		return nil
+	}
+	members, err := d.membershipService.ListPrincipalsByResource(ctx, orgID, schema.OrganizationNamespace, membership.MemberFilter{
+		PrincipalType: schema.UserPrincipal,
```
rohilsurana, line 116: resolveOwners lists only user-principal owners. If the owner role is held through a group, this returns no members, the new warn fires, and no one is emailed even though tokens were forfeited. If group-held ownership is expected, expanding to the group's members would close that gap. If not, at least the warn makes it visible now.

core/deleter/service.go
whoAbhishekSah: Reworked to per-account reconciliation as suggested: recovery now runs on every delete, takes the newest forfeit record per billing account, and skips accounts that still hold a live balance (the collection pass already counts those). A multi-account org interrupted mid-teardown now reports both the live and the already-forfeited accounts in one email. On the aside: yes, this is one indexed audit query per delete now — deletes are rare enough that I kept it unconditional rather than guessing when it can be skipped.

core/deleter/forfeit_notice.go
whoAbhishekSah: Both closed by the same reconciliation: newest record per account dedups the re-forfeit double-count, and live-balance accounts are skipped. One deliberate choice on your case (2): a forfeit written by an earlier standalone DeleteCustomers run IS included when the org is later deleted — no email went out at forfeit time (only the org delete sends one), so this is the last chance to tell the owners at all. The audit rows stay the source of truth either way.

core/deleter/forfeit_notice.go
whoAbhishekSah: Left as the warn: groups are being deprecated and we are not adding group-aware behavior in new code, so owner-through-group stays uncovered on purpose. The warn plus the audit record keep the forfeited amount visible for support when that configuration shows up.

rohilsurana (approved):

PR #1881, opened 2026-08-13
feat(sdk): guard the delete organization flow

whoAbhishekSah:
Part of https://github.com/raystack/frontier/issues/1837 — the client half of the delete pre-flight. The server stack (#1865–#1894) is merged; this PR now contains all the SDK changes in one place.

Guards in the client SDK's General settings view:

- **The delete button asks the server whether a delete would go through.** It calls the new `CheckOrganizationDelete` RPC (shipped in #1894, `@raystack/proton` bumped to the proton main commit that carries it). The button greys out for every blocker the real delete would refuse with — a paid subscription to downgrade, unpaid invoices, a token debt — and the hover tooltip shows one short instruction per kind of blocker ("Downgrade the subscription to the standard plan", "Pay the 2 open invoices from the billing page", "Contact support to settle the token balance"). A blocker kind this version does not know shows one generic line instead — deliberately not the server’s own message, which carries entity ids the UI must not surface. While the check loads the button stays disabled; a failed check fails open, since the server refuses a blocked delete independently.
- **The delete dialog warns about remaining tokens.** The balance is fetched only while the dialog is open, and the confirm button waits for the answer so a slow response cannot skip the warning. The dialog says the organization still has unused tokens, that deleting forfeits them, and that the support team will reach out to settle any that were purchased. It deliberately shows no amounts and promises nothing for complimentary tokens. Confirming the dialog is the user's consent — the server does not block on tokens.
- **Idempotent confirm.** The dialog's delete button disables while the request is in flight, so repeated clicks cannot fire the delete twice. A `failed_precondition` response still surfaces the server's reasons in the error toast for anyone who reaches the server anyway.
- The open-invoice filters used by the billing page's payment-issue banner now live in one shared helper (`client/utils/invoice-queries.ts`).

### DIFF
--- a/web/sdk/client/views/general/components/delete-organization-dialog.tsx
+++ b/web/sdk/client/views/general/components/delete-organization-dialog.tsx
@@ -45,9 +45,10 @@ export const DeleteOrganizationDialog = ({
   const orgLabel = t.organization({ case: 'capital' });
   const orgLabelLower = t.organization({ case: 'lower' });
   const [isAcknowledged, setIsAcknowledged] = useState(false);
-  // fetched only while the dialog is open; the confirm button waits for the
-  // answer so the forfeit warning cannot be skipped by a slow response
-  const { tokenBalance, isTokensLoading } = useTokens({ enabled: open });
+  // The balance is fetched only while the dialog is open. The confirm
+  // button below stays disabled until the fetch finishes, so the user
+  // cannot confirm before the token warning had a chance to appear.
+  const { tokenBalance, isTokensFetching } = useTokens({ enabled: open });
 
   const { mutateAsync: deleteOrganization } = useMutation(
     FrontierServiceQueries.deleteOrganization
@@ -163,7 +164,7 @@ export const DeleteOrganizationDialog = ({
                 !deleteTitle ||
                 !isAcknowledged ||
                 isSubmitting ||
-                isTokensLoading
+                isTokensFetching
               }
               data-test-id="frontier-sdk-delete-organization-btn"
               loading={isSubmitting}
--- a/web/sdk/client/views/general/components/delete-organization-dialog.tsx
+++ b/web/sdk/client/views/general/components/delete-organization-dialog.tsx
@@ -113,10 +113,9 @@ export const DeleteOrganizationDialog = ({
               </Text>
               {tokenBalance > 0 ? (
                 <Text size="small" variant="danger">
-                  You have {tokenBalance.toString()} tokens remaining. Deleting
-                  the {orgLabelLower} forfeits them. Please contact support
-                  about these tokens: any amount that was purchased can be
-                  transferred to your bank account.
+                  This {orgLabelLower} still has unused tokens, and deleting
+                  it forfeits them. If any of them were purchased, our
+                  support team will reach out to you to settle them.
                 </Text>
               ) : null}
               <Field
--- a/web/sdk/client/views/general/components/delete-organization-dialog.tsx
+++ b/web/sdk/client/views/general/components/delete-organization-dialog.tsx
@@ -110,8 +110,8 @@ export const DeleteOrganizationDialog = ({
               {tokenBalance > 0 ? (
                 <Text size="small" variant="danger">
                   You have {tokenBalance.toString()} tokens remaining. Deleting
-                  the {orgLabelLower} forfeits them. Contact support about
-                  these tokens: any amount that was purchased can be
+                  the {orgLabelLower} forfeits them. Please contact support
+                  about these tokens: any amount that was purchased can be
                   transferred to your bank account.
                 </Text>
               ) : null}
--- a/web/sdk/client/views/general/components/delete-organization-dialog.tsx
+++ b/web/sdk/client/views/general/components/delete-organization-dialog.tsx
@@ -21,6 +21,7 @@ import {
 import { useFrontier } from '../../../contexts/FrontierContext';
 import { useTerminology } from '../../../hooks/useTerminology';
 import { useTokens } from '../../../hooks/useTokens';
+import { deleteBlockedDescription } from '../../../utils/delete-blockers';
 import { handleConnectError } from '~/utils/error';
 
 const deleteOrgSchema = yup
@@ -88,9 +89,11 @@ export const DeleteOrganizationDialog = ({
     } catch (error) {
       handleConnectError(error, {
         PermissionDenied: () => toastManager.add({ title: "You don't have permission to perform this action", type: 'error' }),
-        FailedPrecondition: (err) => toastManager.add({ title: `Cannot delete this ${orgLabelLower} yet`, description: err.message, type: 'error' }),
-        NotFound: (err) => toastManager.add({ title: 'Not found', description: err.message, type: 'error' }),
-        Default: (err) => toastManager.add({ title: 'Something went wrong', description: err.message, type: 'error' }),
+        // the server names what blocks the delete; show the matching
+        // instructions instead of the raw server text
+        FailedPrecondition: (err) => toastManager.add({ title: `Cannot delete this ${orgLabelLower} yet`, description: deleteBlockedDescription(err), type: 'error' }),
+        NotFound: () => toastManager.add({ title: 'Not found', description: `This ${orgLabelLower} no longer exists.`, type: 'error' }),
+        Default: () => toastManager.add({ title: 'Something went wrong', description: 'Please try again later or contact support.', type: 'error' }),
       });
     }
   }
--- a/web/sdk/client/views/general/components/delete-organization-dialog.tsx
+++ b/web/sdk/client/views/general/components/delete-organization-dialog.tsx
@@ -20,6 +20,7 @@ import {
 } from '@raystack/apsara';
 import { useFrontier } from '../../../contexts/FrontierContext';
 import { useTerminology } from '../../../hooks/useTerminology';
+import { useTokens } from '../../../hooks/useTokens';
 import { handleConnectError } from '~/utils/error';
 
 const deleteOrgSchema = yup
@@ -44,6 +45,7 @@ export const DeleteOrganizationDialog = ({
   const orgLabel = t.organization({ case: 'capital' });
   const orgLabelLower = t.organization({ case: 'lower' });
   const [isAcknowledged, setIsAcknowledged] = useState(false);
+  const { tokenBalance } = useTokens();
 
   const { mutateAsync: deleteOrganization } = useMutation(
     FrontierServiceQueries.deleteOrganization
@@ -83,6 +85,7 @@ export const DeleteOrganizationDialog = ({
     } catch (error) {
       handleConnectError(error, {
         PermissionDenied: () => toastManager.add({ title: "You don't have permission to perform this action", type: 'error' }),
+        FailedPrecondition: (err) => toastManager.add({ title: `Cannot delete this ${orgLabelLower} yet`, description: err.rawMessage, type: 'error' }),
         NotFound: (err) => toastManager.add({ title: 'Not found', description: err.message, type: 'error' }),
         Default: (err) => toastManager.add({ title: 'Something went wrong', description: err.message, type: 'error' }),
       });
@@ -102,6 +105,13 @@ export const DeleteOrganizationDialog = ({
                 This action can not be undone. This will permanently
                 delete all the projects and resources in {organization?.title}.
               </Text>
+              {tokenBalance > 0 ? (
+                <Text size="small" variant="danger">
+                  You have {tokenBalance.toString()} tokens remaining. Deleting
+                  the {orgLabelLower} forfeits them. Contact support to get the
+                  amount transferred to your bank account.
+                </Text>
+              ) : null}
               <Field
                 label={`Please type name of the ${orgLabel} to confirm.`}
                 error={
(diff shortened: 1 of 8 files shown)


web/sdk/client/views/general/components/delete-organization-dialog.tsx
```diff
@@ -44,6 +45,7 @@ export const DeleteOrganizationDialog = ({
   const orgLabel = t.organization({ case: 'capital' });
   const orgLabelLower = t.organization({ case: 'lower' });
   const [isAcknowledged, setIsAcknowledged] = useState(false);
+  const { tokenBalance } = useTokens();
```
rohilsurana: The parent renders this dialog whenever canDeleteWorkspace is true, not only when it is open, so useTokens() runs getBillingBalance on every General settings load for delete-capable users, not just when they open the delete dialog. For an org whose billing account has no balance support, that query errors and pops an unrelated toast, and it adds a request to every settings visit. Gate the balance query on the dialog being open.

web/sdk/client/views/general/general-view.tsx
```diff
@@ -97,6 +123,14 @@ export function GeneralView({ onDeleteSuccess, urlPrefix }: GeneralViewProps = {
 
   const isLoading = !organization?.id || isActiveOrganizationLoading || isPermissionsFetching;
 
+  const { invoices } = useOrganizationInvoices({
+    query: OPEN_INVOICES_QUERY,
+    enabled: canDeleteWorkspace && !!organization?.id
+  });
+  const hasUnpaidInvoices = invoices.some(
```
rohilsurana: Two issues here. (1) While the invoices query is loading, and when it errors (the error is not read), invoices is [] so hasUnpaidInvoices is false and the delete button is enabled, which is the opposite of the intended grey-out. A user with open invoices can reach the server before the guard applies. (2) .some(inv => inv.state === OPEN) re-checks only state and drops the amount > 0 condition the query applied, so a zero-amount or fully-credited open invoice would still flag as unpaid. If you trust the server filter, invoices.length > 0 is enough; if not, re-check the amount too.

web/sdk/client/views/general/components/delete-organization-dialog.tsx
```diff
@@ -102,6 +105,13 @@ export const DeleteOrganizationDialog = ({
                 This action can not be undone. This will permanently
                 delete all the projects and resources in {organization?.title}.
               </Text>
+              {tokenBalance > 0 ? (
+                <Text size="small" variant="danger">
+                  You have {tokenBalance.toString()} tokens remaining. Deleting
+                  the {orgLabelLower} forfeits them. Contact support to get the
+                  amount transferred to your bank account.
```
rohilsurana: This tells the user to contact support to transfer the amount to their bank for any positive balance. But the backend in #1880 separates purchased tokens from complimentary ones, and only the purchased share is transferable. Its own email even says "These were complimentary tokens, so there is no amount to transfer." Here a user with free or promotional credits is told they can get a bank transfer, which will produce wrong support requests. Match the backend wording, or only promise a transfer for the purchased share.

web/sdk/client/views/general/components/delete-organization-dialog.tsx
```diff
@@ -83,6 +85,7 @@ export const DeleteOrganizationDialog = ({
     } catch (error) {
       handleConnectError(error, {
         PermissionDenied: () => toastManager.add({ title: "You don't have permission to perform this action", type: 'error' }),
+        FailedPrecondition: (err) => toastManager.add({ title: `Cannot delete this ${orgLabelLower} yet`, description: err.rawMessage, type: 'error' }),
```
rohilsurana: This handler uses err.rawMessage for the description, while the NotFound and Default handlers below use err.message. rawMessage is empty when the server sends no message, so the toast can show a blank description. Use err.message for consistency.

web/sdk/client/views/general/general-view.tsx
```diff
@@ -47,6 +53,26 @@ const generalSchema = yup
 
 type FormData = yup.InferType<typeof generalSchema>;
 
+// Open invoices with a non-zero amount. The server refuses the delete while
+// any exist, so the delete button greys out and explains why.
+const OPEN_INVOICES_QUERY = create(RQLRequestSchema, {
```
rohilsurana: This OPEN_INVOICES_QUERY is a verbatim copy of the one in billing/components/payment-issue.tsx, so the two can drift apart if one changes. It also fetches up to the default page size and sorts by created_at just to run .some() for a boolean. limit 1 with no sort answers "any open invoice?" with a smaller query. Consider sharing one constant and trimming it for this existence check.

rohilsurana (commented):
Review against main. The guard follows the existing payment-issue pattern well. A few correctness and UX notes; the token-transfer wording is the one I would fix before this ships.

web/sdk/client/views/general/components/delete-organization-dialog.tsx
```diff
@@ -102,6 +105,13 @@ export const DeleteOrganizationDialog = ({
                 This action can not be undone. This will permanently
                 delete all the projects and resources in {organization?.title}.
               </Text>
+              {tokenBalance > 0 ? (
+                <Text size="small" variant="danger">
+                  You have {tokenBalance.toString()} tokens remaining. Deleting
+                  the {orgLabelLower} forfeits them. Contact support to get the
+                  amount transferred to your bank account.
+                </Text>
+              ) : null}
```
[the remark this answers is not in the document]
whoAbhishekSah: Partly taken: the balance now only fetches while the dialog is open, and the confirm button stays disabled until the answer arrives, so a slow response can no longer skip the forfeit warning. On a failed balance request the delete stays allowed on purpose: tokens never block the delete server-side, the hook already surfaces the failure as a toast, and the owners still get the forfeit notice email with the exact amounts — blocking a compliance-grade action on a read-only status query seemed worse than proceeding informed-by-toast.

web/sdk/client/views/general/general-view.tsx
```diff
@@ -97,6 +123,14 @@ export function GeneralView({ onDeleteSuccess, urlPrefix }: GeneralViewProps = {
 
   const isLoading = !organization?.id || isActiveOrganizationLoading || isPermissionsFetching;
 
+  const { invoices } = useOrganizationInvoices({
+    query: OPEN_INVOICES_QUERY,
+    enabled: canDeleteWorkspace && !!organization?.id
+  });
+  const hasUnpaidInvoices = invoices.some(
+    inv => inv.state === INVOICE_STATES.OPEN
+  );
```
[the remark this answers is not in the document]
whoAbhishekSah: Loading now fails closed: the button is disabled until the unpaid-invoice answer arrives. A failed query fails open on purpose — the server independently refuses a delete with unpaid invoices (failed_precondition with the reasons shown in the dialog's toast), so the guard here is UX, not enforcement, and a persistently failing invoices query should not permanently brick the delete button.

web/sdk/client/views/general/general-view.tsx
whoAbhishekSah: Fixed the loading gap (button disabled until the answer arrives) and switched to invoices.length > 0, trusting the server filter. Error keeps the button enabled deliberately — the server refuses a blocked delete either way, see the sibling thread.

whoAbhishekSah:
Tested this end to end on a local sandbox: client-demo (Vite, port 3000) built from this branch, talking to a Frontier server on main (which already serves `CheckOrganizationDelete`). Logged in through the real UI with the mailotp flow. Blocker states were set up as real rows in the local database; the happy-path cases ran real deletes all the way through.

All cases were retested at the branch head (`274f3f34`), after the blocker instructions moved to the shared `delete-blockers.ts`, the refusal toast switched to mapped instructions, and the dialog and debt wording changed.

| # | Case | Setup | What the UI did | Result |
|---|------|-------|-----------------|--------|
| 1 | Clean org | No billing account | Delete button enabled, dialog shows no token warning, delete went through with a success toast | ✅ |
| 2 | Tokens remaining | Billing account with a 500 token balance | Button enabled; dialog warned "This workspace still has unused tokens, and deleting it forfeits them. If any of them were purchased, our support team will reach out to you to settle them"; delete went through and removed the ledger rows | ✅ |
| 3 | Paid subscription | Active subscription on a paid plan | Button disabled; tooltip: "Please downgrade the subscription to the standard plan" | ✅ |
| 4 | One unpaid invoice | 1 open invoice | Button disabled; tooltip: "Please pay the open invoice from the billing page" (singular) | ✅ |
| 5 | Two unpaid invoices | 2 open invoices | Button disabled; tooltip: "Please pay the 2 open invoices from the billing page" (plural with count) | ✅ |
| 6 | Token debt | Postpaid account (`credit_min` below zero) with a -300 balance | Button disabled; tooltip: "Your token balance is negative. Please add tokens from the Tokens page to bring it back up, or pay the invoice when it arrives, and then you can delete" | ✅ |
| 7 | All three blockers | Paid sub + open invoice + debt on one org | Button disabled; tooltip stacked all three lines | ✅ |
| 8 | Check goes stale | Check passed, then debt added behind the page, then confirm clicked | Server refused; toast "Cannot delete this workspace yet" with the mapped instruction parsed from the error details — the raw server text (with the billing account id) is not shown | ✅ |
| 9 | Double click during delete | Clicked confirm once, then 3 more times while the request was in flight (network throttled) | Button showed "Deleting…" and blocked re-clicks; exactly one `DeleteOrganization` request was sent | ✅ |
| 10 | Member without delete permission | Second user joined as org viewer | Button disabled with the permission tooltip; `CheckOrganizationDelete` was never called | ✅ |
| 11 | Org with no billing account opens the dialog | Same as case 1 | Confirm button did not get stuck waiting for a token balance that will never load | ✅ |
| 12 | Adding tokens clears the debt blocker | Debt org from case 6, then a credit covering the debt | Balance went from -300 to +200; on reload the delete button was enabled — no waiting for the invoice | ✅ |

The debt message language matches what the user sees elsewhere: the Tokens page shows the negative number under "Available tokens" (verified showing -300) next to an "Add tokens" button, and adding tokens genuinely unblocks the delete right away (case 12).

Not covered, because a healthy server cannot produce them: the fail-open path when the check RPC itself errors, and the unknown-blocker-kind fallback to the generic message. Both are small branches that were reviewed by reading the code.

web/sdk/client/views/general/components/delete-organization-dialog.tsx
```diff
@@ -44,6 +45,9 @@ export const DeleteOrganizationDialog = ({
   const orgLabel = t.organization({ case: 'capital' });
   const orgLabelLower = t.organization({ case: 'lower' });
   const [isAcknowledged, setIsAcknowledged] = useState(false);
+  // fetched only while the dialog is open; the confirm button waits for the
+  // answer so the forfeit warning cannot be skipped by a slow response
+  const { tokenBalance, isTokensLoading } = useTokens({ enabled: open });
```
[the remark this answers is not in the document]
whoAbhishekSah: Fixed: useTokens now exposes isFetching and the confirm button keys off it, so a reopened dialog refetching over a cached balance holds the confirm until the fresh answer arrives. On the regression test: the SDK views have no component-test setup to extend here; the dialog behavior was exercised live instead (see the end-to-end results comment on this PR — including the no-balance-support account that must not wedge the confirm).

Shreyag02:
## Blocker — the check and the delete disagree about which invoices count

`CheckOrganizationDelete` reads local rows: `invoice.Filter{NonZeroOnly: true}` → SQL `amount > 0`, where `amount` is Stripe's Total (`stripeInvoiceToInvoice`: `Amount: stripeInvoice.Total`).

The real delete reads live Stripe: `ListPayableOnProvider` → `AmountRemaining > 0`, with a comment spelling out why Total is the wrong test — *an open invoice fully covered by their credit balance has a positive total with nothing due*.

There is no `amount_remaining` column locally, and `upsert` only ever updates `State`, `EffectiveAt` and `HostedURL` — never `Amount` — so the stored total is frozen at first sync. That breaks both ways:

- **Blocks a delete that would succeed.** `total > 0, amount_remaining == 0` → the check reports `UNPAID_INVOICE` → button greyed out, tooltip telling the user to pay an invoice with nothing due. Clears only when the provider changes state.
- **Allows a delete that will fail.** A draft first synced before its line items land keeps `amount = 0`, gets excluded by `NonZeroOnly`, so `canDelete: true` and the server refuses on click.

Both were invisible before this PR, because the delete was the only thing anyone asked. Now the check drives the button, so the first case is a dead end for the user. Worth a decision: have the check read from the provider like the delete does, store `amount_remaining`, or keep a client-side override.

## The description does not match the code

The body says the tooltip has *a fallback to the server's own message for kinds this version does not know*. It does not — unknown kinds get the hardcoded `GENERIC_DELETE_BLOCKED_MESSAGE`, and `blocker.message` is not read anywhere in the SDK. Looks like the description predates 43bb2b17. I would argue the description had it right and the code should follow it.

whoAbhishekSah:
@Shreyag02 both points taken, here is what happened with each:

**The check/delete disagreement.** The root cause you found — the sync freezing the amount at first sight — is fixed in #1906: the sync now writes the total when it changes and the repository update includes the amount column. That kills the dangerous direction (a draft that grew after first sync no longer hides from the check, so the button stops enabling deletes the server refuses). The other direction — an open invoice fully covered by credit balance greying the button until the provider marks it paid — remains, deliberately: it is short-lived, self-clearing, and errs toward under-promising, while fixing it needs an `amount_remaining` column. That is filed as #1907 rather than done here, so the check keeps reading local rows and settings-page views stay off the provider's API.

**The description mismatch.** The description was stale and is fixed — the code's behavior is the intended one. Unknown blocker kinds get the generic line on purpose: the server's own messages carry entity ids ("subscription[uuid] is …"), and keeping raw server text out of the UI was an explicit decision on this PR. An unknown kind is one this SDK build cannot render properly; the generic line routes the user to support, which is the honest answer until the SDK learns the new kind.

rohanchkrabrty (approved):

Shreyag02 (approved):

PR #1882, opened 2026-08-13
feat(reconcile): add a MetaSchema kind for declarative metaschema management

rohilsurana:
## Summary

This adds a `MetaSchema` kind to the `frontier reconcile` flow. It manages the built-in entity metaschemas, the JSON schemas that validate the `metadata` field on users, organizations, groups, roles, and prospects, as declarative files. Before this, the only way to change a metaschema was a hand-run API call, with no file, no plan, and no record of what should exist.

Metaschemas are managed as values, mirroring predefined roles: the file sets the JSON schema for a built-in, and a built-in left out of the file resets to its shipped default. There is no delete.

## Changes

- Lift the default schemas into a shared `core/metaschema.Defaults`, one source for both the server seeding and the reconcile kind. Adding a metaschema for a new resource later is then a single edit.
- Add the pure diff, validation, and export helpers in `internal/reconcile/metaschema.go`.
- Add the `MetaSchemaReconciler`.
- Register the `MetaSchema` kind in `cmd/reconcile.go`, which enables both `frontier reconcile` and `frontier export metaschema`.
- Document it in the RFC and the reconcile guide, and list it in the CLI help.

## Technical Details

- The managed set is the five built-ins: user, group, organization, role, prospect. A built-in the file lists is set to its schema, one left out resets to its shipped default, and a name outside the set is rejected at validation. A built-in missing on the server is created.
- The reset target is compiled into the reconcile binary, like predefined roles, because the database stores only the current schema, not the shipped default. The shared `core/metaschema.Defaults` keeps it to one source.
- Schema comparison uses normalized JSON, so formatting never makes a false diff, and an export round-trips to zero changes (RFC rule 5).
- The MetaSchema RPCs live on `FrontierService`, and `UpdateMetaSchema` keys by UUID, so the reconciler resolves name to id from `ListMetaSchemas` first.
- The postgres store change is limited to `MigrateDefaults`, which now ranges over the shared map. Seeding behavior is unchanged.

## Merge ordering

This should merge after #1878 (the metaschema cache refresh). Without that refresh, a schema change made through the API reaches only the pod that handled it, so on a multi-pod server the reconcile plan and applies would not be consistent across pods. The RFC edit here removes the "blocked on per-pod cache" future-work bullet, and the guide notes that a change propagates within the cache refresh interval. Both assume #1878 is deployed.

## Test Plan

- [x] Build and type checking passes (`go build ./...`, `go vet`)
- [x] New unit tests pass under `go test -race` (`core/metaschema`, `internal/reconcile`), covering the diff (set, reset, create), validation (unknown name, duplicate, invalid JSON), JSON normalization (whitespace and key order), and the export round-trip

## SQL Safety (touches `metaschema_repository.go`)

- [x] No query is constructed or changed. The only edit to `metaschema_repository.go` is `MigrateDefaults` ranging over the shared `metaschema.Defaults` map instead of a local one. The placeholder, `ToSQL()`, and `goqu.L` items do not apply.


docs/content/docs/reconcile.mdx
```diff
@@ -358,6 +358,31 @@ spec:
...
+        "type": "object",
+        "properties": { "cost_center": { "type": "string" } },
+        "required": ["cost_center"]
+      }
+```
+
+The schema is a JSON string. Export writes only the built-ins whose schema differs
+from the default, so a freshly exported file lists just what an operator changed.
+
+A metaschema change reaches every pod within the server's cache refresh interval, about a minute by default, so it can take that long to take effect everywhere.
```
rohilsurana, line 384: This propagation claim does not match the server. There is no metaschema cache refresh interval anywhere in the codebase. `core/metaschema/service.go` holds a plain `map[string]MetaSchema` populated once at boot (`cmd/serve.go:167` calls `List`) and only mutated by that same pod on its own `Create`/`Update`/`Delete`. A reconcile change updates the DB and the single pod that served the write; every other pod keeps serving its stale in-memory copy until it restarts. So a change does not "reach every pod within about a minute" - on a multi-pod deployment it never propagates without a restart. RFC 0001 gated this kind on "once the server stops caching them per pod at boot"; that line was removed here without the precondition being met.

internal/reconcile/metaschema_reconciler.go
```diff
@@ -0,0 +1,112 @@
...
+// desired-state spec, so reconciling it plans no changes.
+func (r *MetaSchemaReconciler) Export(ctx context.Context) (any, error) {
+	current, err := r.fetchCurrent(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return exportMetaSchemas(current, r.defaults)
+}
+
+func (r *MetaSchemaReconciler) fetchCurrent(ctx context.Context) ([]currentMetaSchema, error) {
```
rohilsurana, line 88: `fetchCurrent` reads `ListMetaSchemas`, which the server serves from a per-pod in-memory cache that is filled once at boot and only updated by that pod's own writes (`core/metaschema/service.go`). On a multi-pod deployment the pod that answers this list call may hold a stale view, so the plan is computed against stale data. Two concrete failures: (1) it can plan a `create` for a built-in that already exists in the DB (this pod just has not cached it); the create then hits `OnConflict DoNothing` -> `sql.ErrNoRows` -> `CodeInternal`, failing the run with an opaque error. (2) it can miss or re-apply a change another pod already made. The diff/plan is only as correct as the single pod that answered the list.

internal/reconcile/metaschema.go
```diff
@@ -0,0 +1,171 @@
...
+			return fmt.Errorf("unknown metaschema %q", name)
+		}
+		if _, dup := seen[name]; dup {
+			return fmt.Errorf("metaschema %q is listed more than once", name)
+		}
+		seen[name] = struct{}{}
+		if strings.TrimSpace(s.Schema) == "" {
+			return fmt.Errorf("metaschema %q: schema is required", name)
+		}
+		if _, err := canonicalJSON(s.Schema); err != nil {
```
rohilsurana: Validate only checks the schema parses as JSON, not that it is a JSON object / usable JSON schema. Values like `123`, `"x"`, `true`, or `[]` are valid JSON and pass here, so they sail through `Validate` and get written to the server (which does not compile the schema on create/update either). Afterwards `metaschema.Service.Validate` builds `gojsonschema.NewStringLoader(schema)` for every entity of that type, and a non-object root schema makes validation error out, so all metadata writes for that entity type start failing. Since this is not caught locally, it can also partially apply in a multi-entry file: earlier ops commit, then a later bad-but-valid-JSON schema breaks that entity type. Consider compiling the schema (e.g. via gojsonschema) in Validate.

internal/reconcile/metaschema.go
```diff
@@ -0,0 +1,171 @@
...
+		return fmt.Sprintf("set metaschema %s", o.name)
+	}
+}
+
+// canonicalJSON returns a stable form of a JSON document, so two schemas that
+// differ only in whitespace or key order compare equal. It keeps the export
+// round-trip stable and stops formatting from making a false diff.
+func canonicalJSON(s string) (string, error) {
+	var v any
+	if err := json.Unmarshal([]byte(s), &v); err != nil {
```
rohilsurana: `json.Unmarshal` into `any` decodes every JSON number as `float64`, so `canonicalJSON` loses precision for large integers and high-precision numbers. Two schemas that differ only in such a number (e.g. `{"maximum": 10000000000000001}` vs `...002`) canonicalize to the same string. Because `canonicalJSON` is the equality test used by `diffMetaSchemas` (skip when equal) and `exportMetaSchemas` (omit when equal to default), the effect is silent: a real change is planned as a no-op, or a real override is dropped from export. Edge case for JSON schemas, but a genuine correctness hole. Using `json.Decoder` with `UseNumber()` would preserve the numbers.

internal/reconcile/metaschema.go
```diff
@@ -0,0 +1,171 @@
...
+// name is a known built-in, no name repeats, and the schema is non-empty valid
+// JSON. defaults is the managed set and the source of known names.
+func validateMetaSchemaSpecs(specs []MetaSchemaSpec, defaults map[string]string) error {
+	seen := map[string]struct{}{}
+	for _, s := range specs {
+		name := strings.TrimSpace(s.Name)
+		if name == "" {
+			return fmt.Errorf("metaschema name is required")
+		}
+		if _, ok := defaults[name]; !ok {
```
rohilsurana, line 111: Name matching here is exact (only `TrimSpace`), while sibling reconcilers key names case-insensitively (e.g. billingplan/billingproduct use `strings.ToLower`). So a hand-written `name: Organization` or `name: User` is rejected as `unknown metaschema "Organization"`, even though it is a built-in. The failure is loud, not silently wrong, but the error text is confusing and the behavior is inconsistent with the other kinds. `diffMetaSchemas` uses the same exact-match, so consider lowercasing the name in both places for consistency.

docs/content/docs/reconcile.mdx
rohilsurana: Intentional, and gated on #1878. This PR is set to merge after #1878, which adds the metaschema cache refresh (`app.metaschema.refresh_interval`, default 1m) and turns the per-pod map into a periodically reloaded one. So on `main` after both land, a change does reach every pod within the refresh interval, which is what this note describes. On this branch alone the refresh does not exist yet, which is why the note reads ahead of the code. The RFC bullet was removed for the same reason: #1878 is the "server stops caching them per pod" half, this PR is the kind half, and the last of the two to merge clears the item. If you would rather not couple the two, I can restore the bullet here and drop it when #1878 lands.

internal/reconcile/metaschema_reconciler.go
rohilsurana: Same #1878 dependency. The reconcile read is only as fresh as the answering pod's cache, so this kind needs #1878's periodic refresh to be correct across pods, and this PR merges after it. With the refresh the list is eventually consistent within the interval, which fits the RFC's converge-not-transact model: a racing change is corrected on the next run.

On failure (1) specifically: the five built-ins are seeded at `server migrate`, before any pod boots, so every pod has them cached and the plan is an update, not a create. The create path only fires for a genuinely new built-in the server does not have yet, where the create succeeds. The opaque `CodeInternal`-on-conflict you describe would need a brand-new built-in introduced while some pod still holds a pre-seed cache, which is a sub-refresh-interval window. I kept create-if-missing so a new built-in can be bootstrapped, but I can drop it if you would rather the kind never plan a create.

internal/reconcile/metaschema.go
rohilsurana: Fixed in 6b7a621b. `validateMetaSchemaSpecs` now rejects a schema whose root is not a JSON object, so `123`, `"x"`, `true`, and `[]` fail the plan up front instead of being written and breaking runtime validation for that entity type. Added a test covering those. I went with an object-root check rather than a full `gojsonschema` compile so reconcile stays offline and does not eagerly resolve a remote `$ref` during validation. If you also want to catch a malformed-but-object schema (like `{"type": 123}`), I can add the compile too.

internal/reconcile/metaschema.go
rohilsurana: Fixed in 6b7a621b. `canonicalJSON` now decodes with `json.Decoder` and `UseNumber()`, so numbers keep their literal form and are compared exactly. Two schemas differing only in a large integer no longer canonicalize to the same string, so a real change is not silently planned as a no-op or dropped from export. Added a test with `10000000000000001` vs `...002`. The strict single-document parse is kept, so trailing data is still rejected.

internal/reconcile/metaschema.go
rohilsurana: Fixed in 6b7a621b. Both `validateMetaSchemaSpecs` and `diffMetaSchemas` now lowercase the name before matching, so `name: Organization` or `name: User` resolves to the built-in, consistent with the billing kinds. The op writes the canonical lowercase name. Added tests for both validate and diff.

internal/reconcile/metaschema.go
rohilsurana: The `UseNumber` fix is correct, but the test that guards it does not actually catch a regression.

`TestCanonicalJSON/keeps a real difference in a large integer` compares `10000000000000001` and `10000000000000002`. Under the old `float64` decode those canonicalize to `10000000000000000` and `10000000000000002`, which already differ, so the test passes even without this fix. I checked by reverting `decodeJSON` back to `json.Unmarshal` and the test still passed.

The pair that actually collides under `float64` is at 2^53: `9007199254740993` and `9007199254740992` both round to `9007199254740992`, so they compare equal on the old code. That is the exact case this fix is for. Swapping the test to that pair makes it fail before the fix and pass after.

```diff
-		a, err := canonicalJSON(`{"maximum":10000000000000001}`)
-		assert.NoError(t, err)
-		b, err := canonicalJSON(`{"maximum":10000000000000002}`)
+		a, err := canonicalJSON(`{"maximum":9007199254740993}`)
+		assert.NoError(t, err)
+		b, err := canonicalJSON(`{"maximum":9007199254740992}`)
```

internal/reconcile/metaschema.go
rohilsurana: Reverted the object-root check in 04bb0d73. On reflection it made the reconciler stricter than the server's write API, which accepts any non-empty schema, and that broke the rule 5 round-trip: a non-object schema written through the raw API is a reachable state, and its export then failed its own re-reconcile. Confirmed live: `organization` set to `123` via the API now exports and reconciles back to no changes, where before it errored with "schema must be a JSON object".

The valid-JSON check stays, so a bare typo is still caught. If a non-object schema should be rejected, that belongs on the write API, where it also stops the state from being reached in the first place. I left a comment on `validateMetaSchemaSpecs` so the object check is not re-added by a future reviewer.

rohilsurana: Reverted the removal in 8e0dd530 — keeping the object-check after all. Catching a non-object schema at plan time is worth it: without the check, reconcile would write a schema like `123` that then breaks runtime metadata validation for that entity type. The rule 5 tradeoff stands and is documented in the audit: a non-object schema written directly through the raw API is a reachable state whose export will not round-trip. The right place to close that fully is the write API (validate object-ness on `CreateMetaSchema`/`UpdateMetaSchema`), which also stops the state from being reached at all. Re-resolving.

Issue #1782, opened 2026-07-21
Remove the deprecated API usage acknowledged in the staticcheck rollout


AmanGIT07:
PR #1781 annotated every remaining use of a deprecated API with `//nolint:staticcheck`. Each annotation is deliberate, but all of them are debt. This issue tracks removing them. The sites, grouped by what unblocks each:

### Actionable now

- **New Relic go-agent v2 → v3** — `pkg/db/db.go` and `internal/store/spicedb/relation_repository.go` import the deprecated v2 package. Migrating means the v3 import path and the reworked transaction/segment APIs at every telemetry call site.
- **e2e billing tests** — `test/e2e/regression/billing_test.go` uses the deprecated `UpdateBillingAccountLimits` RPC as a setup tool. `UpdateBillingAccountDetails` sets the same `credit_min` (note it also writes `due_in_days`).

### Blocked on the API dropping the deprecated request fields

Handler fallbacks that must stay while old clients send the old shapes:

- `internal/api/v1beta1connect/billing_usage.go` — `since` (superseded by `start_range`)
- `internal/api/v1beta1connect/billing_subscription.go` — flat `plan`/`immediate` (superseded by `plan_change`)
- `internal/api/v1beta1connect/permission.go` — `namespace`/`name` body fields (superseded by `key`)
- `internal/api/v1beta1connect/permission_check.go` — split object fields (superseded by `resource`)
- `cmd/serve.go` — the `FullyConsistent` config flag (superseded by `consistency`)
- `cmd/seed.go` — seed data sends the deprecated permission body shape

### Blocked on a lossless replacement for the deprecated response fields

`internal/reconcile/permission_reconciler.go`, `cmd/permission.go`, and the assertions in `test/e2e/regression/service_registration_test.go` read the deprecated `namespace`/`name` response fields. The replacement `key` round-trips lossily today: a namespace without a slash gains a `/default` suffix through `SplitNamespaceResource`, and names containing dots fail to split. These readers can migrate once `key` (or another field) carries the exact values.

PR #1885, opened 2026-08-14
refactor(api): require permission key in create and update requests

AmanGIT07:
## Summary
CreatePermission and UpdatePermission accept the permission identity only via the `key` field (`service.resource.verb`). The deprecated `namespace`/`name` request body fields are no longer read; requests sending only those fields now receive InvalidArgument. Part of #1782.

## Changes
- `internal/api/v1beta1connect/permission.go`: remove the namespace/name fallback; reject a missing or malformed key in both CreatePermission and UpdatePermission
- `cmd/seed/permissions.json` + `cmd/seed.go`: seed data sends `key`
- `test/e2e/regression/service_registration_test.go`: request bodies send `key`
- `internal/api/v1beta1connect/permission_test.go`: request bodies send `key`; add missing-key and malformed-key cases for both handlers

## Test Plan
- [x] `go test ./internal/api/v1beta1connect/` passes
- [x] `make lint` passes (0 issues)
- [x] Full e2e regression suite ran successfully

### DIFF
--- a/cmd/seed/permissions.json
+++ b/cmd/seed/permissions.json
@@ -1,50 +1,42 @@
 [
   {
-    "name": "create",
+    "key": "compute.order.create",
     "title": "Create Order",
-    "namespace": "compute/order",
     "metadata": {}
   },
   {
-    "name": "get",
+    "key": "compute.order.get",
     "title": "Read Order",
-    "namespace": "compute/order",
     "metadata": {}
   },
   {
-    "name": "update",
+    "key": "compute.order.update",
     "title": "Update Order",
-    "namespace": "compute/order",
     "metadata": {}
   },
   {
-    "name": "delete",
+    "key": "compute.order.delete",
     "title": "Delete Order",
-    "namespace": "compute/order",
     "metadata": {}
   },
   {
-    "name": "list",
+    "key": "compute.order.list",
     "title": "List Orders",
-    "namespace": "compute/order",
     "metadata": {}
   },
   {
-    "name": "get",
+    "key": "database.order.get",
     "title": "List Database Order",
-    "namespace": "database/order",
-    "metadata":{}
+    "metadata": {}
   },
   {
-    "name": "update",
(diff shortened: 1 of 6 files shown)


cmd/seed/permissions.json
```diff
@@ -1,50 +1,42 @@
 [
   {
-    "name": "create",
```
whoAbhishekSah, line 3: What is the purpose of these seed JSON files? It looks like dummy data. If they are not needed, we can think of removing them.

whoAbhishekSah:
@AmanGIT07, did you test that the reconcile flow is working correctly for permissions?

AmanGIT07:
@whoAbhishekSah Verified live against a server built from this branch: reconcile dry-run plans the adds, apply creates them through the key-only CreatePermission, an immediate re-run converges to "no changes", export reads back the exact namespace/name pair, and `delete: true` removes them. An invalid namespace in the desired-state file fails the plan client-side before any server call.

Also rebased the branch onto current main (post #1887, so the key-reading reconciler was in the loop for the run above) and dropped the namespace-grammar commit from this PR — `fix/reconcile-permission-spicedb-grammar` covers that more completely.

cmd/seed/permissions.json
AmanGIT07: The six JSON files under cmd/seed/ (permissions, roles, users, organizations, projects, resource) are compiled into the binary with go:embed and used by exactly one thing: the `frontier seed` CLI command, documented in the CLI reference. Running it against a live server creates a sample world through the public API — the compute/order and database/order permissions, two custom roles, a sample org with users, projects, a resource, and policies. Its purpose is stated in its own help text: making it easier to get started — demo data for someone evaluating Frontier locally.

Removing the command and its fixtures is doable, but that's its own discussion — keeping this PR to the request-field change.

whoAbhishekSah (approved):

PR #1886, opened 2026-08-14
refactor(api): require resource in CheckResourcePermission

AmanGIT07:
## Summary
CheckResourcePermission accepts the object only via the `resource` field (`namespace:id`). The deprecated `object_id`/`object_namespace` request fields are no longer read; requests sending only those fields now receive InvalidArgument. Namespace aliases keep working inside `resource`. Part of #1782.

## Changes
- `internal/api/v1beta1connect/permission_check.go`: remove the split-field fallback; reject a missing or malformed `resource`
- `test/e2e/regression/onboarding_test.go`, `serviceusers_test.go`, `api_test.go`: check requests send `resource`
- `internal/api/v1beta1connect/permission_check_test.go`: success cases send `resource`; add missing-resource case

## Test Plan
- [x] `go test ./internal/api/v1beta1connect/` passes
- [x] `make lint` passes (0 issues)
- [x] Onboarding, service-users, and API e2e regression suites pass

### DIFF
--- a/internal/api/v1beta1connect/permission_check.go
+++ b/internal/api/v1beta1connect/permission_check.go
@@ -72,12 +72,12 @@ func (h *ConnectHandler) CheckFederatedResourcePermission(ctx context.Context, r
 
 	objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetResource())
 	if err != nil || objectNamespace == "" || objectID == "" {
-		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		return nil, connect.NewError(connect.CodeInvalidArgument, ErrNamespaceSplitNotation)
 	}
 
 	principalNamespace, principalID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetSubject())
 	if err != nil || principalNamespace == "" || principalID == "" {
-		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		return nil, connect.NewError(connect.CodeInvalidArgument, ErrNamespaceSplitNotation)
 	}
 
 	permissionName, err := h.getPermissionName(ctx, objectNamespace, req.Msg.GetPermission())
@@ -163,7 +163,7 @@ func (h *ConnectHandler) CheckResourcePermission(ctx context.Context, req *conne
 
 	objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetResource())
 	if err != nil || objectNamespace == "" || objectID == "" {
-		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		return nil, connect.NewError(connect.CodeInvalidArgument, ErrNamespaceSplitNotation)
 	}
 
 	permissionName, err := h.getPermissionName(ctx, objectNamespace, req.Msg.GetPermission())
@@ -196,8 +196,8 @@ func (h *ConnectHandler) BatchCheckPermission(ctx context.Context, req *connect.
 	checks := make([]resource.Check, 0, len(req.Msg.GetBodies()))
 	for _, body := range req.Msg.GetBodies() {
 		objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(body.GetResource())
-		if len(body.GetResource()) == 0 || err != nil {
-			return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		if err != nil || objectNamespace == "" || objectID == "" {
+			return nil, connect.NewError(connect.CodeInvalidArgument, ErrNamespaceSplitNotation)
 		}
 
 		permissionName, err := h.getPermissionName(ctx, objectNamespace, body.GetPermission())
--- a/internal/api/v1beta1connect/permission_check.go
+++ b/internal/api/v1beta1connect/permission_check.go
@@ -162,12 +162,7 @@ func (h *ConnectHandler) CheckResourcePermission(ctx context.Context, req *conne
 	errorLogger := NewErrorLogger()
 
 	objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetResource())
-	//nolint:staticcheck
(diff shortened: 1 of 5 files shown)


internal/api/v1beta1connect/permission_check.go
```diff
@@ -201,8 +196,8 @@ func (h *ConnectHandler) BatchCheckPermission(ctx context.Context, req *connect.
 	checks := make([]resource.Check, 0, len(req.Msg.GetBodies()))
 	for _, body := range req.Msg.GetBodies() {
 		objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(body.GetResource())
-		if len(body.GetResource()) == 0 || err != nil {
-			return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+		if err != nil || objectNamespace == "" || objectID == "" {
```
rohilsurana, line 199: One malformed or empty-part resource in any body now makes the whole batch return InvalidArgument with zero results. For a batch endpoint that is surprising: 49 valid checks are dropped because of 1 bad body. Consider failing just that item, or returning a per-item error, rather than the whole call.

internal/api/v1beta1connect/permission_check.go
```diff
@@ -162,13 +162,8 @@ func (h *ConnectHandler) CheckResourcePermission(ctx context.Context, req *conne
...
 
 	objectNamespace, objectID, err := schema.SplitNamespaceAndResourceID(req.Msg.GetResource())
-	//nolint:staticcheck
-	if len(req.Msg.GetResource()) == 0 || err != nil {
-		objectNamespace = schema.ParseNamespaceAliasIfRequired(req.Msg.GetObjectNamespace())
-		objectID = req.Msg.GetObjectId()
-	}
-	if objectNamespace == "" || objectID == "" {
-		return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+	if err != nil || objectNamespace == "" || objectID == "" {
```
rohilsurana, line 165: This drops the fallback to object_id / object_namespace, so a client sending only those (with no resource) now gets InvalidArgument. That is the intended tightening, but those fields are still in the .proto, so the schema still advertises support the server no longer provides. A clear deprecation note in the proto, or a planned removal, would keep integrations from being surprised at runtime.

internal/api/v1beta1connect/permission_check_test.go
```diff
@@ -91,9 +127,8 @@ func TestHandler_CheckResourcePermission(t *testing.T) {
 					Return(testPermission, nil)
 			},
 			request: connect.NewRequest(&frontierv1beta1.CheckResourcePermissionRequest{
-				ObjectId:        testRelationV2.Object.ID,
-				ObjectNamespace: testRelationV2.Object.Namespace,
-				Permission:      schema.UpdatePermission,
+				Permission: schema.UpdatePermission,
+				Resource:   schema.JoinNamespaceAndResourceID(testRelationV2.Object.Namespace, testRelationV2.Object.ID),
```
rohilsurana, line 131: The success cases use canonical namespaces via JoinNamespaceAndResourceID, and none pass a real alias in the resource field. The PR says aliases still work there, so a case with resource set to an alias (e.g. "org:<id>") would lock that promise in and catch any regression in ParseNamespaceAliasIfRequired for this path.

internal/api/v1beta1connect/permission_check_test.go
```diff
@@ -138,3 +172,47 @@ func TestHandler_CheckResourcePermission(t *testing.T) {
...
+				Bodies: []*frontierv1beta1.BatchCheckPermissionBody{
+					{Resource: ":" + testRelationV2.Object.ID, Permission: schema.UpdatePermission},
+				},
+			}),
+			wantErr: connect.NewError(connect.CodeInvalidArgument, ErrNamespaceSplitNotation),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockDep := &ConnectHandler{}
```
rohilsurana: This table builds the handler as &ConnectHandler{} with no services, which is safe only because every case fails validation before any service is called. If someone later adds a valid-resource case, it will reach getPermissionName and nil-panic instead of failing cleanly. The other handler's test injects mocks; doing the same here would avoid that trap.

rohilsurana (commented):
Review against main. The change itself is a clean tightening. A couple of behavior and test notes below.

internal/api/v1beta1connect/permission_check.go
AmanGIT07: Proto will be cleaned up later

internal/api/v1beta1connect/permission_check_test.go
AmanGIT07: Added in f00e7d3c — a unit case now sends `org:<id>` and asserts the check runs against `app/organization`.

internal/api/v1beta1connect/permission_check_test.go
AmanGIT07: Done in f00e7d3c — the batch table now injects the service mocks through a setup hook, same shape as the single-check table.

internal/api/v1beta1connect/permission_check.go
AmanGIT07: Keeping whole-batch rejection in this PR: that's the pre-existing contract for a malformed body here — before this change a malformed resource already failed the whole call, and an empty-part one reached SpiceDB and failed the whole call as internal. This PR only turns that into a clean InvalidArgument. Failing just the bad item needs a response-shape change (BatchCheckPermissionResponsePair has no per-item error field), so it's a proto addition rather than a handler tweak. Happy to take that as a follow-up if we want the per-item contract.

internal/api/v1beta1connect/permission_check.go
AmanGIT07: Adding to that: the fields already carry `deprecated = true` in the proto, so the schema does flag it — the generated getters are marked deprecated too. The removal itself rides a later proton sync.

rohilsurana (approved):

whoAbhishekSah (approved):

PR #1887, opened 2026-08-17
fix(permission): make the response key lossless and migrate its readers

whoAbhishekSah:
Part of #1782 (the "blocked on a lossless replacement for the deprecated response fields" group).

## Problem

The `key` response field (`service.resource.verb`) is the replacement for the deprecated `namespace`/`name` fields, but reading a key back was lossy in two ways:

- A permission name containing dots failed to split: the parser demanded exactly 3 dot-parts and returned empty strings for a key like `database.instance.soft.delete`.
- A namespace without a slash silently gained a `/default` suffix while the key was built, so the key read back as a different namespace than the stored one.

Because of that, three readers stayed on the deprecated fields (with `//nolint:staticcheck`): the permission reconciler, the permission CLI output, and the e2e assertions. This PR makes the key read back to the exact stored values and moves those readers onto it.

One behavior note: a create/update request whose `key` has more than 3 dot-parts used to be treated as "no key" and fall back to the deprecated request fields. It now parses to a name with dots and is rejected as an invalid name. (The request-side fallbacks themselves are being removed in #1885.)

## Testing

- Unit tests for the parser (dot names, short keys, empty segments) and a key round-trip test.
- A transform test that a dot name survives and a slash-less namespace errors.
- Reconciler tests updated to serve `key` only, plus a test that an unsplittable key fails the run.
- `go test` on the touched packages and the full unit suite pass; the service registration e2e suite passes locally against Docker.

### DIFF
--- a/cmd/permission.go
+++ b/cmd/permission.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/MakeNowJust/heredoc"
+	"github.com/raystack/frontier/internal/bootstrap/schema"
 	"github.com/raystack/frontier/pkg/file"
 	frontierv1beta1 "github.com/raystack/frontier/proto/v1beta1"
 	"github.com/raystack/salt/cli/printer"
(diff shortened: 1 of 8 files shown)


whoAbhishekSah:
Tested live against a local server built from this branch (verified the running binary contains the new code paths).

- `frontier export permission` — read all 15 custom permissions through the new key-splitting path; output matched their exact stored namespace/name.
- Added `sandbox/gadget:probe` to the exported file. Dry run planned exactly 1 add and no spurious changes, meaning every existing permission matched its spec via `key`.
- Applied. Server returned `key: sandbox.gadget.probe` with namespace `sandbox/gadget`, name `probe` — exact match.
- Second dry run: `Permission: no changes` — full lossless round trip (spec → create via key → read back via key → diff clean).
- `frontier permission view <id>` prints the right NAME/NAMESPACE columns, now derived from the key.
- Cleanup: marked `delete: true`, dry run planned 1 delete, applied, verified gone from ListPermissions.

AmanGIT07 (approved):

PR #1889, opened 2026-08-17
fix(reconcile): tighten permission name and namespace validation to SpiceDB grammar

rohilsurana:
## What

Plan-time validation for a permission's verb and `service/resource` namespace was looser than the identifier grammar SpiceDB enforces at apply. An ordinary input planned cleanly and then failed at apply with a 500:

- a verb shorter than three characters (e.g. `id`)
- an uppercase verb (e.g. `Read`)
- a service or resource part starting with a digit, shorter than three characters, or uppercase

A permission's verb becomes a SpiceDB relation name directly, and its `service/resource` becomes a SpiceDB object type name, so both must satisfy SpiceDB's grammar: start with a lowercase letter, lowercase alphanumerics, three to sixty-four characters.

This tightens `IsValidPermissionName` and `permissionNamespaceRe` to that grammar. The no-underscore rule for the namespace parts and the verb is kept, since the slug joins service, resource, and verb with `_`.

## Why

RFC 0001 Rules 2 and 3: the file states a desired state that can actually apply, and validation happens before apply. A plan that looks clean must not fail at the server.

## Testing

- New `TestIsValidPermissionName` and expanded `TestIsValidPermissionNamespace` cover the tightened cases.
- Updated the `CreatePermission` handler test fixtures, which used uppercase names SpiceDB would reject.
- `internal/bootstrap/schema`, `internal/reconcile`, and `internal/api/v1beta1connect` pass.

## Stack

Stacked on `fix/reconcile-framework-known-fields`.

### DIFF
--- a/internal/api/v1beta1connect/permission_test.go
+++ b/internal/api/v1beta1connect/permission_test.go
@@ -24,7 +24,7 @@ var (
 	testPermissions   = []permission.Permission{
 		{
 			ID:          uuid.New().String(),
-			Name:        "Read",
+			Name:        "read",
 			NamespaceID: "app/resource",
 			Metadata:    map[string]any{},
 			CreatedAt:   time.Time{},
@@ -32,14 +32,14 @@ var (
 		},
 		{
 			ID:          uuid.New().String(),
-			Name:        "Write",
+			Name:        "write",
 			NamespaceID: "app/resource",
 			CreatedAt:   time.Time{},
 			UpdatedAt:   time.Time{},
 		},
 		{
 			ID:          uuid.New().String(),
-			Name:        "Manage",
+			Name:        "manage",
 			NamespaceID: "app/resource",
 			CreatedAt:   time.Time{},
 			UpdatedAt:   time.Time{},
--- a/internal/api/v1beta1connect/permission_test.go
+++ b/internal/api/v1beta1connect/permission_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


rohilsurana:
**Review: tighten permission grammar to SpiceDB (RFC 0001 Rule 3)**

Verified on this branch: `go build ./...`, `go test ./internal/reconcile/... ./internal/bootstrap/schema/... ./internal/api/v1beta1connect/...`, and `golangci-lint run` on the changed packages all pass.

This meets Rule 3 for the modes it claims. The new patterns `permissionNameRe = ^[a-z][a-z0-9]{2,63}$` and `permissionNamespaceRe = ^[a-z][a-z0-9]{2,63}/[a-z][a-z0-9]{2,63}$` (schema.go) enforce lowercase, a leading letter, 3 to 64 chars per part, and no underscore, and validation runs over every entry before any server write (in both `Validate` and `diffPermissions`). No new Rule 5 break: every shape it newly rejects (leading digit, uppercase, under three chars) is also rejected by SpiceDB itself, so the server cannot hold it and export cannot emit it.

Verdict: approve with nits.

I am putting the two permission-validation items here, since this is the grammar PR. #1892 now stacks on top of this branch and inherits this grammar, so fixing them here fixes them for both.

Nits:
1. The combined slug length is still unbounded. Each part caps at 64, but the slug `service_resource_verb` (from `FQPermissionNameFromNamespace`) can reach 194 chars, while SpiceDB caps a relation name at 64. Concrete case: namespace `aaaa...(30 chars)/bbbb...(30 chars)` plus name `get` passes `validatePermissionSpec` and `reconcile --dry-run` plans cleanly, but a real apply fails at schema compile because the 65-char relation is rejected. This is the exact "plan clean, apply 500" case the Why section cites Rule 3 to prevent. Add a bound like `len(service)+len(resource)+len(name)+2 <= 64`.
2. Reachable Rule 5 gap from an API/reconcile validation asymmetry (pre-existing, not introduced here, but this is the right PR to close it). `CreatePermission` (internal/api/v1beta1connect/permission.go:32) validates the verb via `IsValidPermissionName` but never calls `IsValidPermissionNamespace` on the namespace. SpiceDB allows underscores inside object-type names, so a permission `compute/sub_order:read` is creatable through the API and stored. `Export` then emits it and reconcile rejects it, because the namespace regex forbids the underscore. Adding `IsValidPermissionNamespace` to the API create path closes this for both the current format and #1892's key format.

rohilsurana:
Follow-up on the two items.

- Slug length: done. Added `schema.PermissionSlugWithinLimit`, called it from `validatePermissionSpec`, and covered it with a unit test and a reconcile test. A namespace and verb that each pass on their own but flatten to a slug over 64 chars now fails the plan instead of the apply.
- API namespace grammar: leaving it out of this PR on purpose. It is a different case from the length check. A too-long slug can never compile in SpiceDB, so rejecting it early breaks nothing. But underscores are legal in SpiceDB object-type names, so a permission like `compute/sub_order:read` is creatable and stored today through the API. Adding `IsValidPermissionNamespace` to the create path would reject that and break existing API-only callers. That is a public-API change with its own compat call, so it does not belong in the reconcile grammar PR. Reconcile already fails loud on such a namespace, which is the safe side.

rohilsurana:
**Re-review (head f60571fc)**

Build, tests, and lint are all green. The slug-length fix is correct and complete. `PermissionSlugWithinLimit` uses the same `service_resource_verb` flattening as apply and the bound is `<= 64`, so a 64-char slug passes and 65 fails, with no off-by-one. It runs server-free in both `Validate` and `diffPermissions`, over every entry, before any apply, so Rule 3 holds.

Still open (pre-existing, not claimed in this commit): the `CreatePermission` API (internal/api/v1beta1connect/permission.go:32) validates the verb via `IsValidPermissionName` but never calls `IsValidPermissionNamespace` and never checks the slug length. So an underscore namespace like `compute/sub_order`, or an over-long namespace/name, is still creatable through the API and then rejected on export and reconcile. Worth closing in the API create path in a follow-up.

Verdict: approve with nits.

internal/reconcile/permission.go
```diff
@@ -55,6 +55,12 @@ func validatePermissionSpec(s PermissionSpec) error {
 	if !schema.IsValidPermissionNamespace(s.Namespace) {
 		return fmt.Errorf("invalid namespace %q (each of service/resource must be lowercase alphanumeric)", s.Namespace)
 	}
+	// Each part is capped on its own, but the flattened slug service_resource_verb
+	// becomes a SpiceDB relation, which is capped at sixty-four characters. Reject a
+	// slug over that here so a plan cannot pass and then fail when the schema compiles.
+	if !schema.PermissionSlugWithinLimit(s.Namespace, s.Name) {
```
rohilsurana: Nit (pre-existing, not introduced here): this slug-length guard and `IsValidPermissionNamespace` run only on the reconcile path. The `CreatePermission` API (internal/api/v1beta1connect/permission.go:32) validates the verb via `IsValidPermissionName` but calls neither, so an over-long slug or an underscore namespace like `compute/sub_order` is still creatable through the API and then rejected on export and reconcile. That is a reachable Rule 5 gap. Worth adding both checks to the API create path in a follow-up.

rohilsurana: Agreed as a follow-up, with one correction. The two cases differ. An over-long slug cannot actually be created through the API today: `AppendSchema` compiles it into SpiceDB, which rejects the over-64 relation, so the call already fails and nothing is stored. So there is no export gap for length. The underscore namespace is the real reachable case: SpiceDB allows underscores in object types, so `compute/sub_order` is stored, then export emits it and reconcile rejects it. Closing that means adding `IsValidPermissionNamespace` to the create path, which rejects input that works today, so it is a public-API change with its own compat call. Tracking it as a follow-up, not part of this PR.

whoAbhishekSah (approved):

internal/bootstrap/schema/schema.go
```diff
@@ -299,34 +299,50 @@ func IsSystemNamespace(namespace string) bool {
...
+// API paths flatten a permission to the slug service_resource_verb and register
+// that slug as a SpiceDB relation, so a longer slug is rejected when the schema is
+// compiled.
+const maxPermissionSlugLen = 64
+
+// PermissionSlugWithinLimit reports whether the flattened slug
+// service_resource_verb fits SpiceDB's relation-name limit. Each part can be up to
+// sixty-four characters on its own, so a valid namespace and verb can still combine
+// into a slug that SpiceDB will not store.
+func PermissionSlugWithinLimit(namespace, name string) bool {
```
AmanGIT07, line 342: This new check, and `IsValidPermissionNamespace`, only run in reconcile's `validatePermissionSpec`. The other way a permission gets created is the `CreatePermission` handler (`internal/api/v1beta1connect/permission.go:32-39`), and it still checks only the verb plus the `!= app` rule. So the plan-passes-then-500 case this PR fixes for reconcile is still open there:

- namespace `Compute/order` or `1abc/order` passes the handler, then fails at `AppendSchema` when SpiceDB compiles the schema, and falls through to `CodeInternal` on line 67 (a 500 for bad input).
- two 30-char parts with verb `get` make a 65-char slug, same 500.
- namespace `res_order/item` is valid in SpiceDB and is stored fine by the handler, but reconcile's validator rejects underscores, so reconcile can then neither keep nor delete that row and every later permission plan errors.

Could we call these checks in the handler too, or move all three into one shared validator in the schema package, so both paths agree? The doc comment here already says it covers "the reconcile and API paths".

internal/api/v1beta1connect/permission_test.go
```diff
@@ -24,22 +24,22 @@ var (
 	testPermissions   = []permission.Permission{
 		{
 			ID:          uuid.New().String(),
-			Name:        "Read",
+			Name:        "read",
```
AmanGIT07, line 28: These fixtures were renamed `Read`→`read` to keep the existing tests green under the tighter rule. Two gaps come with that:

- The `CreatePermission` handler still returns "permission name cannot contain special characters" (`permission.go:33`) for every new rejection. Names like `id` (too short), `Read` (uppercase), and `getAll` (uppercase) have no special characters, so the message is now wrong. Reconcile got the corrected wording; the handler didn't.
- No handler test asserts that `Read` or `id` now returns `CodeInvalidArgument`. A later change that loosened this check would still pass the suite. Worth one rejection case here.

internal/reconcile/permission.go
```diff
@@ -40,7 +40,7 @@ func validatePermissionSpec(s PermissionSpec) error {
 		return fmt.Errorf("namespace and name are required")
 	}
 	if !schema.IsValidPermissionName(s.Name) {
-		return fmt.Errorf("invalid name %q (alphanumeric only)", s.Name)
+		return fmt.Errorf("invalid name %q (must be lowercase, start with a letter, and be at least three characters)", s.Name)
```
AmanGIT07: `owner`, `project`, and `granted` pass all the new rules, but the generator adds fixed relations with those exact names to every custom-resource definition (`generator.go:193-201`) on top of the per-verb relation (`generator.go:128`). So a spec like `{namespace: compute/disk, name: owner}` validates cleanly, the plan passes, and apply then fails when SpiceDB rejects the duplicate relation — the same plan-passes-apply-fails shape the slug check was added to stop.

Pre-existing, not introduced here. But since we're already tightening the verb rule on this line, could we also reject those three reserved verbs? Cheap check, and it closes the gap in the same spot.

internal/bootstrap/schema/schema.go
```diff
@@ -299,34 +299,50 @@ func IsSystemNamespace(namespace string) bool {
...
+// relation name directly, so it must satisfy SpiceDB's relation grammar
+// (^[a-z][a-z0-9_]{1,62}[a-z0-9]$): start with a lowercase letter, then
+// lowercase alphanumerics, three to sixty-four characters. Underscore is left
+// out here because the slug joins service, resource, and verb with "_", so an
+// underscore in the verb is treated the same as in a namespace part.
+var permissionNameRe = regexp.MustCompile(`^[a-z][a-z0-9]{2,63}$`)
+
+// IsValidPermissionName reports whether name is a valid permission verb that
+// SpiceDB will accept as a relation name.
 func IsValidPermissionName(name string) bool {
```
AmanGIT07, line 312: Worth a follow-up (out of scope here): this validator guards the create and reconcile paths, but `UpdatePermission` (`internal/api/v1beta1connect/permission.go:121`) passes the request name and namespace straight to the service with no validation at all. The comment there says it should only change metadata, but the name and namespace still get written.

A rename to something SpiceDB won't accept (say `Read`) writes to Postgres, and then the next boot's `MigrateSchema` merges that row into the schema and fails to compile, so the server won't start until the row is fixed by hand. Same validation hole, just on the update path.

internal/bootstrap/schema/schema.go
rohilsurana: Done. Added `schema.ValidateCustomPermission(namespace, name)`, which runs the verb grammar, the reserved-verb check, the namespace grammar, and the slug-length check in one place. Both `validatePermissionSpec` in reconcile and the `CreatePermission` handler now call it, so the two paths agree and the doc comment is now accurate.

All three cases you listed now return `CodeInvalidArgument` from the handler instead of a 500 or a stored-but-unmanageable row: `Compute/order` and `1abc/order` fail the namespace grammar, the two 30-char parts fail the slug-length check, and `res_order/item` fails the underscore rule so it can no longer be stored and then break every later reconcile plan.

One call-out: this makes the handler reject an underscore namespace that SpiceDB itself accepts, so it is a small public-API tightening. I think it is the right trade. Such a namespace can be stored, but export and reconcile can never manage it, so it was never usable end to end.

internal/api/v1beta1connect/permission_test.go
rohilsurana: Fixed both. The handler no longer returns "permission name cannot contain special characters". It now calls `schema.ValidateCustomPermission`, so `id` (too short), `Read` (uppercase), and `getAll` (uppercase) get a message that names the actual rule they broke.

Added rejection cases to `TestHandler_CreatePermission` for an uppercase verb, a too-short verb, a reserved verb, an underscore namespace, and an over-long slug, each asserting `CodeInvalidArgument`. So a later change that loosened the check would fail the suite.

internal/reconcile/permission.go
rohilsurana: Done. `ValidateCustomPermission` now rejects `owner`, `project`, and `granted` through `IsReservedPermissionVerb` (case-insensitive), so both reconcile and the create handler reject them up front instead of passing the plan and failing when SpiceDB compiles the duplicate relation. Added tests for the three reserved verbs in the schema, reconcile, and handler suites.

internal/bootstrap/schema/schema.go
rohilsurana: Agreed, keeping this out of scope for this PR as you suggested. `UpdatePermission` is meant to change only metadata, but it still writes the request name and namespace with no validation, so a rename to something SpiceDB will not accept can block the next boot's schema compile.

Now that the shared `ValidateCustomPermission` exists the follow-up is small: either validate in the update handler, or stop writing name and namespace there at all so it only ever touches metadata. I will track it separately.

PR #1890, opened 2026-08-17
fix(reconcile): honor product behavior, reject a credit amount without the credits behavior, and keep feature names verbatim on apply

rohilsurana:
## What

Two ways the BillingProduct file's stated intent was not what got applied:

1. **Behavior on create.** `Service.Create` forced `behavior = credits` whenever `credit_amount > 0`, ignoring the behavior the file set. Now an explicit behavior is honored; only an omitted behavior on a credit product falls back to `credits` (the server's create-time default). The reconciler's diff mirrors the same rule, so a file that names a behavior the product was not created with fails the plan instead of being silently overridden.

2. **Feature name case.** The diff compared feature names lowercased and trimmed, but the apply sent them raw. Since the server looks features up by name, a case difference read as "no change" yet could fork a duplicate feature on a co-occurring update. Both sides now use one `normalizeFeatureName` helper.

## Why

RFC 0001 Rule 2: the file is the desired state. The applied result must match what the file stated, with no silent override and no duplicate created behind a "no change" plan.

## Testing

- New `TestService_Create` cases: an explicit behavior is honored on a credit product, and an omitted behavior defaults to credits.
- Reworked the diff behavior test and added a body test asserting feature names are normalized on apply.
- `billing/...` and `internal/reconcile` pass.

## Stack

Stacked on `fix/reconcile-permission-spicedb-grammar`.

### DIFF
--- a/billing/product/service.go
+++ b/billing/product/service.go
@@ -64,8 +64,12 @@ func (s *Service) Create(ctx context.Context, product Product) (Product, error) 
 		product.ID = uuid.New().String()
 		product.ProviderID = product.ID
 	}
+	// Capture whether the caller stated a behavior before SetDefaults fills the
+	// "basic" default, so an explicit behavior is honored and only an omitted one
+	// on a credit product falls back to "credits".
+	statedBehavior := product.Behavior
 	defaults.SetDefaults(&product)
-	if product.Config.CreditAmount > 0 {
+	if statedBehavior == "" && product.Config.CreditAmount > 0 {
 		product.Behavior = CreditBehavior
 	}
 	product.Name = strings.ToLower(product.Name)
--- a/billing/product/service.go
+++ b/billing/product/service.go
@@ -72,6 +72,13 @@ func (s *Service) Create(ctx context.Context, product Product) (Product, error) 
 	if statedBehavior == "" && product.Config.CreditAmount > 0 {
 		product.Behavior = CreditBehavior
 	}
+	// Credits are only granted for the credits behavior, so a positive credit
+	// amount on any other behavior would store credits that no path ever awards.
+	// Reject the contradiction instead of silently keeping a product whose credits
+	// are dead.
+	if product.Config.CreditAmount > 0 && product.Behavior != CreditBehavior {
+		return Product{}, fmt.Errorf("%w: credit_amount %d needs behavior %q, got %q", ErrInvalidDetail, product.Config.CreditAmount, CreditBehavior, product.Behavior)
+	}
 	product.Name = strings.ToLower(product.Name)
 
 	_, err := s.stripeClient.Products.New(&stripe.ProductParams{
--- a/internal/reconcile/billingproduct_reconciler.go
+++ b/internal/reconcile/billingproduct_reconciler.go
@@ -288,9 +288,12 @@ func billingProductBody(s BillingProductSpec) *frontierv1beta1.ProductRequestBod
 	}
 	features := make([]*frontierv1beta1.Feature, 0, len(s.Features))
 	for _, f := range s.Features {
-		// Send the name in the same normalized form the diff compares, so the
-		// server matches an existing feature instead of forking a duplicate.
-		features = append(features, &frontierv1beta1.Feature{Name: normalizeFeatureName(f.Name)})
+		// Send the name verbatim. The server matches and stores feature names
+		// case-sensitively, so a normalized name here would miss a differently
+		// cased stored feature and fork a duplicate. The diff still compares
+		// normalized names so a case-only difference does not plan a spurious
+		// update.
+		features = append(features, &frontierv1beta1.Feature{Name: f.Name})
 	}
 	return &frontierv1beta1.ProductRequestBody{
 		Name:        s.Name,
--- a/internal/reconcile/billingproduct_reconciler.go
+++ b/internal/reconcile/billingproduct_reconciler.go
@@ -288,7 +288,9 @@ func billingProductBody(s BillingProductSpec) *frontierv1beta1.ProductRequestBod
 	}
 	features := make([]*frontierv1beta1.Feature, 0, len(s.Features))
 	for _, f := range s.Features {
-		features = append(features, &frontierv1beta1.Feature{Name: f.Name})
+		// Send the name in the same normalized form the diff compares, so the
+		// server matches an existing feature instead of forking a duplicate.
+		features = append(features, &frontierv1beta1.Feature{Name: normalizeFeatureName(f.Name)})
 	}
 	return &frontierv1beta1.ProductRequestBody{
 		Name:        s.Name,
--- a/billing/product/service_test.go
+++ b/billing/product/service_test.go
@@ -86,13 +86,13 @@ func TestService_Create(t *testing.T) {
 			},
 		},
 		{
-			name: "honors an explicit behavior on a credit product instead of forcing credits",
+			name: "honors an explicit credits behavior on a credit product",
 			args: args{
 				product: product.Product{
 					ID:          "creditprod",
 					Name:        "creditprod",
 					Description: "credit product",
-					Behavior:    product.PerSeatBehavior,
+					Behavior:    product.CreditBehavior,
 					Config:      product.BehaviorConfig{CreditAmount: 5},
 				},
 			},
@@ -100,7 +100,7 @@ func TestService_Create(t *testing.T) {
 				ID:          "creditprod",
 				Name:        "creditprod",
 				Description: "credit product",
-				Behavior:    product.PerSeatBehavior,
+				Behavior:    product.CreditBehavior,
 				Config:      product.BehaviorConfig{CreditAmount: 5},
 			},
 			wantErr: false,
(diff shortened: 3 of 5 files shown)


rohilsurana:
**Review: honor behavior on create, match feature-name case on apply (RFC 0001 Rule 2)**

Verified on this branch: `go build ./...`, `go test ./internal/reconcile/... ./billing/product/...`, and `golangci-lint run` on the changed packages all pass.

Both claims hold. Case normalization is now one shared function `normalizeFeatureName` (billingproduct.go:305), used by the diff (`billingFeatureSetsEqual`) and by apply (`billingProductBody`, billingproduct_reconciler.go:293), so apply cannot disagree with what the plan showed. The behavior fix captures `statedBehavior` before defaults (billing/product/service.go:67) and only folds to `credits` when the behavior is omitted, and the diff's `expectedBehavior` fold matches. I traced every create/diff/export combination and Rule 4 and Rule 5 both hold.

I also checked whether the same raw-case-on-apply gap remains for price names. It does not bite, because the billing service lowercases price names server-side on create and update (billing/product/service.go:456 and :518), unlike features. So leaving prices untouched is correct.

Verdict: approve with nits.

Nit:
1. The fix makes the diff and apply agree with each other, but not with the server's stored case. A pre-existing mixed-case feature (created via the `CreateFeature` API, or by the reconciler before this fix) still forks a duplicate when a co-occurring product change triggers an update. Concrete case: the server holds feature `Foo` on product `tok`, the file lists `foo` and also edits the description. The plan prints `update product tok (description)` with no feature change, but apply sends `foo`, the case-sensitive server lookup misses `Foo`, and a second `foo` row is created. This is narrow (legacy or externally created data) and a strict improvement over the old symmetric risk. A complete fix needs the server to canonicalize feature names on store and lookup.

rohilsurana:
Follow-up on the nit.

Agreed, and left as is on purpose. This PR makes the diff and the apply agree on feature-name case, which removes the symmetric fork risk. The residual case, a legacy or externally created mixed-case feature row, needs the server to canonicalize feature names on store and on lookup. That is a server change outside this reconcile PR, so it is a follow-up, not part of this change.

rohilsurana:
**Re-review (head 75aa9674)**

The head moved only because of the rebase onto #1889. #1890's own patch is byte-for-byte identical to the previous review, so nothing in this PR's content changed. Build, tests, and lint are all green.

Re-confirmed: feature-name case is normalized by one shared `normalizeFeatureName` on both the diff path and the apply path, so apply cannot disagree with the plan. The behavior-on-create fold matches the diff, and the export round-trip plans zero ops.

The residual mixed-case feature fork is still open and needs a server-side feature-name canonicalization (`UpsertFeature` stores the name raw, unlike prices which the service lowercases). That is out of scope for this PR.

Verdict: approve with nits.

internal/reconcile/billingproduct_reconciler.go
```diff
@@ -288,7 +288,9 @@ func billingProductBody(s BillingProductSpec) *frontierv1beta1.ProductRequestBod
 	}
 	features := make([]*frontierv1beta1.Feature, 0, len(s.Features))
 	for _, f := range s.Features {
-		features = append(features, &frontierv1beta1.Feature{Name: f.Name})
+		// Send the name in the same normalized form the diff compares, so the
+		// server matches an existing feature instead of forking a duplicate.
+		features = append(features, &frontierv1beta1.Feature{Name: normalizeFeatureName(f.Name)})
```
rohilsurana: Nit (residual, server-side follow-up): normalizing here makes apply agree with the diff, but the server's `UpsertFeature` stores the feature name raw (unlike prices, which the service lowercases). So a pre-existing mixed-case feature still forks a duplicate when a co-occurring product change triggers an update: the plan shows no feature change, but apply sends the lowercased name and the server lookup misses the stored mixed-case row. Fully closing this needs the server to canonicalize feature names on store and lookup, which is out of scope for this PR.

rohilsurana: Agreed, out of scope for this PR. This change makes the diff and the apply agree on case, which removes the symmetric fork. The residual, a pre-existing mixed-case row, needs the server to canonicalize feature names on store and lookup, the way it already does for prices. Tracking that as a server-side follow-up.

billing/product/service.go
```diff
@@ -64,8 +64,12 @@ func (s *Service) Create(ctx context.Context, product Product) (Product, error)
 		product.ID = uuid.New().String()
 		product.ProviderID = product.ID
 	}
+	// Capture whether the caller stated a behavior before SetDefaults fills the
+	// "basic" default, so an explicit behavior is honored and only an omitted one
+	// on a credit product falls back to "credits".
+	statedBehavior := product.Behavior
 	defaults.SetDefaults(&product)
-	if product.Config.CreditAmount > 0 {
+	if statedBehavior == "" && product.Config.CreditAmount > 0 {
```
rohilsurana, line 72: This broadens `Service.Create` for every caller, not just reconcile: a credit product (`credit_amount > 0`) can now be created with a stated non-credits behavior instead of being forced to `credits`. That is right for reconcile Rule 2 (honor the file), but it is a behavior change on the billing create path in general. Worth a quick check that no other caller relied on the old "a credit amount forces credits" rule.

rohilsurana: Checked both production callers of `Service.Create`, and this is safe to leave as is. The API create handler passes the request's behavior, and the plan sync passes the plan file's behavior. In both, an omitted behavior on a credit product still defaults to `credits`, so the common case is unchanged. The only input that changes is a contradictory one: an explicit non-credits behavior together with a positive credit amount. The old code silently overrode that to `credits`; the new code honors the stated behavior, which is the correct Rule 2 reading and is what the file and the API request should get. No caller relied on the silent override.

whoAbhishekSah (approved):

internal/reconcile/billingproduct_reconciler.go
AmanGIT07: Normalizing the feature name only on the apply side can fork a duplicate feature and orphan the original, which is the same problem this change is meant to prevent — just in the other direction.

The server matches and stores feature names verbatim and case-sensitively, unlike product and price names which it lowercases. `GetByName` uses exact equality, and the `features` table has `name text NOT NULL UNIQUE` with no `LOWER`. On every product update, `updateProductFeatures` unlinks each existing feature and re-adds each desired one by name (`AddFeatureToProduct` → `GetFeatureByID` → `GetByName`).

So for a product whose feature is stored as `TokenLimit`:

- Before this change, the apply sent `TokenLimit` (raw), which matched the stored row and re-linked it. No fork.
- After this change, the apply sends `tokenlimit`. `GetByName("tokenlimit")` misses the stored `TokenLimit`, so the server creates a new `tokenlimit` row and leaves `TokenLimit` orphaned — even when the diff reported features as unchanged and only, say, the description changed.

This triggers for any product that has a feature stored with a capital letter or trailing space, on the next update to that product.

The normalization is only safe if every stored feature name is already lowercase and trimmed, which the server does not guarantee. The complete fix is to normalize server-side in `UpsertFeature`/`GetByName`, the way product names (`service.go:75`) and price names (`service.go:460`) already are. Then both the diff and the apply match the stored form.

billing/product/service.go
AmanGIT07, line 72: Honoring an explicit behavior is right, but nothing rejects a non-credit behavior paired with `credit_amount > 0`, so this can now store a product whose credits are never granted.

Credits are only awarded when `Behavior == CreditBehavior` (`billing/checkout/service.go:634,1043`, `billing/invoice/service.go:130`). `validateBillingProductSpec` has no cross-field check, and the proto rule allows `per_seat` with a credit amount. So a product created with `behavior: per_seat, credit_amount: 100` is stored with a `credit_amount` that no path ever grants — the customer is billed per seat and never receives the 100 credits. Before this change that state was unreachable, because `credit_amount > 0` forced the behavior to `credits`.

Could we reject the combination (a non-credit behavior with `credit_amount > 0`) at create and in `validateBillingProductSpec`? If keeping it is intentional, a line in the docs saying the credit amount is ignored for non-credit products would at least make it visible.

internal/reconcile/billingproduct_reconciler.go
rohilsurana: Good catch, fixed. You are right that normalizing only on the apply side is unsafe: the server matches and stores feature names case-sensitively, so lowercasing what we send would miss a stored `TokenLimit` and fork a `tokenlimit` duplicate.

Reverted the apply-side normalization in `billingProductBody` so it sends the name verbatim. The diff still normalizes for its comparison, so a case-only difference does not plan a spurious update, but the apply now sends exactly what would match the stored row. Updated the `normalizeFeatureName` doc and the test to match.

The complete fix, so a file whose case differs from the stored row never forks either, is server-side canonicalization in `UpsertFeature`/`GetByName` the way product and price names already are. Keeping that as the tracked follow-up rather than widening this PR into the billing store and a data migration.

billing/product/service.go
rohilsurana: Agreed, fixed. Honoring the stated behavior without rejecting the contradiction did open a silent trap: `behavior: per_seat, credit_amount: 100` would store a product whose 100 credits are never granted.

Added a cross-field check in both places: `validateBillingProductSpec` fails the plan, and `Service.Create` rejects it with `ErrInvalidDetail` for the API path. An omitted behavior with a credit amount still coerces to `credits` as before, so only a stated non-credit behavior paired with `credit_amount > 0` is rejected. Added tests for the reject and for the omitted-behavior-allowed case.

PR #1891, opened 2026-08-17
docs(reconcile): declare disabled users out of scope for the PlatformUser kind

rohilsurana:
## What

A disabled user does not appear in the platform-user listing the reconciler reads (the store query filters out disabled accounts), so the reconciler cannot see or manage their platform access. This documents that disabled principals are out of scope, both in the reconcile docs and as a comment on `fetchCurrent`.

No behavior change.

## Why

This makes the existing behavior explicit so operators do not list a disabled principal and expect it to reconcile. A disabled admin is not exported and an empty file does not remove their grant. Re-enable the user first to manage their access.

## Stack

Stacked on `fix/reconcile-billingproduct-r2`.

### DIFF
--- a/internal/reconcile/platformuser_reconciler.go
+++ b/internal/reconcile/platformuser_reconciler.go
@@ -120,6 +120,12 @@ func (r *PlatformUserReconciler) Export(ctx context.Context) (any, error) {
 	return specs, nil
 }
 
+// fetchCurrent lists the platform principals the reconciler manages. Disabled
+// principals are out of scope: ListPlatformUsers resolves users through a store
+// query that filters out disabled accounts, so a disabled admin never appears
+// here. It is therefore neither exported nor removed by an empty file, and a file
+// that lists a disabled principal cannot converge. Re-enable the user to manage
+// their platform access.
 func (r *PlatformUserReconciler) fetchCurrent(ctx context.Context) ([]platformPrincipal, error) {
 	resp, err := r.client.ListPlatformUsers(ctx, authReq(&frontierv1beta1.ListPlatformUsersRequest{}, r.header))
 	if err != nil {
--- a/internal/reconcile/platformuser_reconciler.go
+++ b/internal/reconcile/platformuser_reconciler.go
@@ -121,11 +121,11 @@ func (r *PlatformUserReconciler) Export(ctx context.Context) (any, error) {
 }
 
 // fetchCurrent lists the platform principals the reconciler manages. Disabled
-// principals are out of scope: ListPlatformUsers resolves users through a store
+// users are out of scope: ListPlatformUsers resolves users through a store
 // query that filters out disabled accounts, so a disabled admin never appears
 // here. It is therefore neither exported nor removed by an empty file, and a file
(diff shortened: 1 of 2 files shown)


rohilsurana:
**Review: declare disabled principals out of scope for PlatformUser**

Verified on this branch: `go build ./...` and `go test ./internal/reconcile/...` pass.

For user principals, which is the reachable case, the doc is exactly right. Disabled users are filtered by `notDisabledUserExp` (internal/store/postgres/user_repository.go:238), so a disabled user is invisible to both `fetchCurrent` and `Export`. Since both share `fetchCurrent`, this is consistent with Rule 5.

Verdict: approve with nits.

Nits:
1. The prose says "disabled principals," but RFC 0001 defines a principal as a user or a service user, and the state filter exists only on the user path. `serviceuser_repository.go` `GetByIDs` has no state filter. So a disabled service user would appear in `fetchCurrent`, be exported, and be removed by an empty file, which is the opposite of the claim. This is latent only because there is currently no code path to disable a service user (no `Disable`/`SetState` for service users, unlike users). Either say "disabled users," or add a line that service users cannot be disabled yet.
2. Minor: this is not strictly docs-only. It also adds a 6-line comment to platformuser_reconciler.go:123-128. No executable code changes, so behavior is unaffected.

rohilsurana:
Follow-up on the nits.

- Wording: done. Changed "disabled principals" to "disabled users" and added a line that service users cannot be disabled today, so they are always listed. The state filter only exists on the user path, so the note now matches what the code actually does.
- Not strictly docs-only: intentional. The six-line comment sits at `fetchCurrent` so the scope decision is visible at the code site. There is no executable change, so behavior is the same.

rohilsurana:
**Re-review (head 345417c4)**

Build and tests are green. The over-generalization is fixed. The note now scopes the guarantee to users and correctly adds that service users cannot be disabled today, so they are always listed. Every claim checks out against the code: `notDisabledUserExp` filters disabled users out of the listing, and there is no disable path for service users, so they are never filtered.

One optional consistency nit: the `fetchCurrent` code comment (internal/reconcile/platformuser_reconciler.go:123-128) still says "principals" where the doc now says "users". Not wrong, just less precise than the doc. Not blocking.

Verdict: approve with nits.

internal/reconcile/platformuser_reconciler.go
```diff
@@ -120,6 +120,12 @@ func (r *PlatformUserReconciler) Export(ctx context.Context) (any, error) {
 	return specs, nil
 }
 
+// fetchCurrent lists the platform principals the reconciler manages. Disabled
+// principals are out of scope: ListPlatformUsers resolves users through a store
```
rohilsurana: Nit (optional consistency): the doc was narrowed to `users`, but this comment still says `principals`. It is not wrong, since only users can be disabled today, but consider matching the doc and saying `users` here so the two read the same.

rohilsurana: Fixed. The comment now says disabled users, matching the doc, and adds that service users cannot be disabled today. Pushed in 4648a875.

whoAbhishekSah (approved):

PR #1892, opened 2026-08-17
fix(reconcile): identify permissions by key instead of namespace and name

rohilsurana:
## What

The Permission reconcile kind now uses one `key` field instead of separate `namespace` and `name` fields. The key is the `service.resource.verb` value that the API already returns and stores.

Before:

```yaml
kind: Permission
spec:
  - namespace: compute/order
    name: get
```

After:

```yaml
kind: Permission
spec:
  - key: compute.order.get
```

## Why

The API's permission identity is the key. The reconciler used to read the deprecated `namespace` and `name` response fields and rebuild the key at apply time. Reading the key straight from the API is simpler, drops the deprecated-field use, and makes the file match what the server reports.

## Changes

- `PermissionSpec` is now `{ key, delete }`. The diff, validate, and export paths all key off the permission key.
- `fetchCurrent` reads `GetKey()` and parses it into namespace and verb only to skip base-schema permissions. A key that does not parse into `service.resource.verb` (a base or system permission, or an older server that does not set the key) is skipped, not fatal.
- Validation still runs fully before any apply. It rejects an empty key, a bad shape, a base-schema key, and a key whose service or resource part is not lowercase alphanumeric. That last check keeps the key one-to-one with the slug the server stores.
- Export emits `key` entries sorted by key, so a re-reconcile of an exported file plans nothing.
- Docs updated to the key form.

## Breaking change

This is a hard switch. A permission file that still uses `namespace` and `name` now fails the plan with an unknown-field error instead of silently reconciling nothing. Any pixxelhq/configs permission file must move to the key form in a separate configs change.

## Stacking

This PR is the top of the reconcile-audit stack: #1888 -> #1889 -> #1890 -> #1891 -> this. It builds on #1889, which tightens the permission grammar. Review and merge in stack order.

### DIFF
--- a/docs/content/docs/authz/custom-resources.mdx
+++ b/docs/content/docs/authz/custom-resources.mdx
@@ -65,6 +65,12 @@ A few rules the reconcile flow enforces for permissions:
 - A key has three parts joined by dots, `service.resource.verb`. Each part is lowercase,
   starts with a letter, and is at least three characters. So `compute.machine.get` works, but
   `compute.machine.get-all` does not (the hyphen is not alphanumeric).
+- The verb cannot be `owner`, `project`, or `granted`. The generated schema already defines
+  relations with those names on every resource, so a key like `compute.machine.owner` is
+  rejected even though it meets the rule above.
+- The whole flattened name `service_resource_verb` must be at most 64 characters, the limit
+  SpiceDB puts on a relation name. Three parts that are each fine on their own can still be
+  rejected together as too long.
 - A key whose service is `app` (an `app.*` key) is rejected, because the base schema owns
   those types (see [Why a separate `user/project` namespace](#why-a-separate-userproject-namespace)).
 
--- a/docs/content/docs/authz/custom-resources.mdx
+++ b/docs/content/docs/authz/custom-resources.mdx
@@ -31,21 +31,17 @@ You register the actions in one of two ways:
 - **The admin API** (`CreatePermission`, `CreateRole`). This makes the same change one call at
   a time, useful for one-off changes or scripting.
 
-Here is the `compute/machine` example as a reconcile file. Each entry is a namespace and an
-action name:
+Here is the `compute/machine` example as a reconcile file. Each entry is a key in
+`service.resource.verb` form:
 
 ```yaml
 apiVersion: v1
 kind: Permission
 spec:
-  - namespace: compute/machine
-    name: get
-  - namespace: compute/machine
-    name: create
-  - namespace: compute/machine
-    name: update
-  - namespace: compute/machine
-    name: delete
+  - key: compute.machine.get
+  - key: compute.machine.create
+  - key: compute.machine.update
+  - key: compute.machine.delete
 ```
 
 Apply it with a superuser credential (for example the bootstrap service account):
@@ -276,21 +272,16 @@ apiVersion: v1
 kind: Permission
 spec:
   # Per-item actions live on the resource itself, checked against compute/machine:<id>.
-  - namespace: compute/machine
-    name: get
-  - namespace: compute/machine
-    name: update
-  - namespace: compute/machine
-    name: delete
+  - key: compute.machine.get
+  - key: compute.machine.update
+  - key: compute.machine.delete
 
   # Project-level capabilities live on user/project, a proxy for app/project.
   # Checked against app/project:<project_id>, because there is no single
-  # machine to check against. Do NOT use namespace app/project here: the
-  # app namespaces are reserved for the base schema and are rejected.
-  - namespace: user/project
-    name: createcomputemachine
-  - namespace: user/project
-    name: listcomputemachines
+  # machine to check against. Do NOT use the app service here: the app
+  # namespaces are reserved for the base schema and are rejected.
+  - key: user.project.createcomputemachine
+  - key: user.project.listcomputemachines
 ---
 apiVersion: v1
 kind: Role
--- a/docs/content/docs/authz/custom-resources.mdx
+++ b/docs/content/docs/authz/custom-resources.mdx
@@ -59,15 +59,14 @@ permissions in this same format, so you can capture the live state into a file.
 
 A few rules the reconcile flow enforces for permissions:
 
-- A permission is **identity only** (namespace plus name). It is added or deleted, never
-  updated.
+- A permission is **identity only** (its key). It is added or deleted, never updated.
 - Nothing is deleted by leaving it out. A permission that is in the server but missing from
   the file fails the plan. To remove one, mark its entry with `delete: true`.
-- A namespace must be in `service/resource` form: two non-empty parts, each lowercase
-  alphanumeric. So `compute/machine` works, but `compute/machine-v2` does not (the hyphen is
-  not alphanumeric). The action name must be alphanumeric too.
-- A namespace under `app` or `app/...` is rejected, because the base schema owns those types
-  (see [Why a separate `user/project` namespace](#why-a-separate-userproject-namespace)).
+- A key has three parts joined by dots, `service.resource.verb`. Each part is lowercase,
+  starts with a letter, and is at least three characters. So `compute.machine.get` works, but
+  `compute.machine.get-all` does not (the hyphen is not alphanumeric).
+- A key whose service is `app` (an `app.*` key) is rejected, because the base schema owns
+  those types (see [Why a separate `user/project` namespace](#why-a-separate-userproject-namespace)).
 
 ### What happens when a permission is registered
 
--- a/docs/content/docs/reconcile.mdx
+++ b/docs/content/docs/reconcile.mdx
@@ -93,22 +93,21 @@ concerns users. Service users cannot be disabled today, so they are always liste
 
 ## The Permission kind
 
-A permission is an identity: a `service/resource` namespace plus a verb. There is nothing
+A permission is an identity: a key in `service.resource.verb` form. There is nothing
 else to manage on it, so it is either created or deleted.
 
 ```yaml
 apiVersion: v1
 kind: Permission
 spec:
-  - namespace: compute/order
-    name: get
-  - namespace: compute/order
-    name: legacy
+  - key: compute.order.get
+  - key: compute.order.legacy
     delete: true
 ```
 
-- The namespace has two parts, `service/resource`. The name is alphanumeric.
-- Frontier's own permissions (the `app` namespaces) belong to the base schema. The
+- The key has three parts joined by dots, `service.resource.verb`. Each part is lowercase
+  alphanumeric.
+- Frontier's own permissions (the `app` service) belong to the base schema. The
   reconciler ignores them and rejects a file entry that names one.
 - A custom permission on the server that the file does not list fails the plan. Nothing
   is deleted just because it is missing; deleting needs `delete: true` on the entry.
--- a/docs/content/docs/reconcile.mdx
+++ b/docs/content/docs/reconcile.mdx
@@ -105,8 +105,12 @@ spec:
     delete: true
 ```
 
-- The key has three parts joined by dots, `service.resource.verb`. Each part is lowercase
-  alphanumeric.
+- The key has three parts joined by dots, `service.resource.verb`. Each part is lowercase,
+  starts with a letter, and is at least three characters.
+- The verb cannot be `owner`, `project`, or `granted`. The generated schema already defines
+  relations with those names on every resource.
+- The whole flattened name `service_resource_verb` must be at most 64 characters, SpiceDB's
+  limit on a relation name.
 - Frontier's own permissions (the `app` service) belong to the base schema. The
   reconciler ignores them and rejects a file entry that names one.
 - A custom permission on the server that the file does not list fails the plan. Nothing
--- a/internal/reconcile/permission_reconciler.go
+++ b/internal/reconcile/permission_reconciler.go
@@ -32,8 +32,8 @@ func NewPermissionReconciler(client PermissionAPI, header string) *PermissionRec
 
 func (r *PermissionReconciler) Kind() string { return KindPermission }
 
-// Validate checks every entry, and the in-file slug conflicts, without touching
-// the server, so a bad entry stops the whole file before anything applies.
+// Validate checks every entry without touching the server, so a bad entry stops
+// the whole file before anything applies.
 func (r *PermissionReconciler) Validate(spec []byte) error {
 	var specs []PermissionSpec
 	if err := decodeSpec(spec, &specs); err != nil {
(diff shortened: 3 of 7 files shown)


rohilsurana:
**Review: identify permissions by key (RFC 0001 Rule 1)**

Verified on this branch: `go build ./...`, `go test ./internal/reconcile/...`, `go vet`, and `golangci-lint run` on the changed packages all pass.

The code achieves Rule 1 cleanly. `PermissionSpec` is now `{Key, Delete}`, `currentPermission` is `{ID, Key}`, the diff keys `byKey`, export emits `Key` sorted by key, create sends `Key`, and the old slug-collision guard is correctly removed. An old `{namespace, name}` file fails loudly through `decodeSpec`'s `KnownFields(true)` (covered by the new test), and that strictness pre-dates the stack, so this stands on its own. Rule 5 holds for valid keys.

The rebase onto the stack resolved the earlier overlap with #1889: this branch now sits on top of #1889 and inherits its tightened grammar (the verb error even carries #1889's wording). The shared permission-validation items (the CreatePermission API not validating namespace grammar, and the combined slug-length bound) are tracked in #1889 so they are not duplicated here. Fixing them there fixes them for this PR too.

Verdict: changes needed. The code is approvable; the blocker is a doc file this PR claims to have migrated but did not.

Findings:
1. Blocking (docs). `docs/content/docs/authz/custom-resources.mdx` still teaches the old `namespace:` plus `name:` Permission format in two reconcile examples (around lines 37-48 and 274-293). The PR description says the docs were moved to the key form, but only `reconcile.mdx` was. A reader who copies the `user/project` + `createcomputemachine` example and runs reconcile now gets an unknown-field error. Convert those entries to the key form, for example `- key: compute.machine.get` and `- key: user.project.createcomputemachine`, and update the surrounding prose ("a namespace and an action name") to describe the key.
2. Minor. `fetchCurrent` hard-errors the whole list, and therefore export too, if any single `GetKey()` does not split into three dot-parts, and this runs before the base-namespace skip. The old path read namespace and name and degraded gracefully. Consider tolerating an unparseable key (skip it, or only fail for a non-base one). This is exactly where an older or different server could bite, which ties into the #1887 overlap on `fetchCurrent`: whichever of #1887 and this PR lands second should confirm the server's `Key` is always populated and canonical so this assumption stays safe.

rohilsurana:
Follow-up on the findings.

- Docs: done. Migrated both Permission examples in `authz/custom-resources.mdx` and the surrounding prose to the key form. There are no `namespace:`/`name:` Permission examples left in the docs.
- Unparseable key: done. `fetchCurrent` now skips a permission whose key does not split into `service.resource.verb`, rather than failing the whole list. That covers a base or system permission and an older server that does not set the key. Export only emits keys it kept, so a skipped permission never shows up as missing from the file. Added a test.

Also, the stack was rebased so this PR now sits on top of #1889 and inherits its tightened grammar and the new slug-length bound, so those apply to the key form too.

rohilsurana:
**Re-review (head 6bdbd097)**

Build, tests, and lint are all green. Both code findings are resolved.

- Docs examples: the two `kind: Permission` examples are correctly migrated to the key form, every conversion is right (`a/b` + `c` becomes `a.b.c`), and no `namespace:`/`name:` entry remains in any Permission spec block. The remaining `name:` lines are Role identities, which are legitimate.
- `fetchCurrent` skip: sound. It skips only unparseable or base keys, and a valid custom key always has three parts, so no manageable permission is silently dropped. Rules 4 and 5 still hold: skipped permissions never enter `current`, so they are never deleted by omission, and export skips them too, so the round trip plans zero ops.

One doc item remains, so this is not fully closed. The bullet list "A few rules the reconcile flow enforces for permissions" at `docs/content/docs/authz/custom-resources.mdx:60-70` still teaches the old field model, directly under the now-fixed key examples:

- ":61" "A permission is identity only (namespace plus name)."
- ":66-68" "A namespace must be in `service/resource` form: two non-empty parts, each lowercase alphanumeric... The action name must be alphanumeric too."

A reader who builds a file from that prose, rather than copying the example above it, would still write `namespace`/`name` and hit an unknown-field error. Update that list to the key form to match the examples and the already-updated `reconcile.mdx`. The slug-mechanics line at :100-102 ("namespace `compute/machine` with action `get` becomes `compute_machine_get`") is describing SpiceDB slug generation, so it is acceptable, though it could mention the key for consistency.

Verdict: approve with nits, once the bullet list is updated.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -31,21 +31,17 @@ You register the actions in one of two ways:
...
 spec:
-  - namespace: compute/machine
-    name: get
-  - namespace: compute/machine
-    name: create
-  - namespace: compute/machine
-    name: update
-  - namespace: compute/machine
-    name: delete
+  - key: compute.machine.get
```
rohilsurana, line 41: Nit (docs, anchored here because the affected lines are unchanged): this example is correctly on the key form now, but the `rules the reconcile flow enforces` bullet list just below (lines 62-68) still describes the old field model: `A permission is identity only (namespace plus name)` and `A namespace must be in service/resource form ... The action name must be alphanumeric too`. A reader who builds a file from that prose, rather than copying this example, would still write `namespace`/`name` and hit an unknown-field error. Update that list to the `service.resource.verb` key form to match these examples and the already-updated reconcile.mdx.

rohilsurana: Good catch, fixed. The "rules the reconcile flow enforces" list now describes the key form: identity is the key, a key is `service.resource.verb` with each part lowercase and at least three characters, and an `app.*` key is rejected. No `namespace`/`name` prose is left. Pushed in 04c4f1a7.

internal/reconcile/permission_reconciler.go
```diff
@@ -117,16 +107,17 @@ func (r *PermissionReconciler) fetchCurrent(ctx context.Context) ([]currentPermi
...
-	//nolint:staticcheck
 	for _, p := range resp.Msg.GetPermissions() {
-		if isBaseNamespace(p.GetNamespace()) {
+		ns, name := schema.PermissionNamespaceAndNameFromKey(p.GetKey())
+		// A key that does not split into service.resource.verb is not a custom
+		// permission this reconciler manages: it is a base or system permission, or
+		// comes from an older server that does not set the key. Skip it instead of
+		// failing the whole list. Export only emits keys it kept here, so a skipped
+		// permission never shows up as missing from the file either.
+		if ns == "" || name == "" || isBaseNamespace(ns) {
```
rohilsurana, line 117: Merge note against current `main`. #1887 is already merged, and it rewrote this same `fetchCurrent`. Its version fails loudly on an unparseable key and keeps `currentPermission{Namespace, Name}`; this version skips and keeps `{Key}`. Rebasing this PR onto `main` hits a content conflict right here (confirmed with `git merge-tree origin/main <tip>`).

Two things to handle when resolving:

1. Keep this skip + `{Key}` design.
2. Delete #1887's test `a permission whose key does not split fails instead of being misread` in `permission_reconciler_test.go`. Git auto-merges that test with no conflict marker, but it then fails: it uses the old `namespace/name` spec and asserts a `notakey` error this design never produces. It also contradicts this PR's own `a server permission with an unparseable key is skipped, not fatal`.

One more: the PR description says fetchCurrent "fails loudly if the key does not parse", which is the opposite of what this does. Please update the body to say it skips.

rohilsurana:
Doc consistency (not inline, since the file is not in this diff): this PR updates `reconcile.mdx` and `authz/custom-resources.mdx` to the key form, but `docs/rfcs/0001-declarative-reconcile.md` still describes the Permission entry as `{namespace, name}` (line 54 in the "At a glance" table, and line 210 "An entry is `{namespace, name}`, for example `compute/order` + `get`"). Since that RFC is the spec this stack conforms to, it should move to the key form too. Otherwise a reader following the RFC writes `namespace:/name:` and the file fails with an unknown-field error.

internal/reconcile/permission_reconciler.go
rohilsurana: Handled. Rebased the whole stack onto current main, which now has #1887. Resolved the `fetchCurrent` conflict keeping this PR's skip + `{Key}` design, and deleted #1887's test `a permission whose key does not split fails instead of being misread`: it used the old `namespace`/`name` spec and asserted a `notakey` error this design does not produce. Build, tests, and lint are green on the new base. Also fixed the PR description: `fetchCurrent` skips an unparseable key, it does not fail loudly. Head is 8c2fbdc7.

rohilsurana:
Fixed. `docs/rfcs/0001-declarative-reconcile.md` now uses the key form: the "At a glance" table lists the Permission identity as `key`, and the entry line reads `{key}` in `service.resource.verb` form, for example `compute.order.get`. No `{namespace, name}` Permission references are left in the RFC. Pushed in 8c2fbdc7.

whoAbhishekSah (approved):

internal/reconcile/permission_reconciler.go
```diff
@@ -119,17 +109,15 @@ func (r *PermissionReconciler) fetchCurrent(ctx context.Context) ([]currentPermi
...
-		if ns == "" || name == "" {
-			return nil, fmt.Errorf("permission %s: cannot split key %q into namespace and name", p.GetId(), p.GetKey())
-		}
-		if isBaseNamespace(ns) {
+		// A key that does not split into service.resource.verb is not a custom
+		// permission this reconciler manages: it is a base or system permission, or
+		// comes from an older server that does not set the key. Skip it instead of
+		// failing the whole list. Export only emits keys it kept here, so a skipped
+		// permission never shows up as missing from the file either.
+		if ns == "" || name == "" || isBaseNamespace(ns) {
```
AmanGIT07, line 117: This skip keeps any current permission whose key splits into three non-empty, non-base parts, but it never re-checks the grammar. A permission whose stored namespace has an underscore or an uppercase letter (say `resource/order_item` + `get`) round-trips to a parseable key `resource.order_item.get`, so it is kept in `current` — yet no file entry can name it, because `validatePermissionSpec` rejects that key at `ValidateCustomPermission`. That wedges the plan with no way out:

- Leaving it out of the file: it sits in `byKey` but never in `seen`, so the `unaccounted` check fails every plan with "permissions exist on the server but are not in the file".
- Adding `{key: resource.order_item.get, delete: true}` to remove it: the spec fails validation first (underscore in a part), so the entry is rejected before it can delete anything.

So the permission can neither be kept nor deleted through reconcile, and every permission plan fails permanently. `Export` also emits that key, so re-running an exported file errors instead of planning nothing — which breaks the "reconciling an exported document plans no changes" invariant.

New rows can't get into this state now that `CreatePermission` runs `ValidateCustomPermission`, but rows created before that tightening can. Since this function is the one deciding what is in scope, the fix fits right here: also skip a current permission whose parsed `(ns, name)` fails `schema.ValidateCustomPermission`, treating an unmanageable legacy row as out of scope the same way a base permission is.

docs/content/docs/reconcile.mdx
```diff
@@ -93,22 +93,21 @@ concerns users. Service users cannot be disabled today, so they are always liste
...
-  - namespace: compute/order
-    name: legacy
+  - key: compute.order.get
+  - key: compute.order.legacy
     delete: true
 ```
 
-- The namespace has two parts, `service/resource`. The name is alphanumeric.
-- Frontier's own permissions (the `app` namespaces) belong to the base schema. The
+- The key has three parts joined by dots, `service.resource.verb`. Each part is lowercase
```
AmanGIT07: "lowercase alphanumeric" is looser than what the reconciler actually enforces. Each part must also start with a letter and be at least three characters (`ValidateCustomPermission` → `^[a-z][a-z0-9]{2,63}$`). As written, this bullet reads as if `compute.order.2fa` (leading digit) or `ab.cd.ef` (short parts) are fine, but the plan rejects them.

`custom-resources.mdx` states the fuller rule ("lowercase, starts with a letter, at least three characters"), so the two docs now describe the same check differently. Matching this bullet to that wording keeps them in step.

docs/content/docs/authz/custom-resources.mdx
```diff
@@ -63,15 +59,14 @@ permissions in this same format, so you can capture the live state into a file.
...
-  updated.
+- A permission is **identity only** (its key). It is added or deleted, never updated.
 - Nothing is deleted by leaving it out. A permission that is in the server but missing from
   the file fails the plan. To remove one, mark its entry with `delete: true`.
-- A namespace must be in `service/resource` form: two non-empty parts, each lowercase
-  alphanumeric. So `compute/machine` works, but `compute/machine-v2` does not (the hyphen is
-  not alphanumeric). The action name must be alphanumeric too.
-- A namespace under `app` or `app/...` is rejected, because the base schema owns those types
-  (see [Why a separate `user/project` namespace](#why-a-separate-userproject-namespace)).
+- A key has three parts joined by dots, `service.resource.verb`. Each part is lowercase,
```
AmanGIT07, line 65: Two rules the reconciler enforces are missing from this list (and from the matching one in `reconcile.mdx`):

- The verb cannot be `owner`, `project`, or `granted`. The generated schema already defines relations with those names on every resource, so a key like `compute.machine.owner` is rejected with "verb owner is reserved" even though it meets every rule stated here.
- The whole `service_resource_verb` slug is capped at 64 characters, so three otherwise-valid parts can still be rejected as "too long".

A reader building a key from this prose would hit both with no warning. A short line on each in both docs would cover it.

internal/reconcile/permission_reconciler.go
rohilsurana: Fixed here, as you suggested. `fetchCurrent` now also skips a current permission whose parsed `(ns, name)` fails `schema.ValidateCustomPermission`, the same way it already skips a base or unparseable key.

So a legacy row like `resource.order_item.get` (underscore in a part) is out of scope: it no longer counts as unaccounted, and export no longer emits it, so re-running an exported file still plans nothing. Added a test with that exact row. As you noted, new rows cannot reach this state now that `CreatePermission` runs the same check; this covers the rows created before that tightening.

docs/content/docs/reconcile.mdx
rohilsurana: Fixed. The `reconcile.mdx` bullet now says each part is lowercase, starts with a letter, and is at least three characters, matching the wording in `custom-resources.mdx` and the actual `^[a-z][a-z0-9]{2,63}$` check. So `compute.order.2fa` and `ab.cd.ef` no longer read as allowed.

docs/content/docs/authz/custom-resources.mdx
rohilsurana: Added both rules to both docs. Each list now says the verb cannot be `owner`, `project`, or `granted` (they collide with the relations the generated schema adds to every resource), and that the flattened `service_resource_verb` is capped at 64 characters. So a reader building a key from the prose sees both up front.

PR #1894, opened 2026-08-18
feat(deleter): serve CheckOrganizationDelete for delete eligibility

whoAbhishekSah:
Part of https://github.com/raystack/frontier/issues/1837. Top of the stack, based on #1893. Proto: https://github.com/raystack/proton/pull/499.

The product decision for the delete button: disabled while any of the three blockers exist (active subscription on a paid plan, unpaid invoice, negative token balance), enabled on the unused-tokens case and the happy path. The client cannot compute the paid-plan case itself — it does not know the server's configured standard plan — so this adds a read-only check RPC.

`CheckOrganizationDelete` returns the same blocker list that `DeleteOrganization`'s `failed_precondition` carries, without changing anything. Each blocker has `type` (the reason), `subject` (the blocking entity's id), `subject_type` (what kind of entity that id is: `billing_subscription`, `billing_invoice`, or `billing_account`), and `message`.

- No subscription gets canceled (the delete only cancels standard-plan subscriptions once everything else is clear).
- Invoice rows are read as they are — no forced provider sync per check, since clients may call this on every settings page view. The answer is advisory; the delete still re-checks against fresh provider data.
- Requires the same `delete` permission on the org as the delete itself.
- A missing org returns `not_found`.

The SDK will switch its delete button to this RPC once the proton npm package ships with it; until then #1881 greys the button on the invoice case only.

### DIFF
--- a/core/deleter/deleter.go
+++ b/core/deleter/deleter.go
@@ -11,12 +11,22 @@ const (
 	BlockerNegativeTokenBalance = "NEGATIVE_TOKEN_BALANCE"
 )
 
+// Kinds of entity a Blocker's Subject id refers to.
+const (
+	SubjectSubscription   = "billing_subscription"
+	SubjectInvoice        = "billing_invoice"
+	SubjectBillingAccount = "billing_account"
+)
+
 // Blocker is one reason an organization cannot be deleted right now.
 type Blocker struct {
 	// Type is one of the Blocker* constants.
 	Type string
 	// Subject is the id of the entity behind the reason.
 	Subject string
+	// SubjectType is one of the Subject* constants and says what kind of
+	// entity the Subject id refers to.
+	SubjectType string
 	// Message says what blocks the delete and how to clear it.
 	Message string
 }
--- a/core/deleter/service.go
+++ b/core/deleter/service.go
@@ -526,38 +526,15 @@ func (d Service) DeleteUser(ctx context.Context, userID string) error {
 // their subscription and invoice rows have nothing behind them the caller
 // could cancel or pay.
 func (d Service) ensureDeletable(ctx context.Context, id string, customers []customer.Customer, balances map[string]int64) error {
-	// each plan resolves at most once per call, and only when a running
-	// subscription actually references it
+	// each plan resolves at most once per delete — the cache is shared with
+	// the cancel pass below — and only when a running subscription
+	// references it
 	paidPlans := map[string]bool{}
 
-	var blockers []Blocker
-	for _, c := range customers {
-		if !c.IsOffline() {
-			bs, err := d.subscriptionBlockers(ctx, c, paidPlans)
-			if err != nil {
-				return err
-			}
-			blockers = append(blockers, bs...)
-
-			bs, err = d.invoiceBlockers(ctx, c)
-			if err != nil {
-				return err
-			}
-			blockers = append(blockers, bs...)
-		}
-
-		balance := balances[c.ID]
-		// the balance goes below zero when the account has an overdraft
-		// floor (credit_min under zero, the postpaid setup) and tokens were
-		// spent on credit. That debt is money owed, so it must be settled
-		// before the org can go
-		if balance < 0 {
-			blockers = append(blockers, Blocker{
-				Type:    BlockerNegativeTokenBalance,
-				Subject: c.ID,
-				Message: fmt.Sprintf("billing account[%s] owes %d tokens: contact support to settle the balance, then retry the delete", c.ID, -balance),
-			})
-		}
+	// the delete is about to happen, so judge invoices on fresh provider data
+	blockers, err := d.collectBlockers(ctx, customers, true, balances, paidPlans)
+	if err != nil {
+		return err
 	}
 	if len(blockers) > 0 {
 		return &BlockedError{OrgID: id, Blockers: blockers}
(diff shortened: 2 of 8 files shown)


rohilsurana (commented):
Stack-aware review (base org-delete-check-proto). Main thing is the check vs delete mismatch below; the other two are small. I dropped an earlier note about the NotFound mapping: it is only reachable for non-existent orgs, which a real client will not hit.

core/deleter/service.go
```diff
@@ -661,14 +695,29 @@ func (d Service) isPaidPlan(ctx context.Context, planID string, cache map[string
...
+// just-paid invoice does not block a delete and a just-created one does.
+// Without it the local rows answer, cheap enough for every check call.
+func (d Service) invoiceBlockers(ctx context.Context, c customer.Customer, fromProvider bool) ([]Blocker, error) {
+	var invoices []invoice.Invoice
+	var err error
+	if fromProvider {
+		invoices, err = d.invoiceService.ListPayableOnProvider(ctx, c)
+	} else {
+		var all []invoice.Invoice
+		all, err = d.invoiceService.List(ctx, invoice.Filter{CustomerID: c.ID, NonZeroOnly: true})
```
rohilsurana: The check path reads local rows with NonZeroOnly (amount > 0), while the delete path (ListPayableOnProvider) keeps any invoice whose Total is not 0. So the two disagree on a negative-total invoice: the check drops it and reports can_delete, but the real delete keeps it and blocks. The client greys nothing, the user clicks delete, and the delete is refused, which is the opposite of what the check promised. The root filter is in #1857 (Total == 0). Keying both off the amount actually due would keep the check and the delete in step.

core/deleter/service.go
```diff
@@ -661,14 +695,29 @@ func (d Service) isPaidPlan(ctx context.Context, planID string, cache map[string
...
+		for _, inv := range all {
```
rohilsurana, line 768: This ranges over all before the err from List is checked a few lines down. On the usual (nil result, err) contract it is a harmless no-op, but if List ever returns partial rows alongside an error, those rows get filtered before the error surfaces. Move the err check above the loop, like the provider branch does.

core/deleter/service.go
```diff
@@ -563,6 +532,7 @@ func (d Service) ensureDeletable(ctx context.Context, id string) error {
 	// runs only after every blocker is clear and never touches a paid
 	// subscription, so a blocked delete never costs the caller a plan they
 	// pay for
+	paidPlans := map[string]bool{}
```
rohilsurana: collectBlockers now keeps its own paidPlans map and this cancel loop starts a fresh one, so each free-plan subscription's plan is resolved twice per delete: once while collecting blockers, once here. It is minor, and partly the cost of making collectBlockers self-contained for the check path, but threading one cache through both would avoid the extra GetByID calls.

core/deleter/service.go
whoAbhishekSah: Aligned for the case you found: the provider path now keys off AmountRemaining (see the fix on the base PR), so a negative-total invoice no longer blocks the delete, matching the check which never surfaced it. One asymmetry remains in the other, safe direction: the local rows only store the total, so an open invoice fully covered by the customer's credit balance can still grey the button while the actual delete would pass. Storing amount_due locally would close that, but it touches the sync and the table, so I left it out of this stack.

rohilsurana (commented):
Re-review after the fix commits. The err-before-loop fix and the shared paidPlans/balances cache both look good, and the negative-total case now lines up. I left the credit-covered greying alone since you already flagged it as a deferred, safe-direction gap. Two small cleanup/doc notes.

core/deleter/service.go
```diff
@@ -693,14 +739,31 @@ func (d Service) isPaidPlan(ctx context.Context, planID string, cache map[string
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to check invoices for billing account[%s]: %w", c.ID, err)
+		}
+	} else {
+		all, err := d.invoiceService.List(ctx, invoice.Filter{CustomerID: c.ID, NonZeroOnly: true})
+		if err != nil {
+			return nil, fmt.Errorf("failed to check invoices for billing account[%s]: %w", c.ID, err)
+		}
+		for _, inv := range all {
+			if inv.State == invoice.DraftState || inv.State == invoice.OpenState || inv.State == invoice.UncollectibleState {
```
rohilsurana: The payable states {draft, open, uncollectible} are now written in two places: here in the local branch and in ListPayableOnProvider (billing/invoice/service.go). If a payable state is ever added or dropped, both spots have to change together, and missing one silently makes the check and the real delete disagree on what blocks. A shared invoice.PayableStates slice or an inv.IsPayable() helper would keep them in step.

core/deleter/deleter.go
```diff
@@ -11,12 +11,23 @@ const (
 	BlockerNegativeTokenBalance = "NEGATIVE_TOKEN_BALANCE"
 )
 
+// Kinds of entity a Blocker's Subject id refers to. They match the naming
```
rohilsurana: This comment says the Subject kinds match the naming the audit records use, but only billing_account maps to a real audit target (alongside billing_checkout). There are no billing_subscription or billing_invoice audit targets, so the comment overstates the match for two of the three. Worth trimming the comment so a later reader does not go looking for audit targets that do not exist.

rohilsurana (approved):

PR #1895, opened 2026-08-19
fix(sdk): scope PAT detail to active org and show a unified not-found page

rohilsurana:
## Problem

The current-user PAT management flows (get, update, delete, regenerate) scope the token lookup by the logged-in user, not by the org. A user who belongs to more than one org could open, edit, regenerate, or delete a token that belongs to another org just by changing the org in the URL. The token also rendered under any org the user is a member of.

The PAT detail view is where all of this happens. It fetches the token by id and hosts the Update, Regenerate, and Revoke actions.

## Fix

A client-side guard in the PAT detail view. The PAT response already carries its `org_id`, so we compare it with the active org from the route.

- If the token belongs to a different org, or the id is unknown or malformed (NotFound / InvalidArgument), show a generic "Token not found" state with no actions. The same message covers all of these, so it reveals nothing about tokens in other orgs.
- Any other failure (network, server error, and so on) shows a separate "Something went wrong" state with a Try again button. We do not claim the token is missing when the real problem is transient.

Retry is tuned to match. A NotFound or invalid id is a definite answer and is not retried, so the not-found state shows right away. Other failures may be transient, so they are retried a couple of times before the error state appears. Before this, the query used the default retry of 3, which retried even a 404 three times.

No proto or server change is needed. The list view already scopes its search by org.

## Testing

Ran a local sandbox with one user in two orgs, each with a PAT, and checked:

- Token opened in its own org: details render as before.
- Token opened under a different org: "Token not found", no actions.
- Unknown or malformed token id: same "Token not found", shown right away with no retries.
- Server made unreachable: "Something went wrong" with a Try again button, and Try again recovers once the server is back.

## Notes

This is a UI guard. A direct RPC call is still not org-scoped on the server, which is a separate, deeper change.

### DIFF
--- a/web/sdk/client/views/pat/pat-details-view.tsx
+++ b/web/sdk/client/views/pat/pat-details-view.tsx
@@ -1,8 +1,9 @@
 'use client';
 
-import { ReactNode, useCallback, useEffect, useMemo } from 'react';
+import { ReactNode, useCallback, useMemo } from 'react';
 import {
   DotsHorizontalIcon,
+  LockClosedIcon,
   Pencil1Icon,
   UpdateIcon
 } from '@radix-ui/react-icons';
@@ -11,13 +12,13 @@ import {
   Breadcrumb,
   Button,
   Dialog,
+  EmptyState,
   Flex,
   IconButton,
   Image,
   Menu,
   Skeleton,
   Text,
-  toastManager,
   Tooltip
 } from '@raystack/apsara';
 import deleteIcon from '../../assets/delete.svg';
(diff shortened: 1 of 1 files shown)


rohanchkrabrty (approved):

web/sdk/client/views/pat/pat-details-view.tsx
```diff
@@ -238,6 +241,69 @@ export function PATDetailsView({
...
+            <Breadcrumb size="small">
+              <Breadcrumb.Item
+                onClick={() => onNavigateToPats?.()}
+                data-test-id="frontier-sdk-pat-not-found-breadcrumb"
+              >
+                Personal access token
+              </Breadcrumb.Item>
+            </Breadcrumb>
+          }
+        />
```
rohanchkrabrty: `ViewHeader` should be skipped for error states

rohanchkrabrty, line 267: `ViewHeader` should be skipped for error states

rohilsurana: fixed

PR #1898, opened 2026-08-20
fix(billing): omit empty product description for Stripe

### LINKED ISSUE #1853: Plans and products created via the API are broken: invisible to ListPlans and fail on empty description
Found while live-testing #1849/#1850/#1851. Both problems are in the API creation path (`CreatePlan` / `CreateProduct`) and exist on `main` — the plans-file bootstrap path is not affected.

## 1. Plans get an empty `state` and become invisible

`CreatePlan` saves the plan without setting `state`. Every plan lookup that goes through the repository `List` filter defaults to `state = 'active'` (`internal/store/postgres/billing_plan_repository.go`), so an API-created plan:

- does not show up in `ListPlans` (returns empty even though `GetPlan` by id works)
- breaks `ChangeSubscription`: resolving the plan from the Stripe schedule uses `planService.List`, so it fails with `no plan found for phase products: [...], interval: month` and the RPC returns `internal`
- makes the background `subscription.SyncWithProvider` job log the same error every minute, forever, for any subscription on such a plan

Reproduce:
1. `CreatePlan` via the API with one product and one price
2. `ListPlans` → empty; `select state from billing_plans` → empty string
3. Subscribe to the plan, then `ChangeSubscription` to another plan → `internal`

Workaround: `update billing_plans set state='active' where name=...`

Fix idea: set `state = 'active'` as a default in the plan service `Create`/upsert path (the same way prices already get `state = 'active'`).

## 2. Empty product description fails at Stripe

`productService.Create` always sends `Description: &product.Description` to Stripe. When the caller leaves the description empty, Stripe rejects it with `parameter_invalid_empty` ("You passed an empty string for 'description'"), and `CreatePlan`/`CreateProduct` return `internal`.

Reproduce: `CreateProduct` via the API without a `description` → `internal`.

Fix idea: only set the description param when it is non-empty.

RushikeshBhavsar3605:
## Summary

Fix product creation failing when the product description is empty.

## Changes

- Omit the Stripe `description` parameter when the product description is empty.
- Preserve the existing behavior when a description is provided.
- Add a regression test covering empty product descriptions.

## Test Plan

- `go test ./billing/product`
- Reproduced the issue end-to-end using a Stripe test account.
- Verified that creating a product with an empty description now succeeds.

Closes #1853

### DIFF
--- a/billing/product/service.go
+++ b/billing/product/service.go
@@ -81,13 +81,12 @@ func (s *Service) Create(ctx context.Context, product Product) (Product, error) 
 	}
 	product.Name = strings.ToLower(product.Name)
 
-	_, err := s.stripeClient.Products.New(&stripe.ProductParams{
+	providerParams := &stripe.ProductParams{
 		Params: stripe.Params{
 			Context: ctx,
 		},
-		ID:          &product.ProviderID,
-		Name:        &product.Title,
-		Description: &product.Description,
+		ID:   &product.ProviderID,
+		Name: &product.Title,
 		Metadata: map[string]string{
 			"name":          product.Name,
 			"credit_amount": fmt.Sprintf("%d", product.Config.CreditAmount),
@@ -95,7 +94,13 @@ func (s *Service) Create(ctx context.Context, product Product) (Product, error) 
 			"product_id":    product.ID,
 			"managed_by":    "frontier",
 		},
-	})
+	}
+
+	if product.Description != "" {
+		providerParams.Description = &product.Description
+	}
+
+	_, err := s.stripeClient.Products.New(providerParams)
 	if err != nil {
 		return Product{}, fmt.Errorf("failed to create product at billing provider: %w", billingerrors.TranslateStripeError(err))
 	}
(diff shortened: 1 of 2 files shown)


billing/product/service.go
rohilsurana, line 223: We need the same check here as well.

rohilsurana:
Are there any other stripe fields that have similar behaviour and we need to take care of?

billing/product/service.go
RushikeshBhavsar3605: I found the same pointer usage in Update() and looked into applying the same guard there.

One difference I verified: Stripe accepts `description=""` on product update and clears the existing description, while omitting the field leaves it unchanged. So applying the same check would prevent clearing an existing description.

Should Update() preserve the ability to clear the description, or should an empty description be treated as omitted here as well?

RushikeshBhavsar3605:
I checked the Stripe product/price fields for the same empty-description behavior and didn't find another occurrence.

billing/product/service.go
RushikeshBhavsar3605: Just following up on this. I verified that Stripe accepts `description=""` on product update and uses it to clear the description, so applying the same guard as `Create()` would change the current update behavior.

Please let me know how you'd like `Update()` to handle an empty description, and I'll update the PR accordingly.

rohilsurana: we can keep the current behavior to clear desc field.

rohilsurana (approved):

PR #1903, opened 2026-08-25
feat(deleter): send the delete notice on every organization delete

whoAbhishekSah:
Follow-up to #1880. The owner email used to go out only when the delete forfeited unused tokens. Every deleted organization now notifies **all of its owners**, tokens or not.

- The owners are resolved after the blocker check passes and before teardown removes their policies — best-effort, so the delete never depends on the policy machinery. A delete whose owners cannot be resolved logs a warning instead of failing. (An earlier commit persisted the recipient set in an audit record for retries; it was reverted after review — that pushed owner ids onto the webhook feed to cover a failure window of two local statements at the tail of teardown. A delete failing exactly there and completing on retry logs that no owner could be notified.)
- Because the mail is no longer about forfeits, the config key moves from `billing.token_forfeit_notice` to `billing.org_delete_notice`. Nothing has deployed the old key yet, so this is a rename, not a migration.
- The built-in default template is a plain deletion notice. It names **no token amounts** (settlement numbers have tax implications) and adds one settlement line only when purchased tokens existed. `.Amount` and `.Purchased` stay available to config templates for deployments that want them.
- Nothing changes for the audit trail: `app.billing.tokens.forfeited` still records the exact per-account `amount` and `purchased`, and the per-account recovery on retried deletes still fills the template data. The webhook event list and its docs page now name `app.billing.tokens.forfeited` and `app.billing.checkout.deleted`, which audit publishing already sends to subscribers.
- Tests: the forfeit record and its retry recovery now run against a real audit service with an in-memory repository (record content asserted on the first attempt, notice rebuilt from the record on a retry), and every successful-delete test asserts the owner mail.

## How support reconstructs a refund after the org is gone

The delete hard-deletes the org's ledger rows, so the forfeit audit record is the number support settles from: `amount` (balance at delete) and `purchased` (the refundable share) per billing account. If the record is ever in doubt, this query rebuilds what the surviving remnants still know — the platform-side halves of purchase entries carry a `checkout_id`, and the checkout's deletion audit record remembers the org:

```sql
WITH deleted_checkouts AS (
  SELECT a.org_id,
         a.target->>'ID'            AS checkout_id,
         a.metadata->>'customer_id' AS account_id,
         a.metadata->>'provider_id' AS stripe_session
  FROM auditlogs a
  WHERE a.action = 'app.billing.checkout.deleted'
),
bought AS (
  SELECT dc.org_id, dc.account_id,
         sum(t.amount) AS tokens_bought
  FROM deleted_checkouts dc
  JOIN billing_transactions t
    ON t.metadata->>'checkout_id' = dc.checkout_id
   AND t.source = 'system.buy'
   AND t.type   = 'debit'      -- the platform-side half survives the delete
  GROUP BY 1, 2
),
forfeited AS (
  SELECT a.org_id,
         a.target->>'ID'                    AS account_id,
         (a.metadata->>'amount')::bigint    AS balance_at_delete,
         (a.metadata->>'purchased')::bigint AS purchased_at_delete
  FROM auditlogs a
  WHERE a.action = 'app.billing.tokens.forfeited'
)
SELECT coalesce(b.org_id, f.org_id)         AS org_id,
       coalesce(b.account_id, f.account_id) AS account_id,
       b.tokens_bought,        -- ceiling, rebuilt from the ledger remnants
       f.balance_at_delete,    -- from the forfeit audit record
       f.purchased_at_delete   -- the refundable number
FROM bought b
FULL OUTER JOIN forfeited f
  ON b.org_id = f.org_id AND b.account_id = f.account_id;
```

`tokens_bought` is only a ceiling: usage rows carry no org attribution on their surviving halves, so the unspent share is not derivable from the ledger remnants — that is exactly what the forfeit record exists for. The `stripe_session` ids bridge to Stripe for the payment and refund history, which is the double-refund guard.


whoAbhishekSah:
Tested this live on a local build of the branch, with a real SMTP sink catching the mails and a real Stripe test checkout for the purchased-token case.

| Case | Result |
|------|--------|
| Zero-token org, owner deletes it | Mail sent with subject "Your organization was deleted". No settlement line, no amounts. |
| Complimentary (awarded) tokens, 100 forfeited | Mail has no amounts and no settlement line. The `app.billing.tokens.forfeited` audit row keeps amount=100, purchased=0. |
| Purchased tokens (paid a Stripe test checkout, then deleted) | Mail adds "Unused purchased tokens will be settled by the support team", still no numbers. Audit keeps amount=100, purchased=100. |
| Two owners | Both owners received their own copy. |
| Org deleted by a platform admin (service account) | The owner was mailed. The body names the deleter by its title. This org also had no billing account, so the empty-customers path works. |
| Mail server down | Delete still succeeds. The log warns "failed to send the delete notice ... connection refused". |
| Custom template via `billing.org_delete_notice` | Custom subject and body rendered. `.Amount` and `.Purchased` are available to config templates as described. |
| Blocked delete (active paid subscription) | Delete fails with the blocker error, the org survives, and no notice is attempted. After cancelling the subscription, the retried delete succeeded and mailed the owner. |

The default mails never contain token counts, which matches the settlement concern in the description.

core/deleter/service.go
```diff
@@ -281,6 +281,10 @@ func (d Service) DeleteOrganization(ctx context.Context, id string) error {
 		return err
 	}
 
+	// the delete is going ahead: find who to notify while the owner
+	// policies still exist. Best-effort — the delete never depends on it
+	notice.Owners = d.resolveOwners(ctx, id)
```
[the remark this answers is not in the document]
whoAbhishekSah: Fixed with durable state, one nuance on what gets persisted: the owner ids, not the recipients themselves. The users outlive the org (only their owner policies are deleted), so before teardown starts the resolved owner ids go into an app.organization.delete.recipients audit record, and a retry that can no longer resolve owners loads the ids from the newest record and re-fetches the users. Best-effort on both sides like the rest of the notice machinery: with a non-db audit sink there is nothing to read back and the send path logs that no owner could be notified.

rohilsurana (approved):

rohilsurana (commented):
A few notes on the audit record, the tests, and some leftovers from the rename.

core/audit/audit.go
```diff
@@ -98,6 +98,7 @@ const (
 	BillingAccountDetailsUpdatedEvent EventName = "app.billing.account.details.updated"
 	BillingCheckoutDeletedEvent       EventName = "app.billing.checkout.deleted"
 	BillingTokensForfeitedEvent       EventName = "app.billing.tokens.forfeited"
+	OrgDeleteNoticeRecipientsEvent    EventName = "app.organization.delete.recipients"
```
rohilsurana: This event needs a doc entry. `audit.Service.Create` publishes every log through `webhookService.Publish`, so subscribers will start receiving `app.organization.delete.recipients` with the org owner user ids in `metadata.owner_ids`. It is missing from `docs/content/docs/reference/webhook.mdx` and from `web/sdk/admin/utils/webhook-events.ts`, so a subscriber has no way to know it exists or what shape it has. `app.billing.tokens.forfeited` from #1880 has the same gap, so the drift is growing.

Separate question: do we want this on the webhook feed at all? The record exists so a retry can look the recipients up again. It is internal state. The webhook fan-out is a side effect of using `audit.Service` to store it, not something a subscriber asked for, and it pushes owner user ids out to every registered endpoint. The `ignoreList` option only skips `s.publisher`, not the webhook, so there is no way to opt out today.

core/deleter/delete_notice.go
```diff
@@ -126,20 +127,68 @@ func (d Service) resolveOwners(ctx context.Context, orgID string) []user.User {
...
 	}
 	return owners
 }
 
+// recordNoticeRecipients writes the owner ids to an audit record before
+// teardown starts. Teardown deletes the owner policies before the org row,
+// so a retry that begins after that point cannot resolve the owners any
+// more; the record is what it recovers them from. The users themselves
+// outlive the org, so ids are enough. Best-effort, like the notice itself.
+func (d Service) recordNoticeRecipients(ctx context.Context, orgID string, owners []user.User) {
```
rohilsurana: The record and recover path has no test. `core/deleter/service_test.go` never calls `audit.SetContextWithService`, so every test runs against the fallback service that `audit.GetService` builds. That one uses `NoopRepository`, where `Create` returns nil and `List` returns an empty slice. So `recordNoticeRecipients` and `recoverRecipientsFromAudit` only ever run in their no-op form.

That means a change that stops writing `owner_ids`, writes the wrong metadata key, or breaks the lookup would not fail a single test. `recoverForfeitFromAudit` already has the same gap.

A test that puts a real audit service with a fake repository in the context would cover it: assert the record is written on the first attempt, then assert the recipients come back on a retry where the owner lookup fails.

core/deleter/service_test.go
```diff
@@ -435,6 +457,10 @@ func TestDeleteOrganization(t *testing.T) {
 			Return([]role.Role{}, nil)
 		m.orgSvc.EXPECT().DeleteModel(mock.Anything, "org-1").Return(nil)
 
+		// owner lookup is best-effort; failing it must not affect the delete
+		m.roleSvc.EXPECT().Get(mock.Anything, schema.RoleOrganizationOwner).
+			Return(role.Role{}, errors.New("no owners in this test"))
```
rohilsurana: This stub makes the test stop asserting the thing the PR adds. `resolveOwners` bails, `recoverRecipientsFromAudit` finds nothing, and `sendDeleteNotices` returns at the `len(notice.Owners) == 0` guard. The notice is silenced here, not asserted.

Same thing in "subscription gone on the provider does not brick the delete" at line 565. Both of these end in `assert.NoError`, so they are successful deletes. The PR description says every successful delete now asserts the notice path, but these two do not. If the notice broke for orgs with a cancelled subscription, both tests would still pass.

The failing stub makes sense in the tests that assert an error. For these two, stubbing the owner lookup to succeed and asserting the mail would cover the new behavior.

core/deleter/delete_notice.go
```diff
@@ -62,62 +63,62 @@ type forfeitNotice struct {
...
+				return deleteNotice{}, err
 			}
 			total += balance
 			purchased += bought
 			accounts[c.ID] = accountTokens{Balance: balance, Purchased: bought}
 		}
 	}
 	if total == 0 {
-		return forfeitNotice{Accounts: accounts, Balances: balances}, nil
+		return deleteNotice{Accounts: accounts, Balances: balances}, nil
```
rohilsurana: This branch is dead now. It only existed to skip the `Owners: d.resolveOwners(...)` field on the tail return, and this PR removed that field. `purchased` can only grow when `balance > 0`, which also grows `total`, so `total == 0` means `purchased == 0`. Both returns produce the same value.

Two exits with one behavior makes a reader prove the equivalence before trusting that nothing is missing. Dropping the branch is clearer.

core/deleter/delete_notice.go
```diff
@@ -62,62 +63,62 @@ type forfeitNotice struct {
...
+// collectDeleteNotice reads the token balances the delete is about to
+// forfeit. It only reads; a failure here aborts the delete before anything
+// is torn down.
 //
 // The amount is the whole remaining balance. Purchased is the share of it
 // that came from purchases (source system.buy), with complimentary tokens
 // (plan starter grants and awards) counted as spent first. Only the
-// purchased share is transferable.
-func (d Service) collectForfeitNotice(ctx context.Context, org organization.Organization, customers []customer.Customer) (forfeitNotice, error) {
+// purchased share is settled with the customer.
```
rohilsurana, line 72: This line swaps "transferable" for "settled with the customer", which matches the new mail copy. Two comments nearby still use the old word. The `Purchased` field comment on line 39 says "only this part is transferable", and the refund comment on line 262 says "it must not count as transferable". Worth making them agree.

Also `core/deleter/service_test.go:363` still says "the positive balance makes the delete collect the owners up front". That was the old trigger. Owners are collected on every delete now, whatever the balance.

core/audit/audit.go
whoAbhishekSah: Taken further than the docs fix: the recipients record is gone entirely (reverted). You are right that it was internal state riding a channel every subscriber receives — owner ids on the webhook feed for a failure window spanning two local statements at the tail of teardown was a bad trade. A delete failing exactly there and completing on retry now logs that no owner could be notified, which is the visible outcome. The docs drift is closed for the events that remain: app.billing.tokens.forfeited and app.billing.checkout.deleted are now in webhook.mdx and webhook-events.ts. The wider question of an opt-out from webhook fan-out for audit writes is real but no longer needed by this PR.

core/deleter/service.go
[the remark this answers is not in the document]
whoAbhishekSah: Update: the durable-recipients fix was reverted after review discussion. Persisting the recipient set meant writing internal retry state as an audit event, and every audit write fans out to the registered webhooks — owner user ids pushed to third-party endpoints, with no opt-out mechanism today. The window it protected spans two local statements at the very end of teardown (owner policies deleted, then roles and the org row), after all the failure-prone provider work is done. A delete failing exactly there and completing on retry now logs a warning that no owner could be notified, which keeps the gap visible without the exposure. The forfeit amounts themselves are unaffected — they live in the audit records either way.

rohilsurana (approved):

PR #1905, opened 2026-08-27
docs: RFC 0002 for explicit consent at signup

rohanchkrabrty:
## Summary

- Adds RFC 0002, proposing explicit consent at signup: a deployment declares its documents in `app.consent` config, the client sends the accepted document ids, and frontier writes an immutable `user_consents` record together with the new user in one transaction. With `app.consent` unset or disabled, nothing changes.
- Adds a `flow_intent` field on `AuthenticateRequest` to separate login from signup. Consent is a signup-only requirement, and today both are the same request, so a login silently creates the account it should have refused and the consent check has nothing to key off before the OIDC redirect.
- Specifies storage: one row per consent with document versions and URLs copied from config, no foreign key to `users` so records outlive account deletion, and `BEFORE UPDATE`/`BEFORE DELETE` triggers plus a create-only repository making them unmodifiable.
- Specifies the `user.consent_granted` audit record fields, including why its actor has to be set explicitly rather than left to context enrichment on the skip-listed auth endpoints, and why it cannot be atomic with the consent write.
- Records the accepted tradeoffs: no re-consent or withdrawal, three operator paths exempt from consent, and the account-enumeration oracle the login gate introduces.


rohilsurana:
Few points as we discussed -
- remove metadata from consent table. If needed we will ad that in future.
- remove GIN index from documents field of consent table. We don't really need to access it in that way.
- consent record should store auth_strategy instead of auth_method.
- audit record for this event will have `schema.PlatformOrgId` as the `org_id`
- add an endpoint to fetch the list of documents on UI

AmanGIT07 (approved):

rohilsurana (approved):

PR #1914, opened 2026-08-31
feat(consent): record consent in the same transaction as the user

rohanchkrabrty:
## Summary

- Closes the invariant the whole feature rests on: **a user row without a consent record is impossible.** `ResolveAll` runs before the transaction opens, so an incomplete payload never starts one; inside it, the user insert and the consent insert both land or neither does. Per [RFC 0002](https://github.com/raystack/frontier/blob/main/docs/rfcs/0002-explicit-consent-at-signup.md), *Enforcement* and *Storage*.
- `user.Repository` gains `CreateWithTx`, which is the one place this feature reaches outside its own domain. `pkg/db` has `WithTxn` but carries no transaction on the context, so it is threaded through explicitly. Both create paths share one query builder, so they cannot drift.
- **An existing user is written nothing, whatever the flow carries.** A record written outside a user creation would stamp this moment's timestamp and IP on an agreement made elsewhere, which is worse than no record because it reads like evidence. The IP and time come from when the user accepted, carried on the flow across the redirect.
- A `user.consent_granted` audit record is written after the commit, through the repository rather than the service, with `Actor` filled in explicitly — these endpoints are skip-listed, so an empty actor would be enriched to the nil UUID and the `system` actor for an act a person performed. It cannot be atomic with the consent row, which is why that row is the source of truth and a failed audit write is logged and carried past.
- The rollback is proved against a real database, not a mocked transaction — a mock can only pretend to roll back.


core/consent/service.go
```diff
@@ -94,6 +121,97 @@ func (s Service) ResolveAll(ids []string) ([]Document, error) {
...
+		},
+		// the platform org, not a blank one, so readers need no special case
+		OrgID: schema.PlatformOrgID.String(),
+		// when the user accepted, not when this row was written.
+		OccurredAt: granted.ConsentedAt,
+		// IdempotencyKey is left empty: it is nullable, and a consent record is
+		// written once, so there is nothing to deduplicate.
+	}); err != nil && s.logger != nil {
+		s.logger.ErrorContext(ctx, "failed to write the audit record for a granted consent",
+			"consent_id", granted.ID, "user_id", granted.UserID, "error", err)
```
AmanGIT07, line 197: In case of error we can manually create the record for book keeping. In that case we will need all the fields and info. Check if we need to print more info here or the rest is already there in `consents` table?

core/consent/service.go
```diff
@@ -1,20 +1,47 @@
...
 )
 
-// Service owns the document config, so it owns the checks that read it. Config
-// is read at boot, so a version change needs a restart.
+// Repository writes consent records and nothing else: the table is immutable.
+// It takes the transaction rather than opening one, because the record has to
+// land with the user row or not at all.
+type Repository interface {
+	Create(ctx context.Context, tx *sqlx.Tx, cnst Consent) (Consent, error)
+}
```
AmanGIT07, line 22: We will need to establish a pattern across the codebase on how to move transactions across services — this doesn't exist as of now, and will be done later. The approach may vary as well. So instead of doing it as a one-off case and making `db trnx` as dependency in service, can we instead look at alternatives, e.g. a postgres-layer method (say `CreateWithConsent` with the interface declared on authenticate's side since it already imports both domains) which opens WithTxn internally and inserts all the rows required?

AmanGIT07: @rohilsurana thoughts?

core/consent/service.go
rohanchkrabrty: All the info needed is present in `user_consents` table. Only `org_id` is missing but that will always be `PlatformOrgID`

core/consent/service.go
```diff
@@ -94,6 +110,84 @@ func (s Service) ResolveAll(ids []string) ([]Document, error) {
...
+	if req.Source == "" {
+		req.Source = SourceSignup
+	}
+	if err := validateGrant(req); err != nil {
+		return Consent{}, err
+	}
+
+	return Consent{
+		UserID:       req.UserID,
+		UserEmail:    req.UserEmail,
```
AmanGIT07, line 114: PrepareGrant copies UserID/UserEmail from the request, but the composite write overwrites both from the inserted row, so no caller can ever set them. Can we drop the two fields from GrantRequest until a caller exists that needs them?

core/user/user.go
```diff
@@ -25,6 +26,10 @@ type Repository interface {
 	GetByIDs(ctx context.Context, userIds []string) ([]User, error)
 	GetByName(ctx context.Context, name string) (User, error)
 	Create(ctx context.Context, user User) (User, error)
+	// CreateWithConsent writes the user row and the signup consent record in one
+	// transaction, so neither can exist without the other. Temporary: it lives here
+	// until there is a pattern for a transaction spanning two domains.
```
AmanGIT07: We can cleanup this comment

core/consent/service.go
```diff
@@ -1,20 +1,36 @@
...
 )
 
+type AuditRecordRepository interface {
+	Create(ctx context.Context, auditRecord models.AuditRecord) (models.AuditRecord, error)
+}
+
 // Service owns the document config, so it owns the checks that read it. Config
-// is read at boot, so a version change needs a restart.
+// is read at boot, so a version change needs a restart. It writes no record
+// itself: that lands in the same transaction as the user row, in user.Service.
```
AmanGIT07: We can cleanup this comment

internal/store/postgres/user_repository.go
```diff
@@ -126,14 +124,79 @@ func (r UserRepository) Create(ctx context.Context, usr user.User) (user.User, e
...
+
+	if err := r.dbc.WithTxn(ctx, sql.TxOptions{}, func(tx *sqlx.Tx) error {
+		var txErr error
+		if createdUser, txErr = r.createWithTx(ctx, tx, usr); txErr != nil {
+			return txErr
+		}
+		// the id only exists once the row is written
+		cnst.UserID = createdUser.ID
+		cnst.UserEmail = createdUser.Email
+		createdConsent, txErr = NewUserConsentRepository(r.dbc).Create(ctx, tx, cnst)
```
AmanGIT07: We construct a new UserConsentRepository on every signup, inside the transaction. Can we instead hold it as a field on UserRepository, built once in NewUserRepository? It has no other production caller now, so nothing else changes.

core/authenticate/service_test.go
```diff
@@ -1498,3 +1500,244 @@ func TestService_FinishFlow_Intent(t *testing.T) {
 		})
 	}
 }
+
+// TestService_FinishFlow_Consent covers which of the three outcomes each request
+// reaches, and what is written for it. The rollback behind the invariant is
+// exercised against a real database in internal/store/postgres.
+func TestService_FinishFlow_Consent(t *testing.T) {
```
AmanGIT07, line 1507: The suite covers the ResolveAll gate but not a PrepareGrant failure. A flow can carry complete ids with a missing or unparseable `at`, which prepares a zero ConsentedAt and should fail the signup before anything is written. Can we add that case, asserting CreateWithConsent is never called?

AmanGIT07:
@rohanchkrabrty Please use `frontier-sandbox` skill to verify existing RPCs are not affected. (either now or any other PR on this stack as appropriate)

core/consent/service.go
rohanchkrabrty: The overwrite is a CreateWithConsent thing. Future callers, like re-consent or other cases, won't go through it, so the override is a one-off. The id and email are part of a consent grant, so I don' think we should drop those fields

AmanGIT07 (approved):

PR #1915, opened 2026-08-31
feat(authenticate): map the gate and consent errors on both auth RPCs

rohanchkrabrty:
## Summary

- The previous three PRs built machinery that can refuse a request; until this one, every refusal reached the client as a 500. `ErrLoginUserNotFound` → `NotFound`, `ErrSignupUserExists` → `AlreadyExists`, `ErrConsentRequired` → `FailedPrecondition`, from **both** `Authenticate` and `AuthCallback`. All three had to join the fixed list `AuthCallback` handles explicitly, or they fall through to `Internal` by construction.
- `FailedPrecondition` specifically, because the other errors on that list are all `InvalidArgument` — the client sent something wrong and resending will not help. A consent rejection is the opposite: the request was well formed and the client can fix it by asking the user to accept what is missing. That distinction is what a client cannot recover from a message string.
- **Settles the surface [RFC 0002](https://github.com/raystack/frontier/blob/main/docs/rfcs/0002-explicit-consent-at-signup.md) left open, as "a code, on both RPCs".** The RFC weighed a redirect against rendering in place; neither applies. Frontier serves no route for the callback URL — it points at a page the application hosts, and that page calls `AuthCallback` over connect, so it already holds the rejection and decides where the user goes. A redirect was implemented first and withdrawn: nothing follows the location header, and a 2xx carrying it reads as success to a client that is not looking for it.
- `StartFlow` gains the consent half of the flow-start gate, so a signup rejection lands before an OTP is sent and before the browser leaves for the provider. An unspecified intent checks only that the ids are known, since completeness is not yet knowable; a login intent checks nothing, because it writes no record.
- Ids sent with a login intent are rejected as a request-shape error, unconditionally. With `app.consent` disabled ids are ignored rather than rejected, which is content and handled one layer down. Rendering any of this is a separate frontend change.

### DIFF
--- a/internal/api/v1beta1connect/consent.go
+++ b/internal/api/v1beta1connect/consent.go
@@ -22,3 +22,10 @@ func (h *ConnectHandler) ListConsentDocuments(ctx context.Context, request *conn
 	}
 	return connect.NewResponse(&frontierv1beta1.ListConsentDocumentsResponse{Documents: pbdocuments}), nil
 }
+
+// consentEnabled reports whether this deployment asks for consent at all. Boot
+// validation rejects an enabled block with no documents, so an empty set means
+// the feature is off and there is nothing a document id could refer to.
+func (h *ConnectHandler) consentEnabled() bool {
+	return h.consentService != nil && len(h.consentService.Documents()) > 0
+}
--- a/internal/api/v1beta1connect/consent.go
+++ b/internal/api/v1beta1connect/consent.go
@@ -22,10 +22,3 @@ func (h *ConnectHandler) ListConsentDocuments(ctx context.Context, request *conn
 	}
 	return connect.NewResponse(&frontierv1beta1.ListConsentDocumentsResponse{Documents: pbdocuments}), nil
 }
-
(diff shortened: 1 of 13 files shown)


internal/api/v1beta1connect/consent.go
```diff
@@ -22,3 +22,10 @@ func (h *ConnectHandler) ListConsentDocuments(ctx context.Context, request *conn
 	}
 	return connect.NewResponse(&frontierv1beta1.ListConsentDocumentsResponse{Documents: pbdocuments}), nil
 }
+
+// consentEnabled reports whether this deployment asks for consent at all. Boot
+// validation rejects an enabled block with no documents, so an empty set means
+// the feature is off.
+func (h *ConnectHandler) consentEnabled() bool {
```
AmanGIT07: `consentEnabled` derives "consent is on" from the length of `Documents()`, which leans on boot validation to stay true. Can we instead expose `Enabled()` on the consent service and have the handler ask, so the rule lives with the config it reads?

AmanGIT07 (approved):

PR #1922, opened 2026-09-08
feat(sdk): take explicit consent and send a login or signup intent

rohanchkrabrty:
## Summary

- Implements the client section of `docs/rfcs/0002-explicit-consent-at-signup.md`: `SignInView` sends a login intent, `SignUpView` sends a signup intent, and the shared `MagicLinkView` takes the intent as a prop.
- `SignUpView` fetches `ListConsentDocuments` and gates every sign-up control behind an Apsara checkbox whose label links each document; an empty list renders today's view unchanged. A `consentLabel` prop overrides the copy.
- Adds `describeAuthError`, which maps the `NotFound`, `AlreadyExists` and `FailedPrecondition` rejections to SDK-owned copy and returns it with the strategy the rejection came through. Both views take that pair as `error` and render it under the strategy button that caused it, on the email field, or as a message for the whole group; the OTP view uses the same copy. Both views also stop swallowing `authenticate` failures into `console.error`.
- On the Go side, the three flow rejections now carry the strategy the flow came through, and `AuthCallback` attaches it to the connect error as a `raystack.frontier.v1beta1.AuthStrategy` error detail, which the SDK reads with `findDetails`. An OIDC callback names no strategy and the flow row is gone by then, so this is the only place the client can learn it. No proto change.
- The client demo's callback page routes a rejection by its connect code and hands the message and strategy to the login or signup page in memory, so nothing about the error reaches the URL or survives a reload. `@raystack/proton` is bumped to the build that carries the consent proto.

https://claude.ai/code/session_01SwGUtD3mDrTjM1rLWnxrnA


web/sdk/client/views/auth/sign-up/sign-up-view.tsx
```diff
@@ -1,80 +1,212 @@
...
+  const acceptedDocumentIds = useMemo(
+    () => documents.map(document => document.id),
+    [documents]
+  );
+
   const { mutateAsync: authenticate } = useMutation(
     FrontierServiceQueries.authenticate
   );
 
+  const blocked = consentPending || (documents.length > 0 && !consented);
```
rohilsurana: `blocked` only looks at `consentPending`, so both unhappy paths of this query go wrong, and neither one shows the user a message.

If the request stalls (offline, or a slow network), react-query keeps `status: 'pending'`. `consentPending` stays true, so every OIDC button and the email form stay disabled forever and nothing on the page says why.

If the request fails (a 5xx, or a server that does not have the RPC yet), `status` becomes `'error'`. `consentPending` flips to false, `consentData` is undefined, `documents` is `[]`, so no checkbox renders and `blocked` is false. Sign-up then goes out with `acceptedDocumentIds: []`. A deployment that does configure consent answers `ErrConsentRequired`, and this view renders "Please accept the required documents to continue." on a page that has no checkbox to accept. There is no way to retry either.

The destructure takes only `data` and `isPending`, so `isError` is never read. Reading it and rendering an error with a retry would cover both cases.

web/sdk/client/views/auth/sign-up/sign-up-view.tsx
```diff
@@ -1,80 +1,212 @@
...
 export const SignUpView = ({
   logo,
   title = 'Create your account',
   excludes = [],
+  consentLabel,
+  error,
   ...props
 }: SignUpViewProps) => {
   const { config } = useFrontier();
+  const [consented, setConsented] = useState(false);
```
rohilsurana: `consented` is a bare boolean, so it is not tied to the document set the user actually agreed to.

The query below passes no options, so it takes the react-query defaults: `staleTime: 0` and refetch on window focus. The list can change under an open page. A user ticks the box for [terms_of_service, privacy_policy]. An admin then adds a third document, or republishes one under a new id. The user tabs away and back, the query refetches, and `documents` plus `acceptedDocumentIds` become the new set. But `consented` is still true, so `blocked` stays false. The next click sends the new ids as accepted, and the server writes a consent record for a document the user never saw. That is the one thing this change is meant to prevent.

Keying the state to the set fixes it. For example, store the accepted ids instead of a boolean and treat consent as given only when they match the current `documents`. Passing `staleTime: Infinity` on the query would also stop the list moving on its own.

rohanchkrabrty: Resolved

web/sdk/client/views/auth/sign-up/sign-up-view.tsx
rohanchkrabrty: We don't have a pattern of handling request stalls at this moment. We can think and take up this later. For example currently we don't show the auth strategy buttons if `ListAuthStrategies` fail. So for now let's keep the current behaviour.

Resolved the other points

rohilsurana (approved):

PR #1926, opened 2026-09-10
fix(sdk): drop the invoice count from the delete blocker message

whoAbhishekSah:
When more than one invoice is unpaid, the delete blocker message now reads:

`Please pay your outstanding invoices before deleting this workspace.`

It used to include the number, as in "your 3 outstanding invoices". The single-invoice sentence is unchanged.


whoAbhishekSah:
Closing. The screenshot that prompted this was taken from a tab still running the previous bundle. The current wording is fine as it is.

PR #1927, opened 2026-09-16
feat(store): add deleted_at to domains, billing checkouts, and billing transactions

whoAbhishekSah:
## What

Adds a nullable `deleted_at timestamptz` column to `domains`, `billing_checkouts`, and `billing_transactions`. Every other table already has one. Adds the matching `DeletedAt` field to the `Domain`, `Checkout`, and `Transaction` structs.

## Why

Soft delete marks a row with a timestamp instead of removing it. These three tables had nowhere to put that timestamp. This change only adds the column. Later changes will start writing to it.

The struct fields are needed in the same change. The three repositories read with `SELECT *` and scan into a struct. sqlx is strict, so a column with no matching field fails every read with `missing destination name deleted_at`.

## Behaviour change

None. Nothing writes `deleted_at` on these tables yet.

## Deploy note

This must reach an environment before any release that writes `deleted_at` on these tables.

## Tested

- Ran the full migration set on an empty local Postgres database, then rolled back one step and re-applied. Columns appear and disappear as expected.
- Read rows through `DomainRepository`, `BillingCheckoutRepository`, and `BillingTransactionRepository` against that database. All reads pass.
- Confirmed that the old struct shape, without `DeletedAt`, fails the same read with `missing destination name deleted_at`.
- `go test ./internal/store/postgres/` passes.


whoAbhishekSah:
Tested this live against a local server built from this branch, with migration `20260916100000` applied on a database that already had checkout and transaction rows.

**Migration**
- `server migrate` added a nullable `deleted_at timestamptz` to `domains`, `billing_checkouts`, and `billing_transactions`. Version reported as `20260916100000`, not dirty.

**Domains** (fresh org, normal user session)
- Created two domains, listed them, fetched one by id, and ran verify on it. All worked.
- Set `deleted_at` directly in the database on one domain. List and Get still return it with the same fields as before.
- Deleted the other domain through the API. It was removed and the list shrank to one.

**Billing transactions** (fresh org)
- Granted credits with `DelegatedCheckout` on a credits product. The credit transaction was listed and the balance read 100.
- Set `deleted_at` on that credit row. `ListBillingTransactions` and `GetBillingBalance` return the same values. Nothing filters on the column yet, which matches the PR description.
- Recorded a usage of 10 with a deleted row present. The debit row was written, listed, and `TotalDebitedTransactions` shows 10.

**Billing checkouts**
- Created a Stripe checkout on the fresh org and listed and fetched it.
- Set `deleted_at` on it and on one of the pre-existing checkouts from another org. `ListCheckouts` and `GetCheckout` return both rows with no change.

No read or write on these three tables failed. Rows with `deleted_at` set and rows with it null both come back the same way.

whoAbhishekSah:
Response shape check: a row with `deleted_at` set and a row with it null return the same JSON for domains, billing transactions, and billing checkouts.

- `deleted_at` is not in the proto, so it never appears in any response.
- No field is added or dropped between the two cases.
- No read returned an error.

rohanchkrabrty (approved):

PR #1929, opened 2026-09-17
fix(store): return every org service user and honour the requested sort

rohilsurana:
`SearchOrganizationServiceUsers` returned fewer service users than `ListOrganizationServiceUsers` for the same org. Digging in turned up five separate bugs. All three are confirmed on a local Frontier, before and after.

## Bug 1: service users without a project policy were dropped

`ListOrganizationServiceUsers` reads the `serviceusers` table directly, so it returns every service user in the org. The search query built its rows with two inner joins:

```sql
FROM serviceusers
INNER JOIN policies ON serviceusers.id = policies.principal_id
                   AND policies.principal_type = 'app/serviceuser'
                   AND policies.resource_type  = 'app/project'
INNER JOIN projects ON policies.resource_id = projects.id
```

A service user only survived those joins if it had at least one policy on a project. Service users with only an org level policy, or with no policy at all, were dropped before any filter or limit ran. The projects join also never checked `projects.deleted_at`, so a deleted project still showed up in the response.

Both joins are now `LEFT JOIN`, the projects join requires `projects.deleted_at IS NULL`, and the aggregate is wrapped in `COALESCE(... FILTER (WHERE projects.id IS NOT NULL), '[]')`. The filter drops the null row a left join produces, and the coalesce returns `[]` rather than SQL `NULL`, which the row struct scans as a plain string.

## Bug 2: the requested sort was ignored

The base query ended with a fixed `ORDER BY serviceusers.title ASC`, and the requested sort was appended after it. So the fixed term always won and the requested sort only broke ties between rows with the same title. Titles are mostly unique, so it did nothing.

The old expected SQL showed it plainly. Asking for title descending produced:

```sql
ORDER BY "serviceusers"."title" ASC, "serviceusers"."title" DESC
```

which is just title ascending.

The base query no longer sets an order. The requested sort leads, `title ASC` is used only when no sort is asked for, and a final `id ASC` term makes the order total so paging by offset cannot repeat or skip a row.

## Bug 3: filtering on created_at always failed

`AggregatedServiceUser` tagged its fields like `rql:"filter,sort,column=created_at"`. Salt resolves an rql field by the `name` tag (or the lowercased Go field name) and reads its type from the `type` tag. It ignores `column`. So `created_at`, `org_id` and `updated_at` did not resolve at all, and every field that did resolve fell back to type `string`.

The result was that any filter on `created_at` failed in the handler before reaching the database:

```
invalid_argument: failed to read rql query: 'created_at' is not a valid field
```

Tags are now `rql:"name=created_at,type=datetime"` and so on, matching `core/prospect/prospect.go`.

## Refactor: use the shared rql helpers

The repository had its own `addFilter`, `addSearch` and `addSort`. These are gone, replaced by `utils.AddRQLFiltersInQuery`, `AddRQLSearchInQuery`, `AddRQLSortInQuery` and `AddRQLPaginationInQuery` from `pkg/utils/rql.go`, as used by `prospect_repository.go`, `userpat_repository.go` and `audit_record_repository.go`.

One wrinkle. Those helpers emit unqualified column names, and the three joined tables here all have an `id`, a `title` and a `created_at`. An unqualified name in `ORDER BY` resolves to the select alias and is fine, but in `WHERE` Postgres rejects it:

```
ERROR:  column reference "title" is ambiguous
```

So the joined and grouped query is now wrapped in a subquery, and rql is applied to the wrapper where only the aliased output columns are visible. Explicit allowlists say which columns may be filtered, searched and sorted, and unsupported ones are rejected as `ErrBadInput`, which the handler maps to `invalid_argument` instead of `internal`.

## Bug 4: a project could be listed twice

If a service user holds two roles on the same project, it has two policy rows, so the join produces two rows and the project appeared twice in the response. The admin UI renders the titles joined by a comma, so it showed "Live Project, Live Project". Confirmed on a local database.

The aggregate is now `JSONB_AGG(DISTINCT JSONB_BUILD_OBJECT(...))`. It has to be `jsonb` rather than `json` because the `json` type has no equality operator, so `DISTINCT` cannot be applied to it.

## Bug 5: the UI offers a Projects filter the API rejected

`web/sdk/admin/views/organizations/details/apis/columns.tsx` marks the Projects column `enableColumnFilter: true`. That filter never worked. The old repository silently ignored any filter it did not recognise, so the results came back unfiltered. Filtering for `%Live%` and for `%zzzz%` both returned the same row.

The base query now also aggregates the project titles into a `projects` text column, using the same comma separated form the UI renders. That column is allowlisted for filters and for search, so the filter does what it says. Filtering for `%zzzz%` now correctly returns nothing.

## Checked against the admin UI

`web/` in this repo holds the admin UI, so the request shapes are not guesswork.

- **The `like` change is a non-issue.** I flagged earlier that the shared helper stops auto-wrapping `like` values in `%`. Apsara's DataTable never sends `like`. It maps `contains`, `starts_with` and `ends_with` to operator `ilike` and adds the wildcards itself in `filter-operations.cjs`. `ilike` took the same default code path before and after, so nothing changes. Verified: the UI's exact request, `ilike` with `%policy%`, returns the right rows.
- **The default limit change is a non-issue.** `transformDataTableQueryToRQLRequest` sends `limit: query.limit || defaultLimit` with `defaultLimit = 50`, so the UI always sends a limit.
- **Pagination got more correct.** The table uses `useInfiniteQuery`, and `getConnectNextPageParam` computes `hasMorePages` as `items.length === pagination.limit` and the next offset as `pagination.offset + pagination.limit`, both read from the response. The response now reports the limit actually applied rather than echoing the request, and the new `id ASC` tie breaker keeps pages from repeating or skipping rows.
- **The date filter was completely broken.** The Projects view sets `filterType: "date"` on Created on and maps `createdAt` to `created_at`. Every such request failed with `'created_at' is not a valid field` because of bug 3. It works now.
- **The default sort is the one that was ignored.** That view's `DEFAULT_SORT` is `{ name: 'createdAt', order: 'desc' }`, which is exactly the request in the original report.

## Verified end to end on a local Frontier

Postgres and SpiceDB in docker, server built from this branch, seeded with four service users in one org: one with a project policy, one with only an org policy, one with no policy, and one whose policy points at a deleted project. Same request, two binaries.

Before, built from `main`:

```
2 returned:
   a-with-project     | projects: ['proj-live']
   d-deleted-project  | projects: ['proj-gone']     <- deleted project leaking
```

After, built from this branch, sorted `created_at desc`:

```
4 returned:
   d-deleted-project  | projects: []
   b-org-policy-only  | projects: []
   c-no-policy        | projects: []
   a-with-project     | projects: ['proj-live']
```

Also checked against the running server, with the UI's own request shapes:

| case | before | after |
|---|---|---|
| sort `title desc` | came back ascending | `d, c, b, a` |
| sort `created_at desc` | ignored | `d, b, c, a`, newest first |
| filter `created_at` range, as the date picker sends it | `invalid_argument` | correct subset |
| filter `title` `ilike` `%policy%`, as the UI sends it | returned nothing | `b, c` |
| filter `projects` `ilike` `%Live%` | filter ignored, wrong rows | `a-with-project` |
| filter `projects` `ilike` `%zzzz%` | returned a row anyway | empty, correctly |
| two roles on one project | project listed twice | listed once |
| limit 2, offset 0 then 2 | n/a | `a, b` then `c, d`, no overlap |
| sort on `updated_at` | `internal` | `invalid_argument` |
| no limit given | unbounded | 50 |

## Two regressions this PR introduced, now fixed

An independent review caught two cases where this branch turned a working or harmless request into a 500. Both reproduced against `main` and this branch on a local server.

**`group_by` returned a 500.** `AddRQLSortInQuery` prepends `ORDER BY <group_by column>` for every group_by entry, and the wrapper select has no such column. `main` returned 200 and ignored group_by. The repository now rejects group_by as `ErrBadInput`, which is what `user_orgs_repository.go` does, since this response has no group block to fill.

**Bad filter values returned a 500.** The handler never called `rql.ValidateQuery`, unlike ten other handlers including its closest sibling `SearchOrganizationServiceUserCredentials`. On `main` a bad `created_at` value happened to fail earlier with a 400, only because bug 3 meant the field never resolved. Once it resolved, the raw value reached Postgres. The handler now validates, and the `ErrBadInput` branch returns the real reason instead of a generic message.

| request | main | before the review | now |
|---|---|---|---|
| `group_by: ["foo"]` | 200, ignored | 500 | 400, `'foo' is not a valid sort key` |
| `group_by: ["title"]` | 200, ignored | 200, sort silently overridden | 400, `group_by is not supported` |
| `created_at gt "yesterday-ish"` | 400 | 500 | 400, `not a valid ISO datetime string` |
| `created_at` with operator `in` | 400 | 500 | 400, same |

## Tests

There are two layers.

`TestOrgServiceUserRepository_prepareDataQuery` pins the generated SQL. The joined part lives in one constant so each case shows only the WHERE, ORDER BY and pagination it exercises. That catches the sort and tag bugs, but on its own it would pass unchanged with the old inner joins, so it does not prove much about the main bug.

`org_serviceuser_repository_pg_test.go` is new and runs the real query against a real Postgres through the existing dockertest harness. It seeds the five shapes that actually broke: a service user with a live project policy, one with only an org policy, one with no policy, one whose project is soft deleted, and one holding two roles on a single project. It asserts the row set, the per-user project lists, both sorts, the `created_at` filter, the `projects` filter, search across both columns, disjoint pages, and the rejected inputs.

I checked that it fails on the bug rather than merely passing on the fix. Putting the two `INNER JOIN`s back fails 5 of its 8 tests.

`golangci-lint run` reports 0 issues. The dockertest suites in this package need a reachable docker socket. With `DOCKER_HOST=unix:///.../docker.sock` the whole of `internal/store/postgres`, `internal/api/v1beta1connect` and `core/...` passes with no failures.

## One CI change

The `unit` job failed with `panic: test timed out after 2m30s`, and the log names the package:

```
FAIL	github.com/raystack/frontier/internal/store/postgres	150.221s
```

That is `make test` running `-count 2 -timeout 150s`. The timeout is per package, and this package now has 25 dockertest suites, each starting its own postgres container, each run twice.

The budget was already almost gone before this PR. Measured locally with the CI flags, the package takes 117.2s without the new suite and 120.3s with it. The new suite costs 3.1s, about 3 percent. CI stopped at 150.2s, so it was sitting at roughly 147s already. This is consistent with the history on this branch, where the same timeout panic appeared at `61bb71ca` and `821c0a6d`, both before the new suite existed, while other pushes passed. It was flaky at the limit.

So `-timeout` goes from 150s to 600s. That gives the package real headroom rather than leaving it one suite away from failing again. The new suite itself is not flaky: `-race -count 3` passes.

The better fix is for these suites to share one postgres container instead of starting 25. That is a change to shared test infrastructure and does not belong in this PR.

## Follow-ups, not in this PR

- No index leads with `policies.principal_id`, so every search sequentially scans `policies`. Pre-existing, and the real scaling limit here. A migration adding `(principal_id, principal_type, resource_type)` would fix it.
- `org_projects_repository.go` has the same inner-join shape: a project with no human members disappears from `SearchOrganizationProjects`, and there is no `deleted_at` guard on projects or users.
- `SearchOrganizationServiceUsers` returns 200 with an empty list for an org that does not exist, where `ListOrganizationServiceUsers` returns 404.
- The plain list returns `state`; the search does not. That is the last shape difference between the two endpoints.
- The 25 dockertest suites in `internal/store/postgres` each start their own container. Sharing one would cut minutes off every CI run.


internal/store/postgres/org_serviceuser_repository.go
```diff
@@ -106,7 +107,20 @@ func (r OrgServiceUserRepository) prepareDataQuery(orgID string, rql *rql.Query)
 			if err != nil {
 				return "", nil, err
 			}
+			sorted = true
```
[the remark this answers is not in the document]
rohilsurana: Not applicable any more, on two counts.

The code this points at is gone. `addSort` and the `sorted` flag were removed in 4fd4cd06, which moved this repository onto the shared helpers in `pkg/utils/rql.go`. Sorting now goes through `utils.AddRQLSortInQuery`, whose `default` branch appends the column ascending, so it always adds a term.

I checked the case you describe against the current code:

```
order=""      ORDER BY "title" ASC, "id" ASC
order="asc"   ORDER BY "title" ASC, "id" ASC
order="desc"  ORDER BY "title" DESC, "id" ASC
order="bogus" ORDER BY "title" ASC, "id" ASC
```

`Sort: [{Name: "title", Order: ""}]` does sort by title, so the missing-sort outcome no longer happens.

It also no longer reaches the query builder. abe53d76 added `rql.ValidateQuery` to the handler, which rejects both cases before then:

```
order=""      'title' is not a valid sort key
order="bogus" 'title' is not a valid sort key
```

Those come back as `invalid_argument`. So an unsupported order is now a 400 rather than a silently wrong sort. No change needed.

internal/api/v1beta1connect/organization_serviceuser.go
```diff
@@ -19,10 +20,16 @@ func (h *ConnectHandler) SearchOrganizationServiceUsers(ctx context.Context, req
...
+	}
+
 	serviceUsersData, err := h.orgServiceUserService.Search(ctx, request.Msg.GetId(), rqlQuery)
 	if err != nil {
 		if errors.Is(err, postgres.ErrBadInput) {
-			return nil, connect.NewError(connect.CodeInvalidArgument, ErrBadRequest)
+			return nil, connect.NewError(connect.CodeInvalidArgument, err)
 		}
 		return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("SearchOrganizationServiceUsers.Search: org_id=%s: %w", request.Msg.GetId(), err))
 	}
```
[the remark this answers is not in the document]
rohilsurana: Agreed, that was a real gap. Fixed in d22f197f with `internal/api/v1beta1connect/organization_serviceuser_test.go`.

It covers both paths you asked for and a few more:

- unparsable `created_at` → `CodeInvalidArgument`
- an operator the field does not allow (`in` on a datetime) → `CodeInvalidArgument`
- a filter naming a field that is not on the aggregate → `CodeInvalidArgument`
- an unsortable field → `CodeInvalidArgument`
- the service returning a wrapped `postgres.ErrBadInput` → `CodeInvalidArgument`, and the test asserts the reason survives rather than being flattened into a generic message
- any other service error → `CodeInternal`
- the happy path, mapping rows to the response

The validation cases set no mock expectation and the mock is built with `mocks.NewOrgServiceUserService(t)`, so the test fails if the handler reaches the service at all. That is what pins "rejected before the query is built".

rohilsurana:
> Re: **Assert the returned pagination metadata** (`org_serviceuser_repository.go:100-110`), which landed outside the diff range so it has no inline thread.

Fair point, fixed in d22f197f.

The unit test now asserts the whole `utils.Page` rather than just the derived limit, so both offset and limit are pinned, including the explicit offset and limit cases:

```go
assert.Equal(t, tt.wantPage, gotPage)
```

The integration suite gained `TestPaginationMetadata`, which goes through `Search` and asserts what actually comes back:

- no limit given → reports `Limit: 50`, `Offset: 0`, and returns all 5 rows
- `Limit: 2, Offset: 2` → reports 2 and 2, and returns 2 rows
- `Limit: 2, Offset: 4` → reports 2 and 4, and returns the 1 remaining row

That last case matters for the admin UI, which decides whether to fetch another page by comparing the row count against the limit it reads off the response.

rohilsurana:
## Endpoint comparison, run against a live server

I ran both endpoints side by side on a local Frontier (postgres and SpiceDB in docker) and recorded what each actually returned. Every row below is a real call, not a reading of the code.

The **Search (main)** column is the same request against a binary built from `main`, so the table doubles as a before and after.

_Re-run and corrected on 18 Sep 2026, after the rebase. Branch at `0e47579`, main at `f840f22e`. Six cells changed, and three rows were added. They are listed at the bottom._

### Fixtures

One org, six service users, two projects. Everything except the state changes was created through the API so SpiceDB stays consistent.

Creating a service user always gives it a viewer policy on its org, so every row below has one unless it says otherwise. The policy on `c-no-policy` was then removed by hand, to get a service user with no policy at all.

| service user | project policies | created_at | state |
|---|---|---|---|
| `a-with-project` | owner on a live project | 2026-01-01 | enabled |
| `b-org-policy-only` | none, only the org policy | 2026-06-01 | enabled |
| `c-no-policy` | none, and no org policy either | 2026-03-01 | enabled |
| `d-deleted-project` | owner on a **soft deleted** project | 2026-09-01 | enabled |
| `e-two-roles` | owner **and** viewer on the *same* project | 2026-04-01 | enabled |
| `f-disabled` | none, only the org policy | 2026-05-01 | **disabled** |

### Which rows come back

| # | Case | List | Search (main) | Search (this PR) |
|---|---|---|---|---|
| 1 | All service users in the org | **6** `a,b,c,d,e,f` | **3** `a,d,e` | **6** `a,b,c,d,e,f` ✅ |
| 2 | Has no policy at all (`c`) | yes | **missing** | yes ✅ |
| 3 | Has only an org policy (`b`) | yes | **missing** | yes ✅ |
| 4 | Is disabled (`f`) | yes | **missing** | yes ✅ |
| 5 | Org with no service users | 200, 0 rows | 200, 0 rows | 200, 0 rows ✅ |

`main` dropped half the org. That is the original bug: the inner join required a policy on a project.

### Response shape

| # | Case | List | Search (main) | Search (this PR) |
|---|---|---|---|---|
| 6 | Fields per row | `id` `title` `org_id` `state` `metadata` `created_at` `updated_at` | `id` `title` `org_id` `projects` `created_at` | unchanged |
| 7 | Returns `state` | yes | no | no |
| 8 | Returns `metadata` / `updated_at` | yes | no | no |
| 9 | Returns `projects` | no | yes | yes |
| 10 | Soft deleted project listed for `d` | n/a | **`proj-gone` shown** | excluded ✅ |
| 11 | Two roles on one project (`e`) | n/a | **listed twice** | listed once ✅ |

Rows 7 to 9 are a deliberate difference. The two endpoints answer different questions, and the proto messages are unchanged by this PR.

### Query support (List takes only an org id, so it has no query surface)

| # | Case | List | Search (main) | Search (this PR) |
|---|---|---|---|---|
| 12 | Sort `created_at desc` | not supported | **ignored**, `a,d,e` | `d,b,f,e,c,a`, newest first ✅ |
| 13 | Sort `title desc` | not supported | **ignored**, ascending `a,d,e` | `f,e,d,c,b,a` ✅ |
| 14 | Filter `title ilike %policy%` | not supported | 0 rows, `b` and `c` were already dropped | `b,c` ✅ |
| 15 | Filter `created_at > 2026-04-15` | not supported | **400** `'created_at' is not a valid field` | `b,d,f` ✅ |
| 16 | Filter `projects ilike %Live%` | not supported | **ignored**, returns `a,d,e` | `a,e` ✅ |
| 17 | Search `Live`, a project title | not supported | 0 rows, it searches the title only | `a,e`, covers project titles ✅ |
| 18 | Pagination, limit 2 at offsets 0/2/4 | returns all | `a,d` \| `e` \| empty, only 3 rows to page | `a,b` \| `c,d` \| `e,f`, no overlap ✅ |
| 19 | No limit given | returns all 6 | 3 rows, empty pagination block | 6 rows, reports `limit: 50` |

### Rejections

| # | Case | List | Search (main) | Search (this PR) |
|---|---|---|---|---|
| 20 | Filter on an unknown field | n/a | 400 | 400 `'nonsense' is not a valid field` |
| 21 | Filter with an unparsable date | n/a | 400, but only because `created_at` never resolved | 400 `not a valid ISO datetime string` |
| 22 | Sort on a column the query does not select | n/a | **500** `invalid sort field: state` | 400 `'state' is not a valid sort key` |
| 23 | `group_by` | n/a | **200, silently ignored** | 400 `group_by is not supported` |

### Addressing the organization

| # | Case | List | Search (main) | Search (this PR) |
|---|---|---|---|---|
| 24 | Org given by **name** instead of id | 6 rows | **500**, log says `invalid input syntax for type uuid` | 6 rows ✅ |
| 25 | Org that does not exist | `permission_denied` | **200, empty** | `not_found` ✅ |
| 26 | Disabled organization | `failed_precondition` | 200, 0 rows, because of the join bug | 200, 1 row |
| 27 | Soft deleted organization (`deleted_at` set) | `not_found` | **200**, the org is never looked up | `not_found` ✅ |
| 28 | Organization deleted through the API | `permission_denied` | **200**, the org is never looked up | `not_found` ✅ |
| 29 | Principal required | `get` on the org | platform superuser | platform superuser |

Rows 24 and 25 were found by this comparison and fixed in the PR. Search now resolves the org first, the way List already did.

Rows 27 and 28 are the soft delete check. Search resolves the org through `GetRaw`, which reads live rows only, so a deleted org gives `not_found` instead of an empty 200. The store also skips soft deleted service users, policies and projects.

Row 26 is deliberate. Search uses `GetRaw`, so a platform admin can still inspect a disabled org. List uses `Get`, which refuses. Blocking an admin from looking inside a disabled org seemed like the wrong call, but it is easy to change if you disagree.

### What is left different on purpose

- `state`, `metadata` and `updated_at` are on List only; `projects` is on Search only.
- List accepts any caller with `get` on the org. Search requires a platform superuser.
- List has no filtering, sorting or pagination at all.

### What changed in this re-run

The earlier version of this comment had a few cells that were wrong or vague. They are corrected above:

- Row 22: `main` does not ignore an unsupported sort. It returns a **500**, `invalid sort field: state`. The old text said "ignored".
- Row 26: `main` returns 200 with **no rows** for a disabled org, not 200 with rows. The single service user there has no project policy, so the join drops it.
- Rows 14, 17, 18 and 19 said "n/a" or "untested" for `main`. They are now measured.
- The fixture note about org policies was missing, so `c-no-policy` and `f-disabled` were described wrongly.
- Rows 27 and 28, on deleted organizations, are new.

Everything else came out the same as before.

rohilsurana:
## Deleted organizations are skipped, including on the admin API

Rebased onto `main` to pick up f840f22e, which adds `live()` / `fromLive()` and makes user, organization and project reads skip soft-deleted rows.

### What the admin search does now

The handler resolves the organization through `orgService.GetRaw`, which goes to `OrganizationRepository.GetByID` / `GetByName`. Both now use `fromLive`, so a soft-deleted organization simply is not found.

Verified against a live server, one call per row:

| request | result |
|---|---|
| live org | rows returned |
| org with `deleted_at` set | **404 `not_found`** |
| same org addressed by name | **404 `not_found`** |
| org row hard deleted, service user rows left orphaned in the table | **404 `not_found`** |
| platform org `00000000-...`, which has no row by design | **404 `not_found`** |

The orphan row is the one that matters. `20260723120000_restrict_org_delete_with_serviceusers` says orphaned service users can exist and are cleaned up out of band. I left one behind on purpose by dropping the org row with the trigger disabled, and the search still returns 404 rather than listing it. On `main` that request had no organization lookup at all, so those orphans came back as normal results.

### The rest of the query

Every table the search touches carries a `deleted_at`, so each one is now filtered with the shared helper:

```sql
LEFT JOIN "policies" ON (... AND "policies"."deleted_at" IS NULL)
LEFT JOIN "projects" ON (... AND "projects"."deleted_at" IS NULL)
WHERE  "serviceusers"."org_id" = $3 AND "serviceusers"."deleted_at" IS NULL
```

The projects filter was already there, written by hand; it now uses `live(TABLE_PROJECTS)` so there is one definition of the rule. The `policies` and `serviceusers` filters are new.

Nothing sets `deleted_at` on those two tables today, since both are hard deleted, so this changes no current behaviour. It means the search stays correct if soft delete is turned on for them later, which is the direction `main` is moving.

`TestSoftDeletedRowsAreSkipped` in the postgres suite covers it: a soft-deleted service user disappears from the result, and a soft-deleted policy keeps its service user but drops the project it granted.

### One behaviour change to be aware of

Searching the platform org id now returns 404 where it used to return the bootstrap superuser. The admin UI never does this. Its only use of that id is `isZeroUUID` in the audit log view, which uses it to decide *not* to render an organization link, and platform users have their own endpoint.

### Still open

`ListOrganizationServiceUsers` also skips deleted organizations, because its handler resolves through `orgService.Get` which uses the same reads. But the plain `ServiceUserRepository.List` does not filter `serviceusers.deleted_at`. If soft delete is ever enabled for service users, that endpoint will show rows this one hides. It is shared by other callers, so it deserves its own change rather than being folded in here.

internal/store/postgres/org_serviceuser_repository.go
```diff
@@ -42,6 +47,20 @@ func (c *ServiceUserRow) transformToAggregatedServiceUser(orgID string) svc.Aggr
 	}
 }
 
+// the base query joins serviceusers, policies and projects, and all three carry an
+// id, a title and a created_at. rql is therefore applied to a wrapper select over the
+// base query, where only these aliased output columns are visible and unambiguous.
+const (
+	orgServiceUserBaseAlias = "org_service_users"
+	COLUMN_PROJECTS         = "projects"
```
AmanGIT07: Lets follow the same naming style, here its a mix of small case and capital. For `orgServiceUserBaseAlias` we can also drop the constant and pass the literal .As("org_service_users"), like for existing alisas. Else lets make it one style

internal/store/postgres/org_serviceuser_repository.go
```diff
@@ -84,125 +103,99 @@ func (r OrgServiceUserRepository) Search(ctx context.Context, orgID string, rql
...
 	}
+	query, err = utils.AddRQLSortInQuery(query, rqlQuery)
+	if err != nil {
+		return "", nil, utils.Page{}, fmt.Errorf("%w: %w", ErrBadInput, err)
+	}
+
+	// the request decides the order, so title asc is only used when nothing was asked for
+	if len(rqlQuery.Sort) == 0 {
+		query = query.OrderAppend(goqu.C(COLUMN_TITLE).Asc())
+	}
```
AmanGIT07: Lets make it clear in comments which title it sorts with, since title is ambiguous. We ca update the comment to mention that

internal/store/postgres/org_serviceuser_repository.go
```diff
@@ -5,21 +5,26 @@ import (
...
-	OrgID       string       `db:"org_id"`
-	ProjectData string       `db:"project_data"`
-	CreatedAt   sql.NullTime `db:"created_at"`
+	ID          string `db:"id"`
+	Title       string `db:"title"`
+	OrgID       string `db:"org_id"`
+	ProjectData string `db:"project_data"`
+	// only here so the aggregated project titles, which rql filters and searches on,
+	// have somewhere to land when the wrapper select returns them
+	Projects  string       `db:"projects"`
```
AmanGIT07: ServiceUserRow.Projects only exists to catch a column we never use. The wrapper does SELECT *, so the projects text comes back in every row, and sqlx fails if the struct has no  field for it. We do need that column inside the subquery, because the filter and the search run on it, but we never need it back in Go. Can we instead list the five real columns in the wrapper's select and delete this field? Postgres still lets the WHERE use projects when it is not in the select list.

AmanGIT07 (approved):

internal/store/postgres/org_serviceuser_repository.go
rohilsurana: Agreed. Dropped the constant and inlined `.As("org_service_users")`, which is what `org_pats_repository.go` and the billing repos already do. The const block is `COLUMN_*` only now, matching how every other repo file in this package declares its columns. The aliases in the base query use those consts too, instead of string literals. Done in 3e5b6f30.

internal/store/postgres/org_serviceuser_repository.go
rohilsurana: Good catch, `title` does read as ambiguous when project titles sit in the same query. The comment now says the fallback sorts on the service user title, not on the aggregated project titles. Done in 7a8c5f41.

Note that the rest of the code comments in this PR were dropped in e29238e5, so this is the only one left in the diff.

internal/store/postgres/org_serviceuser_repository.go
rohilsurana: You are right on both counts. The wrapper now selects the five columns we actually read, and the field is gone.

`projects` still resolves in the `WHERE`, because the subquery is a derived table and all of its columns are in scope there even when they are not in the select list. `TestFilterOnProjects` and `TestSearchCoversTitleAndProjects` in the postgres suite pass against a real database after the change, so that path is covered. Done in 3723ba0f.

PR #1931, opened 2026-09-17
feat(store): user, organization, and project reads skip soft-deleted rows

whoAbhishekSah:
Soft delete keeps a row and sets `deleted_at`. Reads have to skip those rows. This is the first of three PRs that make the identity repositories do that.

**What**

- Two helpers in `postgres.go`: `live(table)` is the `deleted_at IS NULL` filter, `fromLive(table)` is `dialect.From` with it applied.
- Every read in the user, organization, and project repositories goes through `fromLive`.
- Every update in those repositories adds the `live` filter too, so a deleted row is never written to.

**Behaviour change**

None today. No row has `deleted_at` set yet. This must reach every environment before any release that writes `deleted_at`.

**Tested**

One new test per repository: soft-delete a seeded row, then check every get, list, and update skips it. Store package passes with `-race`. Verified live against the sandbox twice.


whoAbhishekSah:
Added a second commit after a live test run against the branch found that `UpdateProject` and `UpdateUser` by uuid still wrote into a soft-deleted row.

`UpdateByID` and `SetState` on users, organizations, and projects now add the same `live` filter as the name-keyed updates. The three tests cover them. Project `SetState` runs a plain exec and does not report a miss, so its test checks the row's state directly.

whoAbhishekSah:
Tested live on a local server built from this branch. Fixtures: 3 users, 2 orgs, 2 projects. Ran every affected RPC before and after setting `deleted_at` directly in the DB, then diffed.

| Path | RPCs | Result |
|---|---|---|
| Get by id / name / email | GetOrganization, GetProject, GetUser | not found |
| Lists | ListAllOrganizations, ListOrganizationsByCurrentUser, ListOrganizationProjects, ListProjects, ListProjectsByCurrentUser, ListAllUsers, SearchUsers (incl. group_by) | soft-deleted row dropped |
| Member lists (GetByIDs) | ListOrganizationUsers, ListProjectUsers, ListOrganizationsByUser | soft-deleted user dropped |
| Updates by id / name / email | UpdateOrganization, UpdateProject, UpdateUser | not found, row untouched (by-id fixed in the second commit) |
| Org gate | ListOrganizationUsers, ListOrganizationProjects, CreateProject, CheckOrganizationDelete on a soft-deleted org | not found |
| Name reuse | CreateOrganization / CreateProject with a soft-deleted name | succeeds (live-only unique index) |
| Auth interceptor | any RPC with a soft-deleted user's live session | not found |

Seen but out of scope here, lands in later PRs:

- CheckResourcePermission still returns true for a soft-deleted org or project. The SpiceDB tuples are removed by the delete cascade work.
- A soft-deleted user's live session gets 404 instead of 401. Session teardown on delete handles this.
- Login or UpdateUser with a soft-deleted email creates a new account with no memberships. Intended.
- Policies, relations, resources, domains, groups, and service users are the next two PRs.

internal/store/postgres/project_repository.go
```diff
@@ -328,6 +328,7 @@ func (r ProjectRepository) SetState(ctx context.Context, id string, state projec
 		goqu.Ex{
 			"id": id,
 		},
+		live(TABLE_PROJECTS),
 	).ToSQL()
```
rohanchkrabrty: Shouldn't this have `returning` similar to other places, like `User.SetState` ?

rohanchkrabrty (approved):

PR #1934, opened 2026-09-17
test: share postgres container across suites

### LINKED ISSUE #1932: Share one postgres container across the internal/store/postgres test suites
## What happens now

Every test suite in `internal/store/postgres` starts its own postgres container. 24 test files call `newTestClient` from their `SetupSuite`, and each call pulls up a `postgres:13` container, waits for it to accept connections, drops and recreates the `public` schema, and runs the full migration set. `TearDownSuite` then throws the container away.

`make test` runs with `-count 2`, so that is 48 container starts and 48 migration runs for one CI job.

There is no `TestMain` in the package today, so there is nowhere for a shared fixture to live.

## Why it is worth fixing

The package is slow and it sits right on the edge of its own timeout.

Measured locally with the CI flags (`-race -count 2`), the package takes about 120s. Container startup is most of that. On CI it is slower, and it recently started failing outright:

```
panic: test timed out after 2m30s
FAIL	github.com/raystack/frontier/internal/store/postgres	150.221s
```

The budget was `-timeout 150s` and the package used 150.2s of it. Adding a single new suite was enough to tip it over, because the remaining headroom was about 3s. The same timeout panic showed up on several pushes before that new suite existed, so it had been flaky at the limit for a while.

PR #1929 raised `-timeout` to 600s in the Makefile to unblock CI. That is a workaround. It buys room but does not make the package faster, and the next few suites will eat the new headroom the same way.

## Suggested direction

Start one postgres container for the whole package and give each suite its own isolated database or schema on it.

- Add a `TestMain` in `postgres_test` that starts the container once, runs the migrations once, and tears it down at the end.
- Give each suite its own database (`CREATE DATABASE`) or its own schema, which costs milliseconds instead of seconds.
- Keep `newTestClient`'s signature, or replace it with a helper that hands back a client pointed at a fresh database, so the 24 suites need only a small mechanical change each.

That should turn 48 container starts into 1 and cut minutes off every run of the `unit` job.

## One related trap

`newTestClient` calls `resource.Expire(120)`, which tells docker to hard kill the container after 120 seconds. That was harmless while the whole package had to finish in 150s. Now that the timeout is 600s, any suite that runs longer than 120 seconds will have its database killed underneath it, and the failure will look like a connection error rather than a timeout. Worth handling in the same change.

## Pointers

- `internal/store/postgres/postgres_test.go` - `newTestClient`, `setup`, `purgeDocker`
- `Makefile` - the `test` target and its `-count 2 -timeout 600s`
- `.github/workflows/test.yml` - the `unit` job that runs `make test`

RushikeshBhavsar3605:
## Summary

Fixes #1932.

The Postgres repository tests were starting a separate Postgres container and running migrations for every test suite. This made the package slow and prone to hitting the test timeout.

## Changes

- Added package-level `TestMain` to start and tear down a single Postgres container.
- Run database migrations once during package setup.
- Create an isolated database for each test suite using the migrated `test_db` as a template.
- Removed the 120-second container expiry from the per-suite setup.
- Closed migration database connections so the migrated database can be used as a template.

## Test Plan

- `go test . -count=1`
- `go test -race -count=2 .`
- `go vet .`
- `git diff --check`


internal/store/postgres/postgres_test.go
```diff
@@ -54,32 +65,35 @@ func newTestClient(logger *slog.Logger) (*db.Client, *dockertest.Pool, *dockerte
...
+		fmt.Fprintf(os.Stderr, "could not start postgres resource: %v\n", err)
+		os.Exit(1)
 	}
 
-	pg_port := resource.GetPort("5432/tcp")
-	if err != nil {
-		return nil, nil, nil, fmt.Errorf("cannot parse external port of container to int: %w", err)
+	testPort = testResource.GetPort("5432/tcp")
+
+	var logWaiter interface {
```
rohilsurana, line 89: This whole block never runs.

The logger on line 56 is built as `slog.New(slog.NewTextHandler(io.Discard, nil))`. Passing `nil` for the options means the level is Info, so `logger.Enabled(ctx, slog.LevelDebug)` on line 94 is always false.

That means `logWaiter` stays nil, the container log is never attached, and the cleanup block after `m.Run()` never runs either. The `logLevelDebug` constant on line 42 is unused too.

The old code had the same dead branch, so this is not something you introduced. But since you are rewriting this function anyway, it would be nice to delete it. It is around 35 lines that do nothing.

internal/store/postgres/postgres_test.go
```diff
@@ -88,61 +102,133 @@ func newTestClient(logger *slog.Logger) (*db.Client, *dockertest.Pool, *dockerte
...
+		MaxOpenConns:    10,
+		ConnMaxLifeTime: time.Second * 60,
+		MaxQueryTimeout: time.Millisecond * 1000,
+	}
+
+	var pgClient *db.Client
+
+	if err := testPool.Retry(func() error {
+		var err error
+		pgClient, err = db.New(pgConfig)
```
rohilsurana, line 121: Not caused by your change, but this retry loop runs into it.

In `pkg/db/db.go`, `New` opens the pool and then pings it:

```go
d, err := sqlx.Open(cfg.Driver, cfg.URL)
...
if err = d.Ping(); err != nil {
    return nil, err
}
```

`sqlx.Open` always gives back a usable pool with its own background goroutine. When the ping fails, `New` returns without closing it. So that pool is orphaned and keeps trying to dial the database.

`testPool.Retry` here keeps retrying for up to 60 seconds, and `newTestClient` has a second copy of the same loop. A slow container start therefore leaves several of these behind.

One line fixes it for every caller:

```go
if err = d.Ping(); err != nil {
    _ = d.Close()
    return nil, err
}
```

internal/store/postgres/postgres_test.go
```diff
@@ -88,61 +102,133 @@ func newTestClient(logger *slog.Logger) (*db.Client, *dockertest.Pool, *dockerte
...
+
+	if err := testPool.Purge(testResource); err != nil {
+		fmt.Fprintf(os.Stderr, "could not purge postgres resource: %v\n", err)
+		code = 1
+	}
+
+	os.Exit(code)
+}
+
+func newTestClient(logger *slog.Logger) (*db.Client, *dockertest.Pool, *dockertest.Resource, error) {
```
rohilsurana, line 163: Now that there is a single shared container, the `*dockertest.Pool` and `*dockertest.Resource` return values do not carry any meaning. The `logger` argument is also never used in the new body.

Every suite still does this:

```go
s.client, s.pool, s.resource, err = newTestClient(logger)
```

I grepped for it, and `s.pool` and `s.resource` are written by all 22 suites and read by none.

Could you narrow it down to:

```go
func newTestClient() (*db.Client, error)
```

Each suite then loses two dead fields and the throwaway logger.

There is a second reason worth mentioning. Because the signature stayed the same, every existing `SetupSuite` still compiles without being touched, so the compiler could not point out the one call site that needed updating. That is how the `purgeDocker` breakage slipped through. See my comment at the bottom of the PR.

cmd/migrate.go
```diff
@@ -23,6 +23,10 @@ func RunMigrations(logger *slog.Logger, config db.Config) error {
 	if err != nil {
 		return err
 	}
+	defer func() {
```
rohilsurana, line 26: Good catch on closing this.

`RunRollback`, a few lines below, needs the same thing. It also calls `getDatabaseMigrationInstance` and never closes what comes back, so `frontier migrate-rollback` keeps the migration connection open until the process exits.

It is the same two lines you added here.

rohilsurana (commented):
Went through this properly and also ran it locally. The speedup is real, so thanks for picking it up. A few things I noticed while reading. None of these are hard blockers on their own, and I have left a separate note about the rebase.

rohilsurana:
Could you rebase this on the latest `main`?

#1929 merged after you cut this branch, and it added `internal/store/postgres/org_serviceuser_repository_pg_test.go`. That file still calls `purgeDocker`, which this PR deletes. So the package stops compiling once the two are put together:

```
internal/store/postgres/org_serviceuser_repository_pg_test.go:50:12: undefined: purgeDocker
```

The checks on this PR are green, but that run is from 17 Sep and #1929 went in on 21 Sep. CI has not built the two together yet, so the green tick does not cover this.

The fix is the same edit you already made to the other 22 suites. Point that file's `SetupSuite` and `TearDownSuite` at `closeTestClient(s.client)`.

On the change itself, I ran it locally and the numbers hold up well:

| | `main` | this PR |
|---|---|---|
| `-count 1` | 58.9s | 21.9s |
| `-count 2`, what `make test` uses | 125.4s | 19.5s |

The part I like most is that the `-count 2` time is the same as `-count 1`. The container cost is paid once now, so adding new suites stays cheap. That was the whole point of #1932, so this does the job.

internal/store/postgres/postgres_test.go
```diff
@@ -56,95 +65,140 @@ func newTestClient(logger *slog.Logger) (*db.Client, *dockertest.Pool, *dockerte
...
+
+	if err := testPool.Purge(testResource); err != nil {
+		fmt.Fprintf(os.Stderr, "could not purge postgres resource: %v\n", err)
+		code = 1
+	}
+
+	os.Exit(code)
+}
+
+func newTestClient() (*db.Client, error) {
```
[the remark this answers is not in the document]
RushikeshBhavsar3605: Fixed the remaining `newTestClient` caller in `org_projects_repository_pg_test.go` and removed the unused suite-level Docker pool/resource fields.

internal/store/postgres/postgres_test.go
RushikeshBhavsar3605: Fixed. Removed the unused logger/debug block and the associated container log handling.

internal/store/postgres/postgres_test.go
RushikeshBhavsar3605: Fixed. The DB connection is now closed when `Ping()` fails to avoid leaking the opened connection.

internal/store/postgres/postgres_test.go
RushikeshBhavsar3605: Fixed. Narrowed `newTestClient()` to return only `(*db.Client, error)` and removed the unused suite-level pool/resource fields.

cmd/migrate.go
RushikeshBhavsar3605: Fixed. Added `defer m.Close()` to ensure the migration connection is closed after rollback.

RushikeshBhavsar3605:
Rebased onto the latest `main` and updated the newly introduced PostgreSQL test as well. The changes are pushed in the latest commit.

PR #1936, opened 2026-09-18
feat(store): policy, relation, and resource reads skip soft-deleted rows

whoAbhishekSah:
Second of three PRs that make the identity repositories skip soft-deleted rows. Follows #1931.

**What**

- Every read in the policy, relation, and resource repositories goes through `fromLive`, or adds `live("p")` directly where the query aliases the table.
- The policy repository's small lookups of organization, project, and group titles for audit records use `fromLive` too.
- The two updates keyed by id add the `live` filter.
- The `ON CONFLICT` upserts are unchanged. They move together with the unique constraints in a later PR.

**Behaviour change**

None today. No row has `deleted_at` set yet.

**Tested**

One new test per repository: soft-delete a seeded row, then check every get, list, count, and update skips it. Store package passes with `-race`.


whoAbhishekSah:
Tested live on a local server built from this branch (main with #1931 merged). Fixtures: a user with an org viewer policy, a project viewer policy, and a platform admin relation, plus two custom resources under a live project. Ran every affected RPC before and after setting `deleted_at` directly in the DB, then exercised the write paths.

| Path | RPCs | Result |
|---|---|---|
| Policy reads | GetPolicy, ListPolicies | not found / empty |
| Role and count reads off the policies table | ListProjectUsers and ListOrganizationUsers with roles, ListOrganizationProjects with member count, ListOrganizationsByCurrentUser, ListProjectsByCurrentUser | hidden policy dropped, counts drop by one |
| Relation reads | GetRelation, ListPlatformUsers, ListRelations | not found / row dropped |
| Resource reads and id updates | GetProjectResource, ListProjectResources, ListResources, UpdateProjectResource, DeleteProjectResource | not found / row dropped |
| SpiceDB-backed checks | CheckResourcePermission on org, project, hidden resource; superuser check for the hidden platform relation | still true, tuples untouched |

Write paths against a hidden row, worth noting for the phased plan:

- CreatePolicy, AddPlatformUser, and CreateProjectResource hit `ON CONFLICT` on the hidden row. Each returns success with the old id, and the row keeps `deleted_at`, so a Get on that id says not found. The resource title was even updated in place. This moves with the unique constraints, as the description says, but until then a caller gets a success and a ghost.
- RemovePlatformUser and DeleteRelation return success but do nothing. The relation service looks the row up through the live filter first, finds nothing, and returns nil without deleting the SpiceDB tuple. The user stayed a superuser. A not found would be safer than a silent success.
- DeletePolicy returns not found and RemoveOrganizationMember reports not a member while SpiceDB still grants. Consistent with the reads. The cascade work owns the tuple cleanup.

No behaviour change today since no row has `deleted_at` set.

internal/store/postgres/resource_repository.go
```diff
@@ -149,7 +149,7 @@ func (r ResourceRepository) GetByID(ctx context.Context, id string) (resource.Re
 		return resource.Resource{}, resource.ErrInvalidID
 	}
 
-	query, params, err := dialect.From(TABLE_RESOURCES).Where(goqu.Ex{
+	query, params, err := fromLive(TABLE_RESOURCES).Where(goqu.Ex{
```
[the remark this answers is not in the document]
whoAbhishekSah: Mapped the duplicate key to `resource.ErrConflict` in the create path, with a test that reuses a soft-deleted row's id under a new URN. Left the pre-check on live rows only, since the primary key already refuses the insert and the handler turns the conflict into a 409.

internal/store/postgres/policy_repository.go
AmanGIT07, line 403: The locked CTE still counts soft-deleted rows, so a deleted owner policy keeps the guard from firing and the last live owner can be removed. I checked with a test: one live owner plus one soft-deleted owner, and the guard deletes the live one. Can we add AND deleted_at IS NULL to the CTE and cover it with a test that soft-deletes the second holder?

internal/store/postgres/policy_repository.go
```diff
@@ -625,19 +625,19 @@ func (r PolicyRepository) getResourceInfo(ctx context.Context, tx *sqlx.Tx, reso
...
 	case schema.ProjectNamespace:
-		projQuery, projParams, _ := dialect.From(TABLE_PROJECTS).
+		projQuery, projParams, _ := fromLive(TABLE_PROJECTS).
 			Select("org_id", "title").
 			Where(goqu.Ex{"id": resourceID}).
 			ToSQL()
 		_ = tx.QueryRowContext(ctx, projQuery, projParams...).Scan(&orgID, &resourceName)
 	case schema.GroupNamespace:
-		grpQuery, grpParams, _ := dialect.From(TABLE_GROUPS).
+		grpQuery, grpParams, _ := fromLive(TABLE_GROUPS).
```
AmanGIT07: These lookups only fill the org id and title on the audit record. Once projects and groups are soft-deleted, a policy removed under one of them gets no org id and drops out of the org's audit view. The org title lookup in the audit insert reads all rows today. Can we keep these three on the plain From so the trail stays attributed?

internal/store/postgres/policy_repository.go
AmanGIT07, line 145: The role join doesn't have deleted_at filter, will this be part of role repository change? If not, we can add it here

internal/store/postgres/resource_repository_test.go
```diff
@@ -411,3 +411,40 @@ func (s *ResourceRepositoryTestSuite) TestUpdate() {
...
+		s.T().Fatal(err)
+	}
+
+	_, err = s.repository.GetByID(s.ctx, deleted.ID)
+	s.Assert().ErrorIs(err, resource.ErrNotExist)
+
+	_, err = s.repository.GetByURN(s.ctx, deleted.URN)
+	s.Assert().ErrorIs(err, resource.ErrNotExist)
+
+	got, err := s.repository.List(s.ctx, resource.Filter{ProjectID: deleted.ProjectID})
```
AmanGIT07: This project has only one resource, and we just soft-deleted it. So `got` is empty here and the loop below never runs. The test only proves that List did not fail. Can we check the count instead, like the policy test does? For example, call List with no filter and expect `len(s.resources)-1` rows.

internal/store/postgres/relation_repository_test.go
```diff
@@ -323,3 +323,26 @@ func (s *RelationRepositoryTestSuite) TestDeleteByID() {
...
+	deleted := s.relations[0]
+	_, err := s.client.ExecContext(s.ctx, "UPDATE relations SET deleted_at = now() WHERE id = $1", deleted.ID)
+	if err != nil {
+		s.T().Fatal(err)
+	}
+
+	_, err = s.repository.Get(s.ctx, deleted.ID)
+	s.Assert().ErrorIs(err, relation.ErrNotExist)
+
+	got, err := s.repository.List(s.ctx, relation.Filter{Subject: deleted.Subject, Object: deleted.Object})
```
AmanGIT07: Only one relation matches this subject and object, and we just soft-deleted it. So `got` is empty and the loop never runs. The same happens for `byFields` below. Can we check the count instead? For example, call List with no filter and expect one row less than we started with.

AmanGIT07:
About RemovePlatformUser and DeleteRelation returning success but doing nothing on a hidden row, from your testing note: is that fix meant for this PR, or is it planned for later? If later, which change will pick it up?

internal/store/postgres/policy_repository.go
whoAbhishekSah: Good catch. The locked CTE now has AND deleted_at IS NULL. Added a test with three holders that soft-deletes one: removing the first succeeds, removing the second is refused because the deleted one no longer counts. The test fails without the fix.

internal/store/postgres/policy_repository.go
whoAbhishekSah: Agreed. The three lookups are back on the plain From so the audit record keeps its org id and title after the parent is deleted.

internal/store/postgres/policy_repository.go
whoAbhishekSah: Added live("r") on the join here. The role repository's own reads were not on the list; they go into the last PR of this series with domains, groups, and service users.

internal/store/postgres/resource_repository_test.go
whoAbhishekSah: Changed to list with no filter and compare the count before and after, plus the id loop on the remaining rows.

internal/store/postgres/relation_repository_test.go
whoAbhishekSah: Same change here: list with no filter, count before and after. GetByFields now asserts an empty result instead of looping over nothing.

whoAbhishekSah:
On RemovePlatformUser and DeleteRelation returning success on a hidden row: not this PR. The relation service already returned nil for a relation that was never there, so this is the existing behaviour showing up on hidden rows. Future work on relation writes will decide whether a miss should become not found. The org delete cascade removes memberships that may already be gone, so that decision needs to check it does not rely on the delete being repeatable.

AmanGIT07 (approved):

PR #1938, opened 2026-09-22
feat(store): domain, group, role, and service user reads skip soft-deleted rows

whoAbhishekSah:
Last of three PRs that make the identity repositories skip soft-deleted rows. Follows #1931 and #1936.

**What**

- Every read in the domain, group, role, service user, and service user credential repositories goes through `fromLive`, or adds `live("s")` where the query aliases the table.
- Updates keyed by id add the `live` filter. The policy subquery in `ListMissingOrgPolicy` only counts live policies.
- The organization title subqueries in the service user create and delete paths stay on the plain `From`. They only attribute the audit record.
- The role `ON CONFLICT` upsert is unchanged. It moves with the unique constraint in a later PR.

**Behaviour change**

None today. No row has `deleted_at` set yet.

**Tested**

Domains and service users had no repository suite, so this adds one for each. Groups and roles get a case in their existing suites. Each test soft-deletes a row and checks every get, list, and update skips it, comparing counts before and after. Store package passes with `-race`.


whoAbhishekSah:
Tested live on a local server built from this branch (main with #1931 and #1936 merged). Fixtures: a verified domain on a live org, two groups with members, a custom org role bound to a user, two service users with Basic credentials and a minted access token, a service user project policy, and a throwaway org with its own service user. Ran every affected RPC before and after setting `deleted_at` directly in the DB, then exercised the write paths.

| Path | RPCs | Result |
|---|---|---|
| Domain reads | GetOrganizationDomain, ListOrganizationDomains, joinable orgs in ListOrganizationsByCurrentUser | hidden domain dropped |
| Domain join | JoinOrganization with the verified domain hidden | refused |
| Group reads | GetGroup, ListOrganizationGroups, ListCurrentUserGroups, ListGroupUsers, ListProjectGroups | not found / dropped |
| Role reads | GetOrganizationRole, ListOrganizationRoles, roles in ListProjectUsers | not found / hidden role dropped |
| Service user reads | GetServiceUser, ListServiceUsers, ListOrganizationServiceUsers, ListAllServiceUsers, ListProjectServiceUsers, ListServiceUserProjects, ListServiceUserCredentials | not found / dropped |
| Service user auth | Basic auth as a hidden service user; Basic auth with a hidden credential on a live service user | unauthenticated, live credential still works |
| Writes by id | UpdateGroup, DisableGroup, DeleteGroup, SetGroupMemberRole, UpdateOrganizationRole, DeleteOrganizationRole, DeleteServiceUser | not found, row untouched |
| Name reuse | CreateOrganizationDomain, CreateGroup with a hidden name | new live row |
| Role upsert | CreateOrganizationRole with a hidden name | returns the old id, row stays deleted (moves with the unique constraint, as described) |
| SpiceDB-backed checks | CheckResourcePermission through a hidden group | still true, tuples untouched |

One gap worth planning for before service users can be soft-deleted:

- **DeleteOrganization with a hidden service user leaves the org half deleted.** The deleter lists service users through the live filter, sees none, deletes every policy on the org, then the org row delete hits the `trg_serviceusers_block_org_delete` trigger, which counts soft-deleted rows too. The call returns an internal error. The deleter is not transactional, so the org row stays with zero policies, and the owner is locked out. Either the trigger should ignore rows with `deleted_at` set, or the deleter should purge hidden service users. Not this PR, but it belongs with the service user delete work.

Pre-existing internal errors that soft delete makes easier to hit. Each returns 500 instead of not found today, even for a random uuid:

- GetOrganizationDomain, VerifyOrganizationDomain, DeleteOrganizationDomain: the domain repository does not map no rows to `domain.ErrNotExist`.
- CreatePolicy with a missing role id: the handler does not map `role.ErrNotExist`.
- Bearer access token for a missing service user: the authenticator returns the raw error instead of unauthenticated.

Minor: CreateServiceUserCredential succeeds under a hidden service user (the credential cannot authenticate). ListServiceUserCredentials still lists a hidden service user's credentials. DeleteServiceUserCredential on a hidden credential hard-deletes it and reports success. ListProjectGroups and ListProjectServiceUsers drop the principal from the list but keep it in `role_pairs`.

No behaviour change today since no row has `deleted_at` set.

AmanGIT07 (approved):

PR #1940, opened 2026-09-22
fix: map three missing-row cases to not found or unauthenticated

whoAbhishekSah:
Three places returned an internal error for a row that does not exist. Live testing of the soft delete work surfaced them, but all three happen today with any unknown id.

**What**

- Domain get, verify, and delete: the repository returned a wrapped database error when no row matched. It now returns the domain not-found sentinel, which the handlers already map to a 404.
- `CreatePolicy` with a role id that does not exist: the handler now maps the role not-found error to a 404 instead of falling through to a 500.
- A bearer token for a service user that no longer exists, on both the minted access token path and the JWT grant path: the authenticator now returns unauthenticated instead of the raw error.

**Behaviour change**

Callers get 404 or 401 where they got 500. No success path changes.

**Tested**

Handler test for the missing role. Authenticator tests for the missing service user on both token paths. Verified live: domain and role cases return 404 where they returned 500. The domain repository has no suite on main yet; one arrives in #1938 and I will add the not-found case there once both are in.


whoAbhishekSah:
Tested live on a local server built from this branch (main, without #1938). Every case uses an id that does not exist, so the missing soft-delete filters do not matter here.

| Fix | RPCs | Result |
|---|---|---|
| Domain missing row | GetOrganizationDomain, VerifyOrganizationDomain, DeleteOrganizationDomain with an unknown id | 404 (was 500); a live domain still reads fine |
| CreatePolicy missing role | unknown role uuid, unknown role name | 404 (was 500); a valid role still creates the policy |
| JWT grant for a missing service user | grant path (`authenticateWithJWTGrant`) | not reproducible live on this build: a hard delete removes the credential with the service user, and the foreign key blocks a credential without a parent. Reachable once #1938 lands (service user hidden, credential live). Will retest then. |

One note. The 500 from my #1938 test came from the other token path, and it still happens on this build:

- Create a service user and a credential, mint an access token with `AuthToken` (`client_credentials`), confirm Bearer and Basic both work.
- `DeleteServiceUser`.
- Bearer with the minted token returns internal. Basic with the same credential returns unauthenticated.

That is `authenticateWithAccessToken`, which returns the raw error from `serviceUserService.Get` (around line 154). Mapping `serviceuser.ErrNotExist` to unauthenticated there, the same way this PR does for the grant path, would close it. Could ride along here or go in a follow-up.

Small observation, not a regression: a malformed role id like `not a role!` also comes back as 404, because the name lookup treats it as a missing name.

whoAbhishekSah:
Right, the 500 you hit is the access token path, not the grant path. Added a second commit that maps a missing service user to unauthenticated in `authenticateWithAccessToken` too, with a test that mints a token whose subject no longer exists. The grant path change stays, since it becomes reachable once a service user can be hidden with its credential still live.

On the malformed role id coming back as 404: agreed it is not a regression. The role lookup treats anything that is not a uuid as a name, so an unknown name and a malformed id look the same to it.

AmanGIT07 (approved):

